Issue #183, opened 2015-01-19
dsolve fails on particular intergral, ignores some bits of forcing


cbm755:
Define an inhomogenous linear system of ODEs:

```
octave:52> syms x(t) y(t)
octave:53> e1 = diff(x,t) == x+4*y - t + 2
e1 = (sym)

  d                                
  ──(x(t)) = -t + x(t) + 4⋅y(t) + 2
  dt                               

octave:54> e2 = diff(y,t) == x - 2*y + 5*t
e2 = (sym)

  d                             
  ──(y(t)) = 5⋅t + x(t) - 2⋅y(t)
  dt                            

octave:55> S = dsolve([e1, e2])
S = 
{
 = (sym)

               -3⋅t         2⋅t   2
  x(t) = 4⋅C₁⋅ℯ     + 4⋅C₂⋅ℯ    - ─
                                  3

 = (sym)

                 -3⋅t       2⋅t   1
  y(t) = - 4⋅C₁⋅ℯ     + C₂⋅ℯ    - ─
                                  3

}
octave:70> X = rhs(S{1}),  Y = rhs(S{2})   # extract solutions
X = (sym)

        -3⋅t         2⋅t   2
  4⋅C₁⋅ℯ     + 4⋅C₂⋅ℯ    - ─
                           3

Y = (sym)

          -3⋅t       2⋅t   1
  - 4⋅C₁⋅ℯ     + C₂⋅ℯ    - ─
                           3

octave:71> 
octave:71> diff(X,t) - (X+4*Y - t + 2)   # should be zero
ans = (sym) t
octave:72> diff(Y,t) - (X - 2*Y + 5*t)
ans = (sym) -5⋅t
```

Notice its as if the `t` terms did not appear in the forcing.

Haven't yet checked if it happens for scalars or just systems.

@maprieto any thoughts?

cbm755:
seems ok for scalar:
````
octave:81> e1 = diff(x,t) == x + 2*t + 2
e1 = (sym)

  d                        
  ──(x(t)) = 2⋅t + x(t) + 2
  dt                       

octave:82> dsolve(e1)
ans = (sym)

         ⎛                 -t⎞  t
  x(t) = ⎝C₁ + 2⋅(-t - 2)⋅ℯ  ⎠⋅ℯ 

octave:83> X = rhs(ans)
X = (sym)

  ⎛                 -t⎞  t
[...]

PR #188, opened 2015-01-24
Update (i)fourier and (i)laplace functions

maprieto:
The workaround for issue #183 is mainly based on the computation of the general solution of the homogeneous system and a particular solution of the ODE system by using Laplace transforms. It only works on sympy 0.7.6 since Laplace transforms are applied to matrices (which is not supported on 0.7.5). As a previous step, the ODE system is written in matrix form.

Some additional features have been added to dsolve for ODE systems:
* Constant parameters can be used in the initial conditions (both in unknown arguments and their values)
* It's possible to describe the initial conditions by means of a linear system of equations with unknowns x(0), y(0),...

Obviously, there are some drawbacks for an implementation based on Laplace transforms:
* Its performance depends basically on the computational cost of the Laplace transform, which is pretty poor in some cases
* The solution is not explicitly computed in those linear systems with a rhs without an explicit Laplace transform or when the eigenvalues of the system matrix are not simple enough.


cbm755:
Have you looked at upstream to see what the problem is?  (I haven't)

I prefer to fix it there (keeping octsympy as *thin* as possible).  If we don't know why this simple thing fails, its a bit worrying to rely on that code in other situations.  In general, I don't want to play wack-a-mole against sympy's bugs, while our dsolve and theirs diverge further and further.

Does this workaround happen *only* when there is a non-constant forcing?  That might be ok temporarily (similar to the ICs thing).  If we have this sort of workaround, it should be as narrowly defined as possible (apologies for not reading your code in enough detail to answer this myself).

I'm not likely to think about this for a week or so :(

maprieto:
I've looked for these issues related to dsolve appled to ODE systems and I couldn't identify the source of the wrong solutions (maybe related to the classification of systems, identifycation of right-hand sides...?). Anyway, I'm totally agree with you to keep the code as thin as possible, but pushed for being able to handle ODE systems in my courses, I've decided to implement a workaround for dsolve. Currently, dsolve use the Laplace transform approach for any kind of ODE system (independently of the type of the forcing terms) and keep the sympy function dsolve for scalar ODEs.

Feel free to merge it temporaly at the master branch. Anyway, I'll use it for my courses while the bug in sympy is not fixed.

cbm755:
That's fine: I assume you have it place for your course, whether I not I merge this?

maprieto:
Right, don't worry!...and, for the following months (from february to june) an undergraduate student (in computer science) will be involved working with octsympy (firstly, studying the implementation and then trying to fix some issues, implementing new functions,...). So, from time to time, we'll propose you a merge with some contributions.

cbm755:
> Right, don't worry!...and, for the following months (from february to
> june) an undergraduate student (in computer science) will be involved
> working with octsympy (firstly, studying the implementation and then
> trying to fix some issues, implementing new functions,...). So, from
> time to time, we'll propose you a merge with some contributions.

Right-on!  Look forward to it.

cbm755:
@maprieto I completely missed this, as I just assumed it was the same PR as before.  But this is different and I'll take a look.

(You should probably use branches for each PR)

maprieto:
My apologies: It was my fault. For the next PR, we'll try to identify the target of the pull request using independent branches.

cbm755:
no problem!  Merged these fourier commits in #246, so closing this.

PR #199, opened 2015-02-06
allow symfun(<symfun>, t)

### LINKED ISSUE #189: Assigning a function to a function fails if the second function's argument is missing in the symbolic expression
The following code

    syms t
    x(t) = t;
    y(t) = x;

results in the error

    error: assignment failed, or no method for '<unknown type> = class'

Note that

    y(t) = x(t);

works fine, but for the sake of MATLAB compatibility both expressions should work.

cbm755:
Fixes #189.  Add tests.

@rocketsound, I've got a fix.  If you have a few mins, can you look at these tests and see if you think they are appropriate and cover the issues you identified?

### DIFF
--- a/inst/@symfun/symfun.m
+++ b/inst/@symfun/symfun.m
@@ -112,8 +112,13 @@ function f = symfun(expr, vars)
     expr = python_cmd (cmd, vars{:});
   end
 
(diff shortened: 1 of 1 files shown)


cbm755:
If you think these tests are ok, I will merge this.

PR #214, opened 2015-03-15
New functions for Catalan and Euler-Mascheroni mathematical constants.

carandraug:
Adds two new functions for the symbolic package. I also have a python script to generate them automatically but in the end I thought that although quite simple, the script would be overkill.

### DIFF
--- a/inst/catalan.m
+++ b/inst/catalan.m
@@ -35,3 +35,5 @@ function g = catalan ()
   g = python_cmd ('return sympy.S.Catalan,');
 end
 
+%!assert (double (catalan ()) > 0.915965594177)
+%!assert (double (catalan ()) < 0.915965594178)
--- a/inst/catalan.m
+++ b/inst/catalan.m
@@ -1,0 +1,37 @@
+%% Copyright (C) 2015 Carnë Draug
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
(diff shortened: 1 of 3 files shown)


inst/catalan.m
```diff
@@ -0,0 +1,37 @@
...
+%% @seealso{eulergamma}
+%% @end deftypefn
+
+%% Author: Carnë Draug
+%% Keywords: symbolic, constants
+
+function g = catalan ()
+  g = python_cmd ('return sympy.S.Catalan,');
+end
+
```
cbm755: trailing white space.

(Usually I want tests although I'm soon going to support doctests so maybe its overkill...)

inst/eulergamma.m
```diff
@@ -0,0 +1,37 @@
...
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @deftypefn {Function File} {} eulergamma ()
+%% Return Euler-Mascheroni constant.
+%%
+%% @example
+%% vpa (catalan ())
```
cbm755: eulergamma

(and with every such error my belief in doctests grows stronger!)

inst/eulergamma.m
```diff
@@ -0,0 +1,37 @@
...
+%% @seealso{catalan}
+%% @end deftypefn
+
+%% Author: Carnë Draug
+%% Keywords: symbolic, constants
+
+function g = eulergamma ()
+  g = python_cmd ('return sympy.S.EulerGamma,');
+end
+
```
cbm755: trailing white space

cbm755:
+1 after those changes.

carandraug:
About the typo on documentation [here's the script](https://bpaste.net/show/16011b9109db) mentioned on the pull request (with this problem fixed) I guess it could be adjusted to fit with the other script you have on src/ but I didn't want to mess with your build system. Also, Matlab seems to only have this two functions, so unless you want to have the other constants in symbolic, there's even less reason for it.

I have also added two simple tests. I only started playing with the symbolic package this weekend (I was having fun implementing `psi` for Octave core) so I'm unsure what would be the best way.

I have also removed the last empty line as requested.

cbm755:
Awesome thanks for this.  those tests should be fine.  This is in.

>  but I didn't want to mess with your build system.

That was very polite ;-)  But yes it is horrid.

I used to do code generation at install time, but now I've just checked the results in: slightly lowers barrier for new contributions (I hope).  

On the topic of build systems, is https://github.com/cbm755/octsympy/issues/196 an easy fix for you?

PR #196, opened 2015-03-17
configure script to detect Python and SymPy

cbm755:
Now that we are the symbolic package, users can install from within Octave using `pkg install -forge symbolic`.  I have not tested this when Python and/or SymPy are missing.

But it would be nice if it gives an error at install time, telling people to install Python and SymPy first...

I'm no expect on autoconf, etc, but this should be easy enough for someone who is!

### DIFF
--- a/HOWTO-release.md
+++ b/HOWTO-release.md
@@ -21,6 +21,8 @@ Checklist
 
   * Update NEWS file (date, version number, reformat).
 
+  * Update the version in src/configure.ac.  Run bootstrap script.
+
   * Packages: need to run the following two scripts.  I recommend
     testing them first without using the tag.  Then test the packages
     by running the test suite.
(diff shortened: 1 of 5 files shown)


cbm755:
Eh!  Did not know you could migrate an issue to a PR!  neat.

I'm looking into this... do you know how to do a version check?  Need minimum version sympy 0.7.5.

carandraug:
Done. It will now fail if version < 0.7.5

yoihito:
Hey-hey! It fails on python 3.4.2. `configure.ac` has some exceptions when it tries to get SymPy version.

cbm755:
Thanks @yoihito   Can you file a new bug with the text of the exceptions?  (If you're on Windows, perhaps its #236?)

carandraug:
@yoihito I have no experience with Python 3 at all. What exception do you get? Does the package even supports python 3 at all? If not, a better check would be to croak for experimental python versions.

cbm755:
thanks @carandraug but @yoihito has already fixed this in ca17768dcb40c9aef47a1fa2a8e759d2ffc115c1.

For the record, we do (recently) support Python 3.  It should run with either Python 2 or Python 3.

PR #225, opened 2015-04-06
Fourier transform in angular frequency.

AlexanderMisel:
I modified and tested the files.

### DIFF
--- a/inst/@sym/fourier.m
+++ b/inst/@sym/fourier.m
@@ -27,11 +27,32 @@
 %%
 %% Examples:
 %% @example
-%% syms x w
+%% syms x s
 %% f = exp(-2*abs(x))
 %% fourier(f)
-%% fourier(f, w)
-%% fourier(f, x, w)
+%% @result{} ans = (sym)
+%%
+%%    4   
+%%  ──────
+%%   2    
+%%  w  + 4
+%%
+%% fourier(f, s)
+%% @result{} ans = (sym)
+%%
+%%    4   
+%%  ──────
+%%   2    
+%%  s  + 4
+%%
+%% fourier(f, x, s)
+%% @result{} ans = (sym)
+%%
+%%    4   
+%%  ──────
+%%   2    
+%%  s  + 4
+%%
 %% @end example
 %%
 %% @seealso{ifourier,laplace}
--- a/inst/@sym/fourier.m
+++ b/inst/@sym/fourier.m
@@ -28,7 +28,7 @@
 %% Examples:
 %% @example
 %% syms x s
-%% f = exp(-2*abs(x))
+%% f = exp(-2*abs(x));
 %% fourier(f)
 %% @result{} ans = (sym)
 %%
@@ -83,8 +83,8 @@ function F = fourier(varargin)
  
   endif
 
-  cmd = { 'sp.FourierTransform._b = -1'
-          'F = sp.fourier_transform(*_ins)'
+  cmd = { 'from sympy.integrals.transforms import _fourier_transform'
+          "F = _fourier_transform(*(_ins+[1,-1,'Fourier']))"
           'return F,'};
   F = python_cmd(cmd,f,x,w);
 
--- a/inst/@sym/fourier.m
+++ b/inst/@sym/fourier.m
@@ -1,4 +1,4 @@
-%% Copyright (C) 2014 Colin B. Macdonald
+%% Copyright (C) 2015 Colin B. Macdonald, Alexander Misel
 %%
 %% This file is part of OctSymPy.
 %%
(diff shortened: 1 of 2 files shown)


inst/@sym/fourier.m
```diff
@@ -17,28 +17,55 @@
...
+%% @seealso{ifourier,laplace}
 %% @end deftypefn
 
 %% Author: Colin B. Macdonald
 %% Keywords: symbolic
 
-function F = fourier(f, x, k)
+function F = fourier(varargin)
+
+  syms f x w;
```
cbm755: remove this line?

inst/@sym/fourier.m
```diff
@@ -17,28 +17,55 @@
...
+  if (nargin == 1)
+    f=varargin{1};
+    x=symvar(f,1);
```
cbm755: I prefer `x = symvar(f, 1)` (just differing whitespace), probably not a big deal but would match the rest of the code

inst/@sym/fourier.m
```diff
@@ -17,28 +17,55 @@
...
+    f=varargin{1};
+    x=varargin{2}; 
+    w=varargin{3};
 
-  cmd = { 'd = sp.fourier_transform(*_ins)'
-          'return d,' };
+  else
+    error('Wrong number of input arguments') 
+ 
+  endif
```
cbm755: for matlab compatibility, please just use `end`.

inst/@sym/fourier.m
```diff
@@ -17,28 +17,55 @@
...
 
-  F = python_cmd (cmd, sym(f), sym(x), sym(k));
+  cmd = { 'sp.FourierTransform._b = -1'
```
cbm755: what is this for?  I cannot find it in docs?  If its important, can you leave a comment here?

inst/@sym/ifourier.m
```diff
@@ -17,31 +17,63 @@
...
+%% @seealso{fourier,laplace}
 %% @end deftypefn
 
 %% Author: Colin B. Macdonald
 %% Keywords: symbolic
 
-function f = ifourier(F, k, x)
+function f = ifourier(varargin)
+
+  syms F w x;
```
cbm755: as above, remove this

inst/@sym/ifourier.m
```diff
@@ -17,31 +17,63 @@
...
-  cmd = { 'd = sp.inverse_fourier_transform(*_ins)'
-          'return d,' };
+  else
+    error('Wrong number of input arguments') 
+ 
+  endif
 
-  f = python_cmd (cmd, sym(F), sym(k), sym(x));
+  cmd = { 'sp.InverseFourierTransform._a = 0.5/S.Pi'
+          'sp.InverseFourierTransform._b = 1'
```
cbm755: again please comment this.  Is it really necessary for us to mess around with `_` variables?  Is this some functionality not exposed in the `.inverse_fourier_transform` cmd?

inst/@sym/ifourier.m
```diff
@@ -17,31 +17,63 @@
...
 
 
-%! % tests in fourier.m
-%!assert(true)
+%!test
+%! syms x k
+%! f = exp(-x^2);
+%! F = fourier(f,x,k);
+%! g = ifourier(F,k,x);
+%! assert(isequal(f,g))
```
cbm755: is just copy-paste of the fourier.m ones?  How about testing that the calling forms without x,k work?  Actually I guess the doctests will get that as you've added it to the docs, so not necessary.  This is ok.

cbm755:
This looks pretty good thanks!  Can you adjust as per my comments about and then I'll merge.

cbm755:
For next time: you should do your work on a branch other than master...   you may have multiple branches submitted for PRs: its likely you'll want/need to revise those, easier if each PR is a new branch (rather than master).

AlexanderMisel:
@cbm755 Maybe you didn't see my comments when I made my first commit, I copy and paste it here:

MATLAB uses non-unitary angular frequency convention, while SymPy uses unitary ordinary frequency.
https://en.wikipedia.org/wiki/Fourier_transform#Other_conventions

Fourier transform has many conventions, which differs only in the choice of `_a` and `_b` of the function called `_fourier_transform` which defined at `class FourierTransform` and `class InverseFourierTranform` in SymPy source, which chose 

<script type="text/javascript" src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=default">
F(k) = \int_{-\infty}^\infty f(x) e^{-2\pi i x k} \mathrm{d} x.
</script>

and the inverse

<script type="text/javascript" src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=default">
f(x) = \int_{-\infty}^\infty F(k) e^{2\pi i x k} \mathrm{d} k.
</script>

AlexanderMisel:
Actually I am new to python, and all I thought was that `sp` is just an object, and changing its `_a` and `_b` value will not make effects globally. But I dunno whether I am right or wrong. I used to consider calling `_fourier_tranform` directly and passing both parameters as well as a `name` string to it, while it can be complicated.

cbm755:
Ok, I did figure that was what is was for but the code needs some comments about this.  I'm not entirely comfortable messing around with "hidden" (_) variables: by convention those are private.
1.  Is there a way to do this with some "kwarg" in SymPy?  If not, is there an easy patch we could submit to SymPy to add this?  There might be an upstream SymPy issue about this already...
2.  We can always merge it here as you have it, provided it has some warning in comments.
3.  Please add some tests, either in the documentation or as "%!" tests that demonstrate this choice of constants.  Easiest to do it in the documentation, that way we test it (with doctests) and document it.  
4.  We should document this choice.  I think maybe a better reason that "be like Symbolic Math Toolbox" might be for compatibility with the Octave `fft` function.  Can you confirm that is the case?

cbm755:
Here's how to do doctests.  Basically it splits around `@result{}` to find input and output and then tests that it works.
````
@example
@group
syms x
f = fourier(...)
@result{} (sym) x^2
@end group
@end example
````

**Comment by AlexanderMisel** [comment]
@cbm755 I feel difficult to write tests since SymPy calculate symbolic fourier simply by integration and thus some general functions or distributions are not integrated (like Dirac's delta function and Heaviside...). Some inverse fourier transforms I tested today gave the wrong answer and I think it's SymPy's error. Maybe you could help me for the doc.

**Comment by cbm755** [comment]
Sure, I can try to help with docs.

1.  Maybe clean it up as best you can re: my other comments above.  We can always do other PRs for the docs: although it would be good if Travis passes.
2.  I looked a bit at `_fourier_transform`.  Can we call that and pass in `a` and `b` values instead of changing `_a` in the class?

**Comment by AlexanderMisel** [comment]
Certainly we can call it with the following command for Fourier transform if you like:
```
  cmd = { 'from sympy.integrals.transforms import _fourier_transform'
          "F = _fourier_transform(*(_ins+[1,-1,'Fourier']))"
          'return F,'};
```

**Comment by cbm755** [comment]
I think that `cmd` looks better than before.

Re: fft: I mean, fft also has choice and Octave uses 1/N in the inverse.  So presumably some symmetry there if we put the `1/(2*pi)` with the inverse right?

See Travis output for doctest failure, you need `2*sym(pi)` somewhere.

**Comment by cbm755** [comment]
The failing test about execfile is my fault: ignore that, I'll fix it in master.  But the tests for fourier and ifourier are failing...

File: inst/@sym/fourier.m [hunk 8]
```diff
@@ -17,28 +17,76 @@
...
-  cmd = { 'd = sp.fourier_transform(*_ins)'
-          'return d,' };
+  else
+    error('Wrong number of input arguments') 
+ 
+  endif
 
-  F = python_cmd (cmd, sym(f), sym(x), sym(k));
+  cmd = { 'from sympy.integrals.transforms import _fourier_transform'
+          "F = _fourier_transform(*(_ins+[1,-1,'Fourier']))"
```
cbm755: For matlab compat, I prefer:
````
 'F = _fourier_transform(*(_ins+[1,-1,"Fourier"]))'
````

cbm755:
There are also still various comments I made earlier to be addressed.  Let me know if you need help.

cbm755:
I think these changes are done in a slightly different way by @maprieto. I've rebased your documentation updates in #247, which I'll merge soon.

Thanks for this, and please let me know if I've missed something here.

PR #240, opened 2015-05-08
Fix configure.ac print version for Python 3

yoihito:


cbm755:
Oops!  Yes that looks good!  Thanks for patch, will merge when tests pass.

Issue #259, opened 2015-07-12
fix up `@result`, `@print` usage in docs


cbm755:
`@print` for console output, `@result{}` for each return value (i.e., use more than one of them for multiple return values).

cbm755:
I think this is mostly done.

PR #260, opened 2015-07-13
Use `@print` and other doctest updates

cbm755:
Probably this won't work in Travis CI until we have bump doctest...  merge after that happens.


cbm755:
Should fix #259, although there are probably other instances to fix too.

Issue #251, opened 2015-06-25
`fourier(sym(2), x, w)` gives `0` instead of dirac


cbm755:

maprieto:
This is issue is confirmed in Sympy. Also other Fourier series are not computed correctly:
```
import sympy as sp
from sympy.abc import x, w
sp.fourier_transform(2, x, w)
sp.fourier_transform(sp.cos(x), x, w)
sp.fourier_transform(sp.sin(x), x, w)
```
So, it could be labelled as upstream (see https://github.com/sympy/sympy/issues/2803)

cbm755:
Possibly related: sympy does something stupid where it assumes the first choice of every piecewise result is the answer (!)  This is one reason so many transforms give 0.  I've been meaning to look into fixing that.

Thanks for the link.

cbm755:
@maprieto has added a work around for us, but it still needs to be fixed in upstream sympy

PR #261, opened 2015-07-15
Integral transforms

maprieto:
This pull request includes patches for the (i)Laplace and (i)Fourier functions to handle Dirac deltas and Heaviside functions adequately and (i)Fourier transforms of simple functions such as constant ones, trigonometric functions (see the new tests included in each function). These patches solve partially issue #251


cbm755:
Why don't we try to get this or at least parts of it into SymPy?  I really like the idea of staying slim and fixing SymPy instead...

cbm755:
I think we need this, its just a think SymPy needs this too!

maprieto:
I agree with you: it would be great to keep the symbolic package as slim as possible. However, since the new patches provide some basic functionality on these functions, it's worth it to include them (at least until sympy has solved them). Feel free to share these patches with those collaborators involved in the sympy development.

cbm755:
Well, I feel it will be a lot of work for me:
1. need to understand these changes
2. need to sort out where to try to patch this into SymPy.
You've already done 1 so are much ahead of me.  Can you help with 2?

There was a recent discussion on the SymPy maillist that touched on this:
https://groups.google.com/forum/#!topic/sympy/2FVQRtZFpeY

maprieto:
Ok! I'll post the possibility of use this patch on SymPy on the discussion that you've referred

cbm755:
I'll bite the bullet and merge this.   I have not carefully vetted this code.  I would still like to see these improvements go upstream.

maprieto:
I agree with you: I'd like to see a new smart implementation of this kind of improvements upstream in Sympy. Meanwhile, it can be considered as a partial remedy.

PR #264, opened 2015-09-19
Fix disp for compatibility with Octave 4.1+

### LINKED ISSUE #263: make @sym/disp compatible with Octave 4.1+
The handle property `FormatSpacing` has been removed in the Octave 4.1+ development branch. Doug Stewart noticed this, see the mailing list thread starting [here](http://lists.gnu.org/archive/html/octave-maintainers/2015-09/msg00080.html).

The replacement is to use the internal function `__compactformat__()` which was added at the same time as the root handle property.

mtmiller:
Fixes #263.

### DIFF
--- a/inst/@sym/display.m
+++ b/inst/@sym/display.m
@@ -78,7 +78,7 @@ function display(x)
     unicode_dec = false;
   end
   display_snippet = sympref('snippet');
-  loose = strcmp(get(0,'FormatSpacing'), 'loose');
+  loose = ! __compactformat__ ();
 
   %% Get terminal width, mainly for snippets
   % works in matlab gui & -nodesktop but not the most up-to-date


cbm755:
ok thanks for this, I will merge.

But AFAICT from online documentation `get(0,'FormatSpacing')` is still how its done in Matlab, so this will need a further patch for Symbolic to keep working under matlab.

mtmiller:
Hmm, I didn't realize this is a Matlab compatibility thing, let's see if we get this property added again in Octave.

cbm755:
We'll I'm not completely sure: the source code you linked seemed suggests its deprecated in Matlab... [See code comments here](http://hg.savannah.gnu.org/hgweb/octave/rev/564e59f75ef8).  But I cannot find confirmation of that anywhere...

cbm755:
FTR, could revert this and ee0649d if Octave adds the `get(0,...)` approach again.

PR #283, opened 2016-01-06
like this, ex handle

latot:


inst/@sym/solve.m
```diff
@@ -111,49 +111,67 @@
...
-            '    d = sp.solve(eqs, symbols, dict=True)'
-            'else:'
-            '    d = sp.solve(eqs, dict=True)'
+	    'try:'
+            '	if len(symbols) > 0:'
+            '		d = sp.solve(eqs, symbols, dict=True)'
+            '	else:'
+            '		d = sp.solve(eqs, dict=True)'
+	    'except Exception as e:'
+	    '	return (str(e), "whatev")'
```
cbm755: Return something like `(False, str(e))`: if `flag` is false then `out` has the error message.

cbm755:
looking good so far, careful with your indenting in Python code.

cbm755:
btw, `inv.m` should do this too I think.

latot:
```
Return something like (False, str(e)): if flag is false then out has the error message.
```
Yea, that was an option, but i think we finally only start adding unnecesary lines while we can do this:
```
 +	if (flag)
 +		error(flag)
 +	end
```

cbm755:
CI tests are failing...

I cannot test on windows either :(

If you have cases that should give an error you can add `%!error <...>` tests for them

inst/@sym/solve.m
```diff
@@ -117,22 +117,30 @@
             '    else:'
             '        symbols.append(arg)'
             '#'
-            'if len(symbols) > 0:'
-            '    d = sp.solve(eqs, symbols, dict=True)'
-            'else:'
-            '    d = sp.solve(eqs, dict=True)'
+	    'try:'
+            '	if len(symbols) > 0:'
```
cbm755: needs to be four spaces indent I think

inst/@sym/solve.m
```diff
@@ -117,22 +117,30 @@
...
+            '		d = sp.solve(eqs, symbols, dict=True)'
```
cbm755: too many spaces indented.

inst/@sym/solve.m
```diff
@@ -117,22 +117,30 @@
...
+            '        return (0, sp.Matrix([r.popitem()[1] for r in d]))'
             '#'
             'if len(d) == 1:'
             '    d = d[0]'
-            'return d,' };
+            'return (0, d)' };
+
+    [flag, out] = python_cmd (cmd, varargin{:});
+
+	if (flag)
```
cbm755: if possible, please use spaces instead of tabs

cbm755:
you can run tests locally like `test @sym/solve` or `octsympy_tests` to run all of them.  Also, see the `continuous-integration` links to Travis CI below.

latot:
okis, how its tabs seems randon now only using spaces, i try tests and now all passes.

inst/@sym/solve.m
```diff
@@ -112,48 +112,64 @@
...
     cmd = { 'eqs = list(); symbols = list()'
             'for arg in _ins:'
-            '    if arg.is_Relational:'
-            '        eqs.append(arg)'
-            '    else:'
-            '        symbols.append(arg)'
+            ' if arg.is_Relational:'
+            '  eqs.append(arg)'
+            ' else:'
+            '  symbols.append(arg)'
```
cbm755: I don't think you changed these four lines...  This indenting was correct before your change: Python should use four spaces for indenting

inst/@sym/solve.m
```diff
@@ -112,48 +112,64 @@
...
             '#'
-            'if len(symbols) > 0:'
-            '    d = sp.solve(eqs, symbols, dict=True)'
-            'else:'
-            '    d = sp.solve(eqs, dict=True)'
+	    'try:'
+            ' if len(symbols) > 0:'
```
cbm755: as above, needs two more spaces

inst/@sym/solve.m
```diff
@@ -112,48 +112,64 @@
...
-            '        return sp.Matrix([r.popitem()[1] for r in d]),'
+            ' if len(d) == 1:'  % one variable, single solution
+            '  return (0, d[0].popitem()[1])'
+            ' else:'  % one variable, multiple solutions
+            '  return (0, sp.Matrix([r.popitem()[1] for r in d]))'
             '#'
             'if len(d) == 1:'
-            '    d = d[0]'
-            'return d,' };
+            ' d = d[0]'
```
cbm755: as above

inst/@sym/solve.m
cbm755: this is still a tab character

inst/@sym/solve.m
```diff
@@ -112,48 +112,64 @@
...
-            '    d.append(sp.Matrix([t[i] for t in solns]))'
-            'return d,' };
+            ' d.append(sp.Matrix([t[i] for t in solns]))'
+            'return (0, d)' };
+
+    [flag, out] = python_cmd (cmd, varargin{:});
+
+	if (flag)
+		error(out)
+	end
```
cbm755: still have tabs here

latot:
sorry, i think was only in the python lines

cbm755:
Great that tests pass!  This is looking better and better!

latot:
okis, now don't exist tabs!

cbm755:
great!  See other comments above too.

latot:
okis, maybe put in the wiki always use 2 spaces as code style?

cbm755:
Yes good idea re: wiki.

The indent style is 2 spaces octave, 4 spaces python.  This is common practice for each language.

latot:
Hi, talking about flag concept ,i check codes with this but its mixed up D:
some files use flag=1 to define an error and others 0, and when you init the flag var normally its with 0 (suppose not error), it too confuse.

so, to standarize, i propose you this:
    flag=0 > No Error
    flag=1 > Error

cbm755:
Ok, this is in.  I rebased your changes to a single commit: ada166b830055f27263c2b3008ae2794cea7586b.

Time time, I suggest doing your PR on a branch instead of "master": sympy has some good documentation about this kind of thing.  One branch per PR.

cbm755:
> sorry for this, but i'm very bad with some concepts, what means self-contained code?

Not sure what are you referring to, but yes not depending on external things sounds about right.

latot:
only to be sure, what happend with the flag concept?, while we will use the propose or not?
if we will use it, i think can help write part of it in the wiki, at least until this can be improve to python_cmd.m

Issue #287, opened 2016-01-08
non-deterministic behaviour in (new) ilaplace


cbm755:
CC @maprieto have you seen this?

I've narrowed down the problem to `L = ilaplace(exp(c*s))` (where c and s are `syms c s`)

Sometimes I get this:
````
pydebug: exp(c*s)
pydebug: ('r=', {a__: 1, b__: c*s})
pydebug: exp(c*s)
pydebug: ('rlist=', [1, c*s])
pydebug: c
pydebug: 1
L = (sym) DiracDelta(c + t)
````

Other times I get:
````
pydebug: exp(c*s)
pydebug: ('r=', {b__: c*s, a__: 1})
pydebug: exp(c*s)
pydebug: ('rlist=', [c*s, 1])
pydebug: 1/s
pydebug: c*s
L = (sym)

  ⎧                             │t│       │c│    
  ⎪           0             for │─│ < 1 ∨ │─│ < 1
  ⎪                             │c│       │t│    
  ⎪                                              
  ⎪        ⎛     │    ⅈ⋅π⎞                       
  ⎨╭─╮0, 1 ⎜1    │ t⋅ℯ   ⎟                       
  ⎪│╶┐     ⎜     │ ──────⎟                       
  ⎪╰─╯1, 1 ⎝   1 │   c   ⎠                       
  ⎪───────────────────────        otherwise      
  ⎪           t                                  
  ⎩                                              
````

One of the new tests fails sometimes but not others...   Looks to be an ordering assumption about lists.  I guess this `rlist=list(r.values())` is non-deterministic (on Python 3 anyway)...

cbm755:
@maprieto This is annoying me as its causing random test failures so I'm going to merge it.  Let me know if this is not the right solution and we can revert it.

maprieto:
Perfect! That's natural to use r[a_] and r[b_] (I can't understand why I'd need to use a list instead of the original dictionary..??)

cbm755:
no problem!  I think its a very common error when switching to Python 3!

Issue #164, opened 2014-12-05
Deprecating SymPy 0.7.5 support


cbm755:
No immediate hurry on this, but eventually we will stop expecting the test suite to pass on 0.7.5.  Here's a list of things we should do then.  Add things to this list, perhaps reference it in the code if some code needs changed.

- [x] horzcat/vertcat: these have tests commented out
- [x] function_handle: docs
- [x] function_handle: lots of tests here that silence a warning, could optionally simplify
- [x] function_handle: drop the various workarounds here (and in private/codegen?)
- [x] dsolve: probably will have an xtest or two
- [x] mod: xtest
- [x] findsymbols: delete workaround for Matrices not having `free_symbol`.
- [x] lambertw test

cbm755:
we have deprecated it, but not dropped all the little workarounds yet: that will happen in PR #321.  But everything mentioned here is either done either in master, or on the branch in PR #321.

PR #289, opened 2016-01-10
handle, flags, bye more tabs

latot:
Hi again, apply things, removed all tabs. pass al tests of changed files.

### DIFF
--- a/inst/@sym/function_handle.m
+++ b/inst/@sym/function_handle.m
@@ -125,10 +125,10 @@ function f = function_handle(varargin)
 
     if (~worked)
       if (strcmp(out, 'Language ''octave'' is not supported.'))
-	error('function_handle: your SymPy has no octave codegen, cannot workaround');
+        error('function_handle: your SymPy has no octave codegen, cannot workaround');
       else
-	ou
-	error('function_handle: Some other error from SymPy code gen?  file a bug!');
+        ou
+        error('function_handle: Some other error from SymPy code gen?  file a bug!');
       end
     end
     M.name = out{1}{1};
--- a/inst/@sym/function_handle.m
+++ b/inst/@sym/function_handle.m
@@ -127,7 +127,7 @@ function f = function_handle(varargin)
       if (strcmp(out, 'Language ''octave'' is not supported.'))
         error('function_handle: your SymPy has no octave codegen, cannot workaround');
       else
-        ou
+        out
         error('function_handle: Some other error from SymPy code gen?  file a bug!');
       end
     end
--- a/inst/@sym/horzcat.m
+++ b/inst/@sym/horzcat.m
@@ -116,7 +116,7 @@ end
 %! q = sym(ones(1, 0));
 %! assert (isequal ([v q], v))
 
-%!error <ShapeError: `self` and `rhs` must have the same number of rows.>
+%!error <ShapeError>
 %! % FIXME: clean-up when we drop 0.7.5 support (Issue #164)
 %! if (str2num(strrep(python_cmd ('return sp.__version__,'),'.',''))<=75)
 %!   disp('skipping: test passes on sympy >= 0.7.6')
--- a/inst/@sym/horzcat.m
+++ b/inst/@sym/horzcat.m
@@ -30,7 +30,8 @@ function h = horzcat(varargin)
 
   % special case for 0x0 but other empties should be checked for
   % compatibilty
-  cmd = { '_proc = []'
+  cmd = {
+          '_proc = []'
           'for i in _ins:'
           '    if i.is_Matrix:'
           '        if i.shape == (0, 0):'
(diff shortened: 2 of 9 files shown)


inst/@sym/function_handle.m
```diff
@@ -125,10 +125,10 @@
 
     if (~worked)
       if (strcmp(out, 'Language ''octave'' is not supported.'))
-	error('function_handle: your SymPy has no octave codegen, cannot workaround');
+        error('function_handle: your SymPy has no octave codegen, cannot workaround');
       else
-	ou
-	error('function_handle: Some other error from SymPy code gen?  file a bug!');
+        ou
```
cbm755: I guess I meant "out" here not "ou" :) can you change it please?

inst/@sym/horzcat.m
```diff
@@ -117,7 +116,7 @@
 %! q = sym(ones(1, 0));
 %! assert (isequal ([v q], v))
 
-%!error <ShapeError>
+%!error
```
cbm755: Can we keep some <text> here?  Does it not pass with `<ShapeError>` here?  I thought it would as the error should still include the words `ShapeError`

inst/@sym/inv.m
```diff
@@ -51,30 +51,21 @@
 
 function z = inv(x)
 
-  [n,m] = size(x);
-  if n ~= m
-    error('matrix is not square')
-  end
-
```
cbm755: I like this change.

But can use also add a test that we get an error if you try to invert a non-square matrix (my fault: I should've added this originally)

inst/octsympy_tests.m
```diff
@@ -226,7 +226,7 @@ function print_pass_fail (p, n, xf, sk)
         dn += n;
         dxf += xf;
         dsk += sk;
-	FWT{end+1} = f;
+        FWT{end+1} = f;
```
cbm755: In future, send these kinds of white-space only changes in a different commit (I don't mind if its same PR but use different commit).

Its just my preference, no need to redo anything!

cbm755:
One test failure is not your fault (its #287).  But you have a failing `doctest` in `inv.m` (the output of the `@example` blocks is checked.)

cbm755:
'fix typo" would  be a better commit message than "ups" (again no need to change, just for next time).

latot:
mm, let my search about the errors, because if sympy send more that one message i think will be difficult

latot:
mm, the problem seems python send various messages and octave only handle the last one, in horzcat should be :
```<sym/horzcat method failed>```
but i don't think have many sense this message...

**Comment by cbm755** [comment]
hmmm, that is unfortunate if user gets *less informative* error message.  Maybe this approach is still not quite right.

**Comment by latot** [comment]
but this is only in the test phase, if the user run the command will get all the error message, well we ca keep the shape message in the comments of the test if you like it

**Comment by cbm755** [comment]
I just tested it with `%!error <ShapeError>` and it works fine for me...?
````
>> test @sym/horzcat 
PASSES 7 out of 7 tests
````

**Comment by cbm755** [comment]
So yes keep it

**Comment by latot** [comment]
sadly here seems, don't like it:
```
test @sym/horzcat
***** error <ShapeError>
 % FIXME: clean-up when we drop 0.7.5 support (Issue #164)
 if (str2num(strrep(python_cmd ('return sp.__version__,'),'.',''))<=75)
   disp('skipping: test passes on sympy >= 0.7.6')
   error('ShapeError')   % pass the test with correct error
 else
   v = [sym(1) sym(2)];
   q = sym(ones(3, 0));
   w = [v q];
 end
!!!!! error failed.
Expected <ShapeError>, but got <sym/horzcat method failed>
```

**Comment by cbm755** [comment]
hmmm, what is different?  I have Octave 4.0.0 and SymPy 0.7.7-dev.

**Comment by latot** [comment]
mm, here, Linux, Octave 4 git, sympy 0.7.6, let me try upgrade sympy, the top i can is 0.7.6-r1

**Comment by cbm755** [comment]
I just tried sympy 0.7.6, fine there too...

**Comment by cbm755** [comment]
I'm on Fedora 23.  Maybe it is Octave hg tip versus 4.0?

**Comment by cbm755** [comment]
ahh, yes it fails for me with a more recent octave...  Maybe its octave bug?

**Comment by latot** [comment]
thats it!, i try now with normal octave and there works, so its posible we can have this problem in the future, or is a hg octave

**Comment by cbm755** [comment]
So let's keep the `<blah>` for now, and worry about this later: many other tests will fail too I think.

File: inst/@sym/lhs.m [hunk 5]
```diff
@@ -81,18 +76,18 @@
...
+%!error
 %! syms x
 %! rhs(x)
 
-%!error <lhs:.* entries have no>
+%!error
 %! lhs(sym(true))
 
-%!error <lhs:.* entries have no>
+%!error
```
cbm755: these will have to change slightly

cbm755:
When did you build Octave from hg?  I will report to mailing list, but my build is a bit old...  Nice if I can say its recent...

latot:
some days ago, i can build it again and test it

latot:
test upgrades, and now all pass

inst/@sym/vertcat.m
```diff
@@ -124,7 +123,7 @@
 %! q = sym(ones(0, 2));
 %! assert (isequal ([v; q], v))
 
-%!error <ShapeError>
+%!error <ShapeError: `self` and `bott` must have the same number of columns.>
```
cbm755: Ok.

Just FYI for next time: its not necessary to quote the exact error: if SymPy makes a minor change then I don't need the test to fail.  The idea is it should fail with ShapeError rather than e.g., a "syntax error".

cbm755:
some tests are failing :(

Also:

> But you have a failing doctest in inv.m (the output of the @example blocks is checked.)

latot:
okis, changing...

latot:
D: i don't have idea why fails the doctest (i never used before)

cbm755:
No problem, I fixed this with 1cc36127a0c6ea0ba66fe0e0bd49ae104e9b1290.

Thanks for this.

cbm755:
I emailed octave-maintainers about the error bug.  TODO: If I file a bug I will link it here.

Issue #268, opened 2015-12-04
incompatible with Octave 3.8


mtmiller:
Due to the whole compact spacing compatibility mess, it seems that disp won't work with Octave 3.8 (the `__compactspacing__` function was added in 4.0). Installing the package with 3.8 works but can't show any variables. The logic could be further fixed to handle both Octave ways of checking whether compact formatting is on or off, or simply add a dependency on Octave >= 4.0 in the DESCRIPTION file.

cbm755:
thanks Mike.  My tests are not run regularly on 3.8.1; I guess I'll drop support for it soon.

Assumptions seem to be broken too, I'll see if I can fix that up then do a new release.

PR #299, opened 2016-01-12
Deprecate Octave 3.8

### LINKED ISSUE #271: deprecate Octave 3.8 support
Perhaps after next Ubuntu LTS is ok?  16.04?  

or maybe sooner, I'm not running tests on 3.8 any more...

There is still some 3.6 cruft around too, clean that up as well.

cbm755:
Fixes #271.

I probably won't merge this until the next ubuntu LTS is out: lots of people are still on Octave 3.8 because of 14.04 LTS.

Also, we can currently run CI tests again for 3.8 so not too much bit-rot (I hope avoiding regressions like #268).

TODO list:
- [x] Add minimum octave version to DESCRIPTION file.


mtmiller:
Does Travis support Ubuntu 16.04 yet? That's probably the primary reason not to drop support for Octave 3.8 yet.

You already have a minimum version of SymPy that is newer than comes with Ubuntu 14.04, requiring users to install a new SymPy is easier, but not much different from asking them to install a new version of Octave (and I can provide the PPA to install 4.0.x on 14.04).

cbm755:
thanks!  I already use someone's PPA for Travis to get Octave 4.  I'll probably go ahead with this.

mtmiller:
I see, I should have looked at the config file first :wink: 

The `ppa:octave/stable` PPA is the one I maintain.

So yeah, you are already handling 14.04 fine with an updated Octave and SymPy, so I don't know that there's much need to hold on to Octave 3.8 support if it gets in the way.

cbm755:
merged to `dev_v250` branch.

PR #300, opened 2016-01-12
Deprecate SymPy 0.7.5

cbm755:
After this, we could not expect tests to pass on 0.7.5 and special hacks to support 0.7.5 could be dropped.


cbm755:
Note: rebased

PR #307, opened 2016-01-15
Have `double(zoo)` return `Inf + Infi`

cbm755:
> Why?  This will remind anyone converting `sym('zoo')` that it is
> complex infinity.

### DIFF
--- a/inst/@sym/double.m
+++ b/inst/@sym/double.m
@@ -78,8 +78,8 @@ function y = double(x)
   end
 
   cmd = { '(x,) = _ins'
-          'if x == zoo:'
-          '    return (float(sp.oo), 0.0)'
+          'if x == zoo:'  % zoo -> Inf + Infi
+          '    return (float(sp.oo), float(sp.oo))'
           'if x == nan:'
           '    return (float(nan), 0.0)'
           'x = complex(x)'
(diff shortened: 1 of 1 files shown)


cbm755:
Not sure if this is a good idea or not.  But its here so I don't lose it.

latot:
to test all this things:
```
%!test
%! z = sym('zoo');
%! assert( isequal( double((i + 1)*z), Inf + Inf*i))
%! assert( isequal( double(zoo), Inf))
%! assert( isequal( double(i*zoo), NaN + Inf*i))
%! assert( isequal( double(NaN*zoo), NaN))
%! assert( isequal( double((NaN + i)*zoo), NaN + Inf*i))
%! assert( isequal( double((NaN + 1)*zoo), NaN))
```

cbm755:
I'm not convinced those are good tests.  Some are just wrong (e.g., `i*z` should be `zoo`), and at least one of them contradicts this PR.

Anyway, this is very old, still seems like a reasonable idea to me so I'll merge it after 2.5.0.

latot:
mm, only as a record, the tests i put above is the octave behavior.

cbm755:
There is no "octave behaviour": Octave does not have the complex point at infinity.

cbm755:
Those might be what Octave does for `(i + 1)*inf` but I don't see how that is relevant here.

latot:
don't is exactly relevant, only as a reference, i mean we will don't have the same behavior from sympy and octave.

cbm755:
I still don't know what you mean.  Please give a concrete example.  Is there something not tested for how `sym(inf)` behaves with `double`?  Is so please give example, or file a bug because that is different from this PR.

There is nothing to compare to for `sym('zoo')` because floating point only has "inf" not complex point at infinity.

For reference, Sympy gives an error for `Float(zoo)`.

Issue #303, opened 2016-01-14
Improve Intervals


latot:
Hi, well i was thinking this can help a lot in octave, can we add it here?
if you are agree what strcture do you recomend?

But probably we will be need add classes from here:
sympy/sets/sets.py

cbm755:
Sounds good.   Here's some things that would be nice to have:

````
>> evalpy('x = Interval(2,4)')
pydebug: Treating unexpected SymPy obj as scalar: <class 'sympy.sets.sets.Interval'>
x = (sym) [2, 4]
>> evalpy('y = Interval(1,10)')
y = (sym) [1, 10]
>> evalpy('z = Interval(9,10*S(pi)/3-1)')
z = (sym)

  ⎡        10⋅π⎤
  ⎢9, -1 + ────⎥
  ⎣         3  ⎦

>> evalpy('a = y.intersect(z)', y, z)
a = (sym)

  ⎡        10⋅π⎤
  ⎢9, -1 + ────⎥
  ⎣         3  ⎦
````

Also
````
>>> pprint(x.union(z))
         ⎡        10⋅π⎤
[2, 3] ∪ ⎢9, -1 + ────⎥
         ⎣         3  ⎦
````

latot:
how structure this? we can't use the word union or similar, or use interal_function like interval_union?

cbm755:
I wonder if `@sym/interval.m` is the right way to make such things?  A quick search suggests Matlab's SMT does not expose intervals.

However, Octave has an "interval" package (which does floating point interval arithmetic: mostly unrelated).  Need to make sure they don't fight...

cbm755:
>> how structure this?

maybe `setdiff.m`, `setxor` `union.m` etc?  And `@sym/interval` to make them...
````
x = interval(sym(2), sym(4))
````

latot:
mm, but the union, intersect, etc, have practically others functions and behaviors...

cbm755:
> or you mean check the class type in union.m etc..

Yes.  But I'm not sure.

I even thought about extending `colon`: i.e., see that `y = 2:x` would make an `Interval`.   But I now think this is a bad idea.

latot:
bad idea improve intervals or change colon on that way?

latot:
i already check the interval package, i seems don't have any interval.m file

cbm755:
bad idea to change colon in that way.

cbm755:
[...]

PR #311, opened 2016-01-16
set opertation

latot:
Working...

Fixes https://github.com/cbm755/octsympy/issues/303

i don't know how add classes for intervals D:


inst/@sym/intervals.m
```diff
@@ -0,0 +1,73 @@
+%% Copyright (C) 2014 Colin B. Macdonald
```
cbm755: 2016 <Your name here>

cbm755:
Cool.  Do we need to make sets of intervals?

I was thinking something like this:
````
I1 = interval(sym(1), 2)
I2 = interval(-6, -sym(pi))
I3 = union(I1, I2)
⇒ I3 = (sym) [-6  π] U [1 2]
````

**Comment by latot** [comment]
uu, i was thinking the same :D 
bout union and similars, check var to var i think will be a waste of time and resource so,  can we use the normal calls of it?, like:
union(varargin, 'intervals')
union(I1, I2, I3, 'intervals')

File: inst/@sym/intervals.m (hunk 1 above)
Comment (latot) [inline inst/@sym/intervals.m]: do you mean?
```
+%% Copyright (C) 2016 Colin B. Macdonald and Lagu
```

**Comment by latot** [comment]
okis, sadly  can't test the interval part:
```
p=intersect(a, b, 'intervals')
error: 'intervals' undefined near line 44 column 16
error: called from
    intersect at line 44 column 14
```
https://github.com/cbm755/octsympy/issues/310

**Comment by cbm755** [comment]
Don't worry about the trivial overhead. Users will prefer to just `union(I1, I2)` I think.

Just your name on copyright header.

**Comment by latot** [comment]
pending:

https://github.com/cbm755/octsympy/issues/301
And add tests to intervals.

**Comment by cbm755** [comment]
Sorry, but I still don't see why we want to make more than interval at a time.  Making one is complicated enough.

I've sent you a pull request which changes it to `@sym/interval`.  What do you think?

**Comment by cbm755** [comment]
There seems to be a SymPy bug about ascii rendering of unions, I'll try to fix it...

**Comment by latot** [comment]
when i try make intervals i think in the concept of the function syms and sym, the first to make a lot of syms faster and sorted, the second transform one type to sym, and the multidimensional functions.
i think its necessary some function to can do something like this:
```
varargin=intervals(varargin)
```
or 
```
varargin=intervals(varargin{:})
```
sadly if we wanna keep multidimensional we can't use only interval, we can't know the size of the dimentions because its decision of the user so i think can help i we use both, interval and intervals, but i would recommend put both in inst/ instead of inst/@sym, intervals need works with varargin so only works in inst, and interval in @sym only will works with interval(1, sym(2)), and how don't have file collisions i think can works.

**Comment by cbm755** [comment]
>  the concept of the function syms and sym

I don't think its worth it.  I wrote down [my reasoning on the wiki](https://github.com/cbm755/octsympy/wiki/Development-Guide#adding-new-functions--designing-interfaces--etc)

> we wanna keep multidimensional

Can you give me an example (in SymPy maybe) of what one of these is? `srepr(A)` where `A` is one of these.

**Comment by latot** [comment]
its about this: https://github.com/cbm755/octsympy/issues/295
actually sympy don't support multimentional, basically, thinking in the future, although worthwhile, the code is done, works and keep comatibility with sympy, so its an option to implement, or maybe use interval as main and move intervals to private, personally i like keep intervals access to the user so there we can use it to make our own functions when we need it (out of the package), obs i'm trying avoid diverge from sympy, and keep octave style.

**Comment by cbm755** [comment]
I think I understand.  But I don't want to merge `intervals` at this time.

**Comment by latot** [comment]
okis, actually union and intersect use this:
```
varargin=intervals(varargin)
```
so to keep it working its posible move intervals to @sym/private/cell_to_intervals.m? if not, i'll need convert manually all varargin to intervals....

**Comment by cbm755** [comment]
I don't think those functions will need `intervals`.  Inside the python code, they might have to do something different depending on whether the input is `Interval`, `FiniteSet` or `Matrix`.

**Comment by latot** [comment]
okis i'll change it

**Comment by cbm755** [comment]
See also #314

**Comment by cbm755** [comment]
> posible move interval to inst instead of `inst/@sym`?

Maybe, e.g., if we need `interval(1, 2)` to work...  but otherwise I'd rather not: to keep the namespace as clean as possible.

**Comment by latot** [comment]
okidoki.

**Comment by cbm755** [comment]
How about this?
````
function r = union(a, b)

  cmd = { 'a, b = _ins'
          'if isinstance(a, sp.Set) or isinstance(b, sp.Set):'
          '    return sp.Union(a, b),'
          '# special case for vector input'
          'A = sp.FiniteSet(*(list(a) if a.is_Matrix else [a])'
          'B = sp.FiniteSet(*(list(b) if b.is_Matrix else [b])'
          'C = sp.Union(A, B)'
          'return sp.Matrix([[list(C)]]),' };

  r = python_cmd(cmd, sym(a), sym(b));
````

**Comment by cbm755** [comment]
I committed 0bdcc7db3e1802ee37f97075c846494f07bb779b to master would should get rid of the warnings with `Interval` and `FiniteSet` objects.

**Comment by latot** [comment]
hi sadly i think we shouldn't use that code, first Union only works if both are intervals, if you insert arrays, or one array and one interval we will get errors, second, the behavior is too different of octave, will break compatibility with other functions (and don't pass the tests), with this in mind i try make the function with this in count, help of Union:
```
 -- Function File: C = union (A, B)
 -- Function File: C = union (A, B, "rows")
 -- Function File: [C, IA, IB] = union (...)
```
in the function first enable 
```
 -- Function File: C = union (A, B, ..., "intervals")
```
but it recognizes if one of the arguments are intervals and call that section, if you send normal inputs, arrays, will return the normal octave output but with syms. this is applied in union and intersect.

**Comment by cbm755** [comment]
>  so we can't skip the "detection of intervals" phase, and if is executed we need check every variable and transform it to a interval if is an array or number,

I disagree.  I think these two lines from my code above deal with `Interval`, `FiniteSet`or anything else that is a subclass of `Set`.
````
          'if isinstance(a, sp.Set) or isinstance(b, sp.Set):'
          '    return sp.Union(a, b),'
`````
(this will probably raise an erorr if you try to union an Interval with a Matrix: good)

**Comment by cbm755** [comment]
I'm sorry, had extra brackets, I think I should've said `return sp.Matrix([list(C)]),`

**Comment by cbm755** [comment]
3932fc5496bf3eac294f78df8d543be6c20f5178 fixes the `KeyError: 'only unicode'` error with union.

You can cherry-pick first 0bdcc7db3e1802ee37f97075c846494f07bb779b and then 3932fc5496bf3eac294f78df8d543be6c20f5178 to get these fixes into your branch.  (or merge from `cbm755/master`) 

Hint: when I do stuff barely understand in git, I first `git checkout -b tmp`: that way if I break it, I can just go back to the `intervals` branch.  OTOH, if it works, then I do the same command on my `intervals` branch.

**Comment by latot** [comment]
Hi, seems you prefer write all in python, so here is.
a lot of lines are  for detect the type of var to transform it to intervals.

**Comment by latot** [comment]
yay now works.
```
>> union(sym([1]), [sym(15)/10, 2], [6, 10], 'intervals')
ans = (sym) {1} ∪ [3/2, 2] ∪ [6, 10]
```

**Comment by cbm755** [comment]
Can you give an *actual example* that this code does, that my **6-line** code does not do?

**Comment by cbm755** [comment]
For example:

> `union(sym([1]), [sym(15)/10, 2], [6, 10], 'intervals')`

I think this should be written as:
````
A = union(interval(sym(15)/10), interval(6,10))
B = union(A, interval(sym(1),1)
````
Or if you want to support multiple inputs:
````
B = union(interval(sym(15)/10), interval(6,10), interval(sym(1),1))
````
Then there is no need for `'intervals'` flag.  And no need for all this fragile input checking.

**Comment by latot** [comment]
np, its supported (yours examples should works), if you like its posible remove it, but i prefer thinking in performance and octave syntax of union/intersect keep it.

**Comment by cbm755** [comment]
> if you like its posible remove it, 

yes please!

> contain better the intervals

I don't think its better.  Example?

**Comment by latot** [comment]
ex, we have a list of intervals of normal code of octave (out of symbolic) like this:
```
a=[[1, 2], [4, 5], [8, 10], [50, 54]];
```
so if we wanna apply union there we can just do:
```
union(a{:}, 'intervals')
```
personally simplify a lot this.

**Comment by latot** [comment]
Fixes https://github.com/cbm755/octsympy/issues/147

**Comment by cbm755** [comment]
> ex, we have a list of intervals of normal code of octave (out of symbolic) like this:
> a={[1, 2], [4, 5], [8, 10], [50, 54].....};
> so if we wanna apply union there we can just do:
> union(a{:}, 'intervals')

That will not work with `@sym/union` because no input is a `sym`.

> personally simplify a lot this, and without intervals don't are to much ways to do easly this.

Sorry but I disagree.  I expect it can be done (by user herself, not by us) using `cellfun`.  I just don't think its worth us having complicated code.

**Comment by cbm755** [comment]
> little question, [5, 7] should be equal to [7, 5]?

`interval(7,5)` should give `EmptySet`.

**Comment by cbm755** [comment]
Here is a possible suggestion to move this forward: submit a separate PR for `interval.m`: we'll get that reviewed and merged, then come back to this PR for the set operation.

For example, I don't think your `interval.m` is passing tests.

**Comment by latot** [comment]
hi, my bad, i don't write correctly the example:
```
a={[1, 2], [4, 5], [8, 10], [50, 54]};
union(sym(a){:}, 'intervals')
ans = (sym) [1, 2] ∪ [4, 5] ∪ [8, 10] ∪ [50, 54]
```
if you incert this in a nomal function you only will get the union of elements, but we want union of intervals, that's the utility.

Note, i try use cellfun but i can't put it works with interval...

latot:
here only check one python script, the other are basically the same.

cbm755:
Maybe some feature will not work on SymPy 0.7.6 then.  That's ok.

cbm755:
Its great that tests pass but I don't feel I can merge this.

Here's a suggestion: let's forget about array/vector input.  SLet's make these functions work with `Interval` and `FiniteSet` only.  Make a new branch, leaving this branch alone and doing a new PR.  

Hopefully we only need two or three lines of python code per .m file.

(just leave this PR: we can always come back to it later).

Issue #316, opened 2016-01-19
with new error handling, we lose the stack trace


cbm755:
Maybe we should return the stacktrace as a string...

cbm755:
its sort of back as of #357.  We `raise` in python so the errors appear on the stderr (and thus they are still lost on Windows with popen2, #82).

In principle, we could pass the string of the stracktrace back to Octave: I doubt if this is worth doing at the moment, but I'll leave this open for now.

genuinelucifer:
@cbm755 I saw that raise errors are being handled as per line numbers. Is passing the string still required? If so, I could work on this.. Or any other issue you would suggest?

cbm755:
I think this is low priority  How's your systems programming?  Want to take a look at #82 in core Octave?

genuinelucifer:
Hi @cbm755!
I saw a few of the files related to issue 82. I will start work on it soon.
But, (as I mentioned in my GSoC proposal)  my exams have started and will
be going on till the end of this month. Hence, I would like to start
working on it from 1st May.

cbm755:
TODO: double-check that stderr is shown on screen on Windows with Octave > 4.0.2 (e.g., by raising an exception, we should see Python stacktrace like on GNU/Linux).

cbm755:
Hmm, I cannot even see stacktraces on Windows or GNU/Linux anymore, since 2.2.4: have not looked further yet.

PR #317, opened 2016-01-19
add number line error

latot:
add number line error, i think this can help, maybe instead of all backtrace or while https://github.com/cbm755/octsympy/issues/316


cbm755:
Neat idea!  Is that the lineno within the lines of our string `cmd`?  I will try to test later...

latot:
that right, should the line error in cmd.

latot:
its posible merge this?

cbm755:
One problem that some of the ipc stuff puts more lines into cmd...
````
>> sympref ipc system
>> python_cmd('raise')
OctSymPy v2.3.0-dev: this is free software without warranty, see source.
You are using the slower system() communications with SymPy.
Warning: this will be *SLOW*.  Every round-trip involves executing a
new python process and many operations involve several round-trips.
error: TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType (line 272)
````

**Comment by cbm755** [comment]
What happens if this command fails?  Could that ever happen?  Maybe that needs a `try:` `except: -1` logic around it...
`sys.exc_info()[-1].tb_lineno - 1`

Also, I'll send a pull request to put the lineno in a new list entry...

**Comment by latot** [comment]
how we can know what ipc option we are using? if don't exist an option should be add an sympref option

**Comment by cbm755** [comment]
````
>> s = sympref ipc
s = default
````
But default is platform dependent (e.g., on Octave its `popen2`, whereas on Matlab unix it is `system` or Matlab windows its `sysoneline`).

Maybe the message should just say: "Error might be near line xy (assuming you are using popen2 communication)."  Its probably not worth trying too hard here: pytave will fix all this for us someday ;-)

**Comment by latot** [comment]
D:
```
>> s = sympref ipc
parse error:

  syntax error

>
```

**Comment by cbm755** [comment]
sorry I meant `s = sympref('ipc')`

**Comment by latot** [comment]
i think its enough for now, at least until pytave

**Comment by cbm755** [comment]
:-)  we'd get some kind of award for worst use of "unnamed magic numbers" ;-)

I suggest:
````

  % Careful: fix this constant if you change the code below.
  % Test with "python_cmd('raise')" which should say "line 1".
  LinesBeforeCmdBlock = 3;

  cmd = indent_lines(cmd, 8);
  ...

fprintf('Python has raised an error.  It may have occurred near line %d', A{3} - LinesBeforeCmdBlock - 1);
fprintf('of the python code block (assuming popen2 ipc).\n');
````
(and ignore the other cases, the 275 will bitrot too quickly.

**Comment by latot** [comment]
okidoki :D

**Comment by cbm755** [comment]
I'm still nervous that a further exception can be thrown by this code.

E.g., in IPython:
````
In [27]: raise ValueError(42)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-27-12fa2e30513e> in <module>()
----> 1 raise ValueError(42)

ValueError: 42

In [28]: sys.exc_info()
Out[28]: (None, None, None)
````
Note that `None` will lead to trouble!

But OTOH, I see other people recommending this approach on stack overflow...

I'll merge for now.

**Comment by latot** [comment]
here at least works :D 

```
python_cmd('raise ValueError(42)')
Python has raised an error. It may have occurred near line 1 of the python code block (assuming popen2 ipc).
error: ValueError: 42
error: called from
    python_cmd at line 155 column 5
```

PR #318, opened 2016-01-19
add interval only

latot:
To pass the tests we will need https://github.com/cbm755/octsympy/pull/311

### DIFF
--- a/inst/@sym/interval.m
+++ b/inst/@sym/interval.m
@@ -39,7 +39,7 @@
 %% @example
 %% @group
 %% interval(sym(1), 1)
-%%   @result{} (sym) {1}
+%%   @result{} (sym) @{1@}
 %% interval(sym(2), 1)
 %%   @result{} (sym) ∅
 %% @end group
@@ -67,3 +67,11 @@ end
 %!test
 %! a = interval(sym(1), 2);
 %! assert (isa (a, 'sym'))
+
+%!test
+%! % some set subtraction
+%! a = interval(sym(0), 4);
+%! b = interval(sym(0), 1);
+%! c = interval(sym(1), 4, true);
+%! q = a - b;
+%! assert (isequal( q, c))
--- a/inst/@sym/interval.m
+++ b/inst/@sym/interval.m
@@ -1,0 +1,85 @@
+%% Copyright (C) 2016 Colin B. Macdonald and Lagu
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


inst/@sym/interval.m
```diff
@@ -0,0 +1,85 @@
...
+
+function I = interval(varargin)
+
+  if (nargin < 1 || nargin > 2)
+    print_usage();
+  end
+
+  varargin = sym(varargin);
+
+  % FIXME: Maybe in the future SymPy support more than 2 dimensions.
```
cbm755: Drop this note: perhaps you mean `ProductSet` and we'll do that elsewhere.

inst/@sym/interval.m
```diff
@@ -0,0 +1,85 @@
...
+
+  if nargin == 1
+
+    I = python_cmd ('return Interval(_ins[0], _ins[0]),', varargin{:});
+
+  else
```
cbm755: I don't think we need to support `interval(2)`: that sounds confusing.  we may add `infiniteset` or something later which can be used to make sets of just one thing.

inst/@sym/interval.m
```diff
@@ -0,0 +1,85 @@
...
+%%
+%% @seealso{union, intersect, setdiff, unique, ismember}
+%% @end deftypefn
+
+%% Author: Colin B. Macdonald
+%% Keywords: symbolic
+
+function I = interval(varargin)
+
+  if (nargin < 1 || nargin > 2)
```
cbm755: this doesn't look right.  And the doctests fail, maybe this is why.

inst/@sym/interval.m
latot: mm, i think can help to works because the out of some outputs are
```
c = (sym) {2}
```
so at least to me make sense can write ```interval(2)```, i'm thinking if we delete this part the user should start handling this from octave while its easly from here. and it feels the concepts don't match spliting this.

inst/@sym/interval.m
latot: only to be sure, i mean like this ```Interval(1, 2, 3)``` its posible with product set?

latot:
problem in doctest:
doctest interpret don't intepre {} in examples:
```
%%   @result{} (sym) {1}
```
is interpreted like
```
%%   @result{} (sym) 1
```

inst/@sym/interval.m
latot: fixed, my bad in a previus commit.

inst/@sym/interval.m
cbm755: > np forget it :D 

gladly!  `Interval(1,2,3)` is nonsense :)

cbm755:
Ok I'll look into that...

I don't like `interval(sym(2))`.  Let's not do that.

latot:
okidoki

cbm755:
I sent you a pull request to fix the doctest, then I'll merge this.

latot:
oks

cbm755:
great!

PR #320, opened 2016-01-20
Version

latot:
i'll need help with this:
```
@sym/function_handle.m:        if (str2num(strrep(python_cmd ('return sp.__version__,'),'.',''))<=75 ...
```

### DIFF
--- a/inst/@sym/isinf.m
+++ b/inst/@sym/isinf.m
@@ -37,7 +37,7 @@ function r = isinf(x)
   if isscalar(x)
 
     cmd = { 'd = _ins[0]'
-            'if sympy.__version__ == "0.7.5":'
+            'if Version(spver) < Version("0.7.6"):'
             '    d = d.is_unbounded'
             'else:'
             '    d = d.is_infinite'
--- a/inst/@sym/min.m
+++ b/inst/@sym/min.m
@@ -233,9 +233,6 @@ end
 
 %!test
 %! % empty columns
-%! if (str2num(strrep(python_cmd ('return sp.__version__,'),'.',''))<=75)
-%!   disp('skipping: SymPy 0.7.5 is too old')  % clean up in Issue #164
-%! else
 %! A = sym(zeros(0, 4));
 %! [m, I] =  min(A, [], 1);
 %! assert (isequal (size(m), [0 4]))
@@ -243,13 +240,9 @@ end
 %! [m, I] =  max(A, [], 1);
 %! assert (isequal (size(m), [0 4]))
 %! assert (isequal (size(I), [0 4]))
-%! end
 
 %!test
 %! % empty rows
-%! if (str2num(strrep(python_cmd ('return sp.__version__,'),'.',''))<=75)
-%!   disp('skipping: SymPy 0.7.5 is too old')  % clean up in Issue #164
-%! else
 %! A = sym(zeros(3, 0));
 %! [m, I] =  min(A, [], 2);
 %! assert (isequal (size(m), [3 0]))
@@ -257,7 +250,6 @@ end
 %! [m, I] =  max(A, [], 2);
 %! assert (isequal (size(m), [3 0]))
 %! assert (isequal (size(I), [3 0]))
-%! end
 
 %!test
 %! % another empty case
--- a/inst/@sym/mpower.m
+++ b/inst/@sym/mpower.m
@@ -72,9 +72,10 @@ function z = mpower(x, y)
   end
 
   cmd = { 'x, y = _ins'
-          'if Version(spver) >= Version("0.7.7.dev") or not(x.is_Matrix and not y.is_Matrix):'
+          'if x.is_Matrix and not y.is_Matrix:'
+          '    return sympy.MatPow(x, y).doit(),'
+          'else:'
           '    return x**y,'
-          'return sympy.MatPow(x, y).doit(),'
         };
 
   z = python_cmd (cmd, sym(x), sym(y));
--- a/inst/@sym/mpower.m
+++ b/inst/@sym/mpower.m
@@ -72,7 +72,7 @@ function z = mpower(x, y)
   end
 
   cmd = { 'x, y = _ins'
-          'if Version(sv) >= Version("0.7.7.dev"):'
+          'if Version(spver) >= Version("0.7.7.dev"):'
           '    return x**y,'
           'if x.is_Matrix and not y.is_Matrix:'
           '    return sympy.MatPow(x, y).doit(),'
@@ -114,7 +114,7 @@ end
 
 %!test
 %! % non-integer power
-%! if (python_cmd ('return Version(sv) < Version("0.7.7.dev"),'))
+%! if (python_cmd ('return Version(spver) < Version("0.7.7.dev"),'))
 %!   disp('skipping known failure b/c SymPy <= 0.7.6.x')
 %! else
 %! A = sym([1 2; 0 3]);
--- a/inst/@sym/mpower.m
+++ b/inst/@sym/mpower.m
@@ -72,6 +72,8 @@ function z = mpower(x, y)
   end
 
   cmd = { 'x, y = _ins'
+          'if Version(sv) >= Version("0.7.7.dev"):'
+          '    return x**y,'
           'if x.is_Matrix and not y.is_Matrix:'
           '    return sympy.MatPow(x, y).doit(),'
           'else:'
--- a/inst/@sym/mpower.m
+++ b/inst/@sym/mpower.m
@@ -72,12 +72,9 @@ function z = mpower(x, y)
   end
 
   cmd = { 'x, y = _ins'
-          'if Version(spver) >= Version("0.7.7.dev"):'
+          'if Version(spver) >= Version("0.7.7.dev") or not(x.is_Matrix and not y.is_Matrix):'
           '    return x**y,'
-          'if x.is_Matrix and not y.is_Matrix:'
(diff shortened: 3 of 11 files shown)


inst/@sym/isinf.m
```diff
@@ -37,16 +37,11 @@
   if isscalar(x)
 
     cmd = { 'd = _ins[0]'
-            'if sympy.__version__ == "0.7.5":'
-            '    d = d.is_unbounded'
-            'else:'
-            '    d = d.is_infinite'
-            'if d == True:'
```
cbm755: not yet please,  There is a `drop_sympy075` branch that does this sort of thing but I wasn't going to merge it yet

inst/@sym/min.m
```diff
@@ -233,9 +233,6 @@
 
 %!test
 %! % empty columns
-%! if (str2num(strrep(python_cmd ('return sp.__version__,'),'.',''))<=75)
-%!   disp('skipping: SymPy 0.7.5 is too old')  % clean up in Issue #164
-%! else
```
cbm755: yes, thanks!

inst/@sym/mpower.m
```diff
@@ -72,6 +72,8 @@
   end
 
   cmd = { 'x, y = _ins'
+          'if Version(sv) > Version("0.7.7.dev"):'
+          '    return x**y,'
```
cbm755: I'm guessing this leads to test failures.  Please do a different PR as it might get complicated for MatrixExpr...

inst/findsymbols.m
```diff
@@ -62,15 +62,7 @@
...
-            'if sympy.__version__ == "0.7.5":'   % deprecate with Issue #164
-            '    if not x.is_Matrix:'
-            '        s = x.free_symbols'
-            '    else:'
-            '        s = set()'
-            '        for i in x.values():'
-            '            s = s.union(i.free_symbols)'
-            'else:'
-            '    s = x.free_symbols'
+            's = x.free_symbols'
```
cbm755: As above, leave this change for drop_sympy075 branch.

inst/private/python_header.py
```diff
@@ -23,6 +23,7 @@ def myerr(e):
 try:
     import sympy
     import sympy as sp
+    from sympy import __version__ as sv
```
cbm755: maybe `as spver`?  or `as sympyver`?

Then `Version(sympyver) >= `... is also more readable.

cbm755:
this looks like a great idea.  See comments above.

Also, can you confirm this `Version` command speaks pep0440?  (i.e., does right thing with `-dev`?)

cbm755:
> i'll need help with this: @sym/function_handle.m:   

Probably can be a version check for SymPy < 0.7.6.  Or just leave this alone, it will get deleted in the drop_sympy075 branch.

inst/@sym/mpower.m
latot: ups my bad >=
i already test this on 0.7.7.dev

latot:
no idea why is failing, actually have this line
```
from packaging.version import Version
```

cbm755:
Where does this module come from?  Perhaps its not part of the core Python install.

I don't want to increase our dependencies beyond core Python and SymPy...

You could check how SymPy itself deals with version numbers?

cbm755:
> AttributeError: 'MatPow' object has no attribute 'values'

Note I suggested above doing your mpower changes as a separate PR...

latot:
sadly i try back to the previus version of mpower but fails anyway, D:

latot:
mm, test will fail anyway, behavior of matpow in 0.7.6 its too bad, you can experiment it here:http://live.sympy.org/ practically you can't use a symbol as exponent in that version, so that test will fails.

cbm755:
Did you read my comment?  "packaging" doesn't seem to be a core module.

latot:
sorry i don't understood correctly that comment, changed to other func.

cbm755:
It means our software would have to depend on some other software that is not included in the default python.  Your new approach would depend on `python-setuptools`... But perhaps SymPy already does...

Take the MatPow change out: do it in a separate PR.  Simple small changes means easier reviewing, easier time with the tests, quicker merging...

latot:
okis, pckaging replaced with toher func of python, and back mpower

inst/@sym/mpower.m
```diff
@@ -129,7 +129,7 @@
 %! C = 10 + B + B^2;
 %! D = subs(C, n, 1);
 %! E = 10 + A + A^2;
-%! assert (isequal (D, E))
+%! assert (isequal (simplify(D), simplify(E)))
```
cbm755: no need to change this test

latot: its need, some versions of sympy give different results and fails, simplify fix it.

inst/findsymbols.m
```diff
@@ -62,7 +62,7 @@
 
   if isa(obj, 'sym')
     cmd = { 'x = _ins[0]'
-            'if sympy.__version__ == "0.7.5":'   % deprecate with Issue #164
+            'if Version(spver) <= Version("0.7.6"):'   % deprecate with Issue #164
```
cbm755: `< 0.7.6`?

inst/assumptions.m
```diff
@@ -134,7 +134,7 @@
     'adict_even_076 = {"real":True, "even":True, "commutative":True, "noninteger":False, "hermitian":True, "complex":True, "rational":True, "integer":True, "imaginary":False, "odd":False, "irrational":False}'
     'adict_integer = {"real":True, "commutative":True, "noninteger":False, "hermitian":True, "complex":True, "rational":True, "integer":True, "imaginary":False, "irrational":False}'
     'adict_rational = {"real":True, "commutative":True, "hermitian":True, "complex":True, "rational":True, "imaginary":False, "irrational":False}'
-    'if sympy.__version__.startswith("0.7.6"):'
+    'if Version(spver) >= Version("0.7.6"):'
```
cbm755: no I think this should be >= 0.7.6 and < 0.7.7-dev.

inst/findsymbols.m
latot: ups D:

cbm755:
This is great for the code!

I'm not sure how to figure out if the dependency on `setuptools` is a problem...  Not listed here:  https://github.com/sympy/sympy/wiki/Dependencies.  Any idea?

inst/assumptions.m
latot: fixed

latot:
sadly no idea D:

the new function maybe is included in python and not in sympy...

cbm755:
It is definitely not included in my Python on Fedora 23.  (of course its easy to install, but everything is easy to install on GNU/Linux).

Might be a problem on Windows.  And its already hard to tell people to instlal Python and SymPy there...

latot:
okis, so lets search other function

cbm755:
up to you...  other option is just leave my ugly kludges until we can drop 0.7.6.x.

latot:
i think you will need something like this in the future, so lets use an ugly but functional function.

cbm755:
I think I like 2a9deb52849b73f5430c55422fa58812a642ddd2 ("other ugly solution") as it avoids dependencies---I think.

Another possible treatment for dev would be something like (`0.7.7-dev` -> `0.7.6.999`)

I.e., `"n.m.p-dev"` maps to `(n, m, p-1, 999)`.  ugly, but safe enough for us?

latot:
its an option, at least best than https://github.com/cbm755/octsympy/commit/2a9deb52849b73f5430c55422fa58812a642ddd2

sympy uses i'm not wrong https://www.python.org/dev/peps/pep-0440/ to versions.

the reason of i avoid use tuples is because in every tuple function i try i normally found an error, for example in this is case: ```V("3.2.1")<=V("3.2.dev")``` this will get true D:
thats the reason finally of a lot of trys use pkg_resources, maybe if we start making or own code to PEP 0440 finally will be to complicated.. 

other things if your sympy gets 0.7.7-dev we will ahve conflicts with systems, in my case i get 0.7.7.dev...

latot:
well again two options:
first: https://github.com/latot/octsympy/commit/22e008972a77ed5ac90fe9ae68881c98709f9f72
second: https://github.com/latot/octsympy/commit/337d7db9a6aa85234d101e986c16eb6221bab68c

the second its a fragment of code of pkg_resources, only have the essential to work the version function.

cbm755:
I vote for 2a9deb52849b73f5430c55422fa58812a642ddd2.  It has `0.7.7 > 0.7.7.dev` as false (but should be true).  I don't think that is a problem.

Could write it as:
````
# Note: not pep0440
def Version(v):
    return LooseVersion(v.replace('.dev', ''))
````

Let's do that and merge this no more bikeshedding ;-)

cbm755:
I'd like a cleaner commit log.  Do you know how to rebase or shall I do it?

inst/@sym/mpower.m
cbm755: then do it in a new PR, its out of scope here

cbm755:
Or maybe this PR is useful history for someone later: could squash them on a new branch and do a new PR, then just close this if you want.

Or, anyway let me know and I'll do it.

cbm755:
Also, reminder to drop the `simplify` change for now...

latot:
okis, now are other pr, only for the record, this only will works while sympy don't use other word than ".dev", if this fails in the future i recomend merge this pr to finish this.

PR #323, opened 2016-01-20
move iscell check above error reporting

cbm755:
Also, inform the user that she might have been supposed to return
a tuple.

### DIFF
--- a/inst/python_cmd.m
+++ b/inst/python_cmd.m
@@ -141,15 +141,17 @@ function varargout = python_cmd(cmd, varargin)
 
   [A, db] = python_ipc_driver('run', cmd, varargin{:});
 
-  if (strcmp(A{1}, 'COMMAND_ERROR_PYTHON'))
-    error(A{2});
-  end
-
   if (~iscell(A))
+    disp('python_cmd: The return value was:');
     disp(A)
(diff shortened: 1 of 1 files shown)


cbm755:
@latot can you review this?

latot:
mm, i only can say 2 things, in my experience only are two instances when this happens, first, we forget tuple in return, (here good), the other, broken octsympy/octave, would be great can do something here.

cbm755:
ok, I added more to the error message.

latot:
i was thinking, if we enable the normal return of python?
```
  if (~iscell(A))
    A={A};
  end
```
the normal return don't is an error... so i think can help.

cbm755:
Good point, I filed another bug.

These blocks are still in the wrong order so I will merge this for now.

cbm755:
(tests didn't pass, will try again after merge)

latot:
mm, here always fails 3 tests, dsolve 2 tests and 1 test in orth.
using octave-4.0.0(no hg) and sympy git version

cbm755:
yeah dsolve is pretty $%^ked up in SymPy, haven't had to time/motivation to fix it.

But I should fix up that test so it doesn't fail on 0.7.7-dev.

As you've surely noticed, I mostly want the tests to pass on Travis CI...

Issue #314, opened 2016-01-17
FiniteSet support


cbm755:
When I first implemented setdiff and setxor, they use vectors to represent sets.

SymPy has fancier classes like `FiniteSet`.  And we want to work with `Intervals` as well.  Also `Union` of `Interval` and `FiniteSet`.

So there are three things we might support:

1. [ ] `union([sym(1) 2 3], [3 4])`.  This is for compatibility with Octave double behaviour.  I'm not 100% sure we want to support this.

2. [x]  `union(interval(sym(0),2), interval(sym(3),pi)]`

3. [x]   `union(<some FiniteSet>,  <some other FiniteSet>`.  We don't have `FiniteSet` yet, but one way to make one is `interval(2,2)`.

4. [x]  `A = interval(0,1), B = interval(2,2), C = interval(9,10)`.  Then `D = union(A, B)` should work.  Then `E = union(D, C)` should work too.

- - - - - -

I'm not sure we should support 1.  If we do support it, should the output be a Matrix or a FiniteSet.

2.,3.,4. are straighforward in SymPy.

cbm755:
TODO: 
- [ ] what should we call the function to create a `FiniteSet`?   `@sym/finiteset`?  or `@sym/set` (this might be confusing re: core `set` function.
- [x] how do we convert a vector to a `FiniteSet`?  A: use the previous function (?)
- [ ] how to we change a FiniteSet back into a vector?  We need to allow this because you cannot index `FiniteSet`...

cbm755:
The more I think about this, the more I think we should follow SymPy behaviour (and not Octave double behaviour).  Regarding: 1., maybe we should not return matrices.

Example: `union([sym(1) 2 3], [3 4])`
Here the user gives us a matrix, we either raise error, or convert the inputs to `FiniteSet`s.  And give her back a FiniteSet: `{1 2 3 4}`

PR #327, opened 2016-01-22
add finiteset

cbm755:
Its in global name space so we can create the empty set.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,7 @@ octsympy 2.3.0-dev
           intersect
           interval
           ismember
+          finiteset
           linspace
           logspace
           max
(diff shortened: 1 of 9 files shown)


cbm755:
@latot what do you think of this?

(I don't think Symbolic Math Toolbox has this functionality so we can call it whatever we want.)

latot:
m like this, i think call it finitset its fine, the only think, i would like have some write compatibility, for example, [3] or {3}.this pr https://github.com/cbm755/octsympy/pull/311 have some compatibility.
(ideally add arrays to this) https://github.com/cbm755/octsympy/issues/314

now as a user, if i don't know about sympy or python its difficult know about this, may be add an example in union/intersect where the solution its a finitset and say it.

cbm755:
Ok, good idea about docs, adding some, will do more in a bit.

I don't think `uniop_helper` is useful here: it does element-by-element operations on a matrix.  My feeling is `finiteset(matrix)` should create a set containing that matrix: I added an example to the docs to show this.

PR #328, opened 2016-01-22
Sets

cbm755:


cbm755:
Lots of questions now that we have sets!  Here is special treatment for `isempty`.  What about `numel`, and `size` and `length` and... and ...

cbm755:
#542 is a thread which proposes something like `@symset`, a subclass of `@sym` to deal with the specific features of Sets, Intervals, etc (such as those highlighted above).  This is somewhat comparable to how `@symfun` overrides the behaviour of `()`-indexing.

I'm uncertain whether this is a good idea or not.  `@symfun`s are already troublesome (but needed for SMT compat).

PR #333, opened 2016-01-23
enable normal return python

latot:
you know, lets try.

```
>> python_cmd("return 'hi'")
warning: called from
    python_cmd at line 170 column 5
ans = hi
```

:O num 333


cbm755:
Add a new test?
````
>> c = python_cmd('return 6')
c =  6
````
We'll need to update `help python_cmd` as well.

latot:
tests added!

PR #334, opened 2016-01-23
let syms handle errors

latot:
well the idea is stop this:
```python_cmd(blablabla, sym(x), sym(y)....)```

if can't be converted the sym function will handle the error, its posible continue adding supports normaly in python_copy_vars_to..

maybe a try catch there in sym, but with other error message

### DIFF
--- a/inst/private/python_copy_vars_to.m
+++ b/inst/private/python_copy_vars_to.m
@@ -91,8 +91,16 @@ function a = do_list(indent, in, varlist)
       c=c+1; a{c} = sprintf('%s%s.append(dict(zip(%s,%s)))', sp, in, inkeys, invalues);
 
     else
-      i, x
-      error('don''t know how to move that variable to python');
+
+      x = sym(x);
+      c=c+1; a{c} = [sp '# sym'];
+      % need to be careful here: pickle might have escape codes
+      % .append(pickle.loads("""%s"""))', x.pickle)
+      % The extra printf around the pickle helps if it still has
+      % escape codes (and seems harmless if it does not)
+      % Issue #107: x.pickle fails for matrices, use char() as workaround
+      c=c+1; a{c} = sprintf('%s%s.append(%s)', sp, in, sprintf(char(x)));
+
     end
   end
   c=c+1; a{c} = [sp '# end of a list'];

### CODE REVIEW

**Comment by cbm755** [comment]
Can you give an *explicit* example of what this allows?  I guess its something like 
````
python_cmd(blablabla, x, y....)
python_cmd(blablabla, sym(x), sym(y)....)
````
doing the same thing...

I don't think I like the IPC making new syms: it is supposed to be general, not sym specific.

**Comment by latot** [comment]
from setxor:
```  r = python_cmd (cmd, sym(a), sym(b));``` to   ```r = python_cmd (cmd, a, b);```

we get this on tests:
```
octave:3> test @sym/setxor
i =  2
x =

   1   2   4

***** test
 % one nonsym
 A = sym([1 2 3]);
 B = [1 2 4];
 C = setxor(A, B);
 D1 = sym([3 4]);
 D2 = sym([4 3]);
 assert (isequal (C, D1) || isequal (C, D2))
!!!!! test failed
don't know how to move that variable to python
```

sadly seems syms support more type of formats that python_copy_vars_to, like this example, and i think its necessary stop doing a sym() in all inputs in the functions, basically with that change it covers the difference between both and enable the rest of supports.

**Comment by cbm755** [comment]
right, but python_cmd should probably make a numpy (double) array from B in that case---although it doesn't support that yet.

I think this is a "no" for me.

**Comment by latot** [comment]
sorry i don't explain correctly my intentions, was remove sym() from functions, or in other way remove all sym in functions (python_cmd()), and pass all tests.

**Comment by latot** [comment]
other thing, actually sym tranform the octave to python, it feels can replace part of the python_copy_vars_to, or at least try join both systems.

**Comment by cbm755** [comment]
Anyway, the main problem with this approach is if I want:
`python_cmd(cmd, x, y, a)`
but `a` *should not* be converted to `sym`.  Currently, that is easy: `python_cmd(cmd, sym(x), sym(y), a)`.  How can I express this logic with your change?

**Comment by latot** [comment]
okis, now i'm understanding your concept, so how you think implement arrays in python_copy_vars_to.m? only the idea.

**Comment by cbm755** [comment]
> how you think implement arrays in python_copy_vars_to.m? only the idea.

Depends how efficient it should be...  for big matrices (>= 100x100) you'd probably want the raw bytes of the Octave representation.  IMHO, not worth trying given our current poor IPC.

Easier option would be something like doing it as a list of lists, one for each row, sort of like how     the `elseif (iscell(x))` block of `python_copy_vars_to.m` works...  See also the bit about struct to dict.

Please note: I don't want (yet) a dependency on numpy.  (`pytave` will almost certainly have one, that's the right time)

**Comment by latot** [comment]
Hi, well if i understand your idea, you need transform the octave arrays to python octave, sadly i don't know how works this part but i can do the function:
```
function r=m(a)
  t="";

  if !isscalar(a(1, :)(1,1)) || rows(a) > 1
    t=strcat("[", t);
    t=strcat(t, m(a(1, :)));

    for i = 2:rows(a)
      t=strcat(t, ", ");
      t=strcat(t, m(a(i, :)));
    end

    t=strcat(t, "]");

  else

    t=strcat("[", t);
    t=strcat(t, mat2str(a(1, 1)));

    for i = 2:columns(a)
      t=strcat(t, ", ");
      t=strcat(t, mat2str(a(1, i)));
    end
    
    t=strcat(t, "]");

  end

  r=t;

end
```
ex:
```
>> q
q =

   1   2   3
   2   3   4

>> m(q)
ans = [[1,2,3],[2,3,4]]
```

cbm755:
Why do you need this feature?  I don't have time to devote much time to such a thing without a clear use case.

Can you file another issue?  I'd like to close this one.

latot:
okidoki

PR #337, opened 2016-02-03
Little Fixes

latot:
simplify fix this difference in sympy 0.7.7.dev:
```
>> test @sym/mpower
***** test
 % matpow
 syms n
 A = sym([1 2; 3 4]);
 B = A^n;
 C = 10 + B + B^2;
 D = subs(C, n, 1);
 E = 10 + A + A^2;
 assert (isequal (D, E))
!!!!! test failed
assert (isequal (D, E)) failed
```


inst/@sym/mpower.m
```diff
@@ -72,10 +72,10 @@
   end
 
   cmd = { 'x, y = _ins'
-          'if x.is_Matrix and not y.is_Matrix:'
-          '    return sympy.MatPow(x, y).doit(),'
+          'if not(x.is_Matrix and not y.is_Matrix) or Version(spver) >= Version("0.7.7.dev"):'
+          '    return x**y'
           'else:'
-          '    return x**y,'
+          '    return sympy.MatPow(x, y).doit()'
```
cbm755: Ok, I think I prefer:
````
if x.is_Matrix and not y.is_Matrix and Version(spver) < Version("0.7.7"):
    return sympy.MatPow(x, y).doit(),'
````
(is that the same?)

latot:
okis, using new easy syntax

cbm755:
Thanks.

Just a note to self: that test really doesn't do what I wanted anymore (in SymPy 0.7.7).  It was supposed to create a MatPow but now SymPy 0.7.7 can actually do the symbolic power.  Perhaps it should be deleted.

PR #338, opened 2016-02-03
new line error handle

latot:
now works with all ipc systems.


latot:
maybe use ```sysmpref('ipc', 'linerror')``` instead

cbm755:
Sorry, but this scares me.

1.  Suppose there is an error related to the IPC.  Then we are using more IPC to get the error message...  seems fragile.  A possible workaround would be to do this when the IPC first starts and cache the value.

2.  I'm not sure the code is the same length every time.  Especially with the system (and similar) IPC, the python code to copy the variables over the python is auto-generated: it will be different numbers of lines each time.  So how can this possibly work?  (Maybe I'm wrong: I haven't actually tried it).

latot:
Hi, well i change the linerror to the same script in a persistent var, now ipc is free, this basically get the line error of the actual ipc and with it calcs the line error, actually i try it with all ipc available and seems works.
scares! D:

latot:
okis, now should works.

latot:
hi?

cbm755:
ok, I will try it

inst/python_cmd.m
```diff
@@ -138,11 +138,20 @@
...
-                      '%s'], errlineno, A{2});
-    error('OctSymPy:python_cmd:exception', errstr);
+
+    ex_popen2 = exist('popen2') > 1;
+    ipc_s = sympref('ipc'); 
+
+    if strcmp(ipc_s, 'popen2') || strcmp(ipc_s, 'sysoneline') || (strcmp(ipc_s, 'default') && ex_popen2)
+      errlineno = A{3} - line_error_python('get');
+    elseif strcmp(ipc_s, 'system') || strcmp(ipc_s, 'systmpfile') || (strcmp(ipc_s, 'default') && !ex_popen2)
+      errlineno = A{3} - line_error_python('get') - (nargin - 1)*2;
```
cbm755: this assumes every input takes exactly two lines?  I'm not sure that's true: maybe list or struct/dict breaks this?  Did you test with struct?

Even if true now, might not be true in future.

cbm755:
This is impressive, but I'm not convinced its correct.  Even if it is, it still scares me.

Perhaps the `python_ipc_*` files could report this information back (e.g., by counting newlines as the code is generated).  `[A, input_len, db] = python_ipc_driver('run', cmd, varargin{:});` or something like that.  Not sure.  But anything else is just too "spaghetti".

latot:
hi, checking the code seems do that will be very complicated, its posible merge this with some tests? i check the code and the system ipc use two blocks with the inputs while popen2 use one, so make sense, i want avoid complicate more this code.

cbm755:
Did you read my comment?

> I'm not sure that's true: maybe list or struct/dict breaks this? Did you test with struct?

I have now tried this, indeed it is not correct:
````
>> sympref ipc system
>> s.a = 1; s.b = 42; s.c = 'word'
s =

    a =  1
    b =  42
    c = word

>> python_cmd('raise NameError("hello")', s)
error: NameError: hello (line:11)
error: called from
    python_cmd at line 154 column 5
````
(note **line:11**)

**Comment by latot** [comment]
ups sorry i was confused, i'll try work on that part.

**Comment by latot** [comment]
okidoki, now works and with a lot of test to be sure.

**Comment by latot** [comment]
concept, first we remove this lines line_error_python, the function give all lines but without input, like constant, then, every type have two parts, first number of blocks by input line, and the number of lines to start and end the type, so resuming, ax-b

**Comment by cbm755** [comment]
I sent you a pull request which should demonstrate another possible way of doing this.  Its basically a two-line patch.

**Comment by latot** [comment]
yay, that better

**Comment by latot** [comment]
i remove the private script, reason, we can use it to handle automatic the rest of the lines, but if some system can't use it we will get a lot of errors, so handle it manually via LinesBeforeCmdBlock

if you think that don't will be the case i'll recover it

**Comment by cbm755** [comment]
>  so, actually are some system without support of popen2?

* matlab has no popen2.
* octave on windows has partially broken popen2, `system` is useful for debugging.

> i remove the private script

ok, I don't think we need it.  Sorry for the waste of time though!

File: inst/python_cmd.m [hunk 2]
```diff
@@ -138,11 +138,8 @@
   end
 
   if (strcmp(A{1}, 'COMMAND_ERROR_PYTHON'))
-    errlineno = A{3} - LinesBeforeCmdBlock - 1;
-    errstr = sprintf(['Python raised exception, maybe at line %d of code ' ...
-                      'block (assuming popen2)\n' ...
-                      '%s'], errlineno, A{2});
-    error('OctSymPy:python_cmd:exception', errstr);
+    errlineno = A{3} - db.prelines - LinesBeforeCmdBlock;
+    error(strcat(A{2}, ' (line: ', mat2str(errlineno), ')'));
```
cbm755: I liked that it specified it was a python error.  How about keep it like it was before, something like:
````
error: Python exception near line 1 of the code block\n
NameError: hello
````
(File a separate bug if you feel strongly about fitting it on one line).

File: inst/python_cmd.m [hunk 3]
```diff
@@ -116,7 +116,7 @@
 
   % Careful: fix this constant if you change the code below.
   % Test with "python_cmd('raise')" which should say "line 1".
-  LinesBeforeCmdBlock = 3;
+  LinesBeforeCmdBlock = 4;
```
cbm755: leave as `4` as that is more correct (the number of lines of boilerplate in `cmd`).

Then use `A{3} + 1 - ...` below?

inst/python_cmd.m
latot: mm, i prefer all in one line, like "all you need here".
and the other reason, maybe can complicate the tests...
i should test it with \n

cbm755: > mm, i prefer all in one line, like "all you need here".

Then file a new bug.  Let's keep this one simple.  One change per PR.

> and the other reason, maybe can complicate the tests...

so far, it has not.

> do you know how use patterns in %error <some pattern>?

Yes, somewhat.

latot: ha he hi ho hu
pattern, like ".. he .. ho .."

inst/python_cmd.m
latot: 4 its he number of line to get error line 1.

cbm755: I know.  But please read my suggestions:

> Then use `A{3} + 1 - ...` below

See the `+ 1`?

latot: mm, should be A{3} - 1

latot:
can you rebase this plis?
my rebase of git is bug.

cbm755:
Ok, I should be able to do that.

PR #347, opened 2016-02-12
Improve factor

cbm755:


inst/@sym/factor.m
```diff
@@ -17,30 +17,110 @@
...
+
+  if ((nargin > 1) || (~isempty (findsymbols (f))))
+    %% have symbols, do polynomial factorization
+
+    if (nargout > 1)
+      print_usage ();
+    end
+
+    cmd = { 'p = factor(*_ins)'
+            'return p,' };
```
latot: its posible only a return? like
```return factor(*_ins)```

File: inst/@sym/factor.m [hunk 2]
```diff
@@ -17,30 +17,110 @@
...
+    p = python_cmd (cmd, sym(f), varargin{:});
```
latot: f its already a sym, line 102

cbm755:
Fixed, thanks.  Will merge.

Issue #346, opened 2016-02-11
errors in variable input are not reported correctly


cbm755:
We have nice shiny new error handing code, but it may not properly handle the case where python exceptions are raised during variable import.

See #118.  

#344 will have to deal with this too: it probably related to `myerr`.  E.g., the new error handling code uses a different magic string to indicate and error as occurred.  This might actually be an easy fix---by hacking `myerr` command in `python_header.py`.

cbm755:
Fixed in #357.

PR #353, opened 2016-02-15
general error little rewrite

latot:
at least now we don't have a mess with an exception.

Fixes https://github.com/cbm755/octsympy/issues/346

### DIFF
--- a/inst/python_cmd.m
+++ b/inst/python_cmd.m
@@ -124,7 +124,16 @@
 function varargout = python_cmd(cmd, varargin)
 
   if (~iscell(cmd))
-    cmd = {cmd};
+    if (isempty(cmd))
+      cmd = {};
+    else
+      cmd = {cmd};
+    end
+  end
+
+  % empty command will cause empty try: except: block
+  if isempty(cmd)
+    cmd = {'pass'};
   end
 
   %% IPC interface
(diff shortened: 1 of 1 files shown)


latot:
mm, i think i'll do this pr after here, little but functional.

inst/private/python_copy_vars_from.m
```diff
@@ -16,13 +16,9 @@
...
 %% If not, see <http://www.gnu.org/licenses/>.
 
-function s = python_copy_vars_from(out, tryexcept)
+function s = python_copy_vars_from(out, tryexcept = true)
 %private function
 
-  if (nargin == 1)
-    tryexcept = true;
-  end
-
```
cbm755: this won't work in Matlab (default argument value is Octave specific syntax).

cbm755:
This looks like a good idea!  But I'm not sure its safe as is.  `myerr` can be called before `output_drv` is even defined...  Perhaps that's what I had the raw version of myerr?

latot:
sadly that version of myerr brokes octave D:
well i found when we have one of this types of errors the system is stalled, waiting some input of ```output_drv```, so now when we have some error it send an output and python_cmd do the rest, i move the vital functions to the top of python_header, to avoid call ```output_drv``` when don't is defined.

cbm755:
Just a tip: rebasing is generally not a good idea while work is being discussed.  Sometimes is appropriate right before merging.

latot:
okidoki

latot:
conflicts solved

cbm755:
Let's try to do one thing at a time: two output blocks are causing the problem. So maybe we can have a minimal patch that fixes that.  (We can think about reordering everything in python_header later.)

I'm experimenting a bit in #357.

latot:
mm, i choose reorder the header here because influences this issue

inst/private/python_header.py
```diff
@@ -11,16 +11,12 @@
...
-    print(str(e[1]).replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;"))
-    print("</f>\n</item>")
-    print("</output_block>\n")
-
+def myerr(error):
+    try:
+        octoutput_drv(("INTERNAL_PYTHON_ERROR", error))
+    except:
+        pass
+    raise
```
cbm755: Do we *need* to raise if the `octoutput_drv` works?  Or maybe its safer to do so?

(If I understand correctly, raise will just print the stacktrace to stderr)

latot: aa, this solve a little bug, basically when we have an error here before two things, first the system was stalled waiting some return, and second, the error message was passed as ans, myerr now handle the python error and say to python_cmd was an error there to avoid stall it, and then show the error with raise and stop the script, the try its only for when fails ```octoutput_drv``` or when not is defined following your advice.
wen i test the works fine, so i suppose its safe do this, at least now we don't "leave a mess"

PR #355, opened 2016-02-15
Allow `None` to be returned from Python

cbm755:
First two commits are the main change here.

Later commits are just to allow non-SymPy objects (like None) into some routines.  I don't know of any other objects that pass thru `_sympify()` but None does.

Anyway, I guess I have a slightly preference for `isinstance(foo, sp.MatrixBase)` over `foo.is_Matrix`---more pythonic as not does assume `foo` is SymPy.


latot:
i like this :D

cbm755:
Thanks, but unfortunately its not quite right yet.  I think if `q` is a `MatrixExpr`, `q.is_Matrix` is true but `isinstance(q, MatrixBase)` is false, so this change is not neutral...

cbm755:
redone, and rebased.

cbm755:
for the record, my original comment is misleading.  `is_Matrix:` includes both MatrixBase and MatrixExpr objects: some thought is generally required which to use...

PR #356, opened 2016-02-15
Just one try: except: block for cmd

### LINKED ISSUE #344: refactor error handling
Currently, this command `python_cmd('raise NameError("hello")')` generates the following Python code:
````
  [1,1] = try:
  [1,2] =     def _fcn(_ins):
  [1,3] =         _outs = []
  [1,4] =         try:
  [1,5] =             raise NameError("hello")
  [1,6] =             return _outs
  [1,7] =         except Exception as e:
  [1,8] =             exc_type, exc_obj, tb = sys.exc_info()
  [1,9] =             ers = type(e).__name__ + ": " + str(e) if str(e) else type(e).__name__
  [1,10] =             return ("COMMAND_ERROR_PYTHON", ers, sys.exc_info()[-1].tb_lineno)
  [1,11] =     _outs = _fcn(_ins)
  [1,12] = except:
  [1,13] =     octoutput_drv("PYTHON: Error in cmd")
  [1,14] =     myerr(sys.exc_info())
  [1,15] =     raise
````
Is the `octoutput_drv, myerr` bit just dead code now that we have the line 10?  Anyway, some refactoring is probably possible here.

Maybe we can drop `myerr` altogether?  But careful I think its used in the load variables code too...

### PR DESCRIPTION
**Pull request by cbm755**
Previously, we had a nested try: except: block.  Add some tests
for strange blank lines: all seems safe when used inside
a try:, even with popen2.

Fixes #344.

### CODE REVIEW

File: inst/private/python_ipc_popen2.m [hunk 1]
```diff
@@ -160,16 +160,14 @@
   % we send variables before, this is always zero.
   info.prelines = 0;
 
-  %% The actual command
-  % cmd will be a snippet of python code that does something
-  % with _ins and produce _outs.
-  s = python_format_cmd(cmd);
+  % code for output, or perhaps a thrown error
+  s2 = python_copy_vars_from('_outs')
```
latot: ups missed ";"
```
s2 = python_copy_vars_from('_outs');
```

inst/private/python_ipc_system.m
```diff
@@ -72,7 +72,7 @@
   %% The actual command
   % cmd will be a snippet of python code that does something
   % with _ins and produce _outs.
-  s2 = python_format_cmd(cmd);
+  s2 = cmd;
```
latot: its posible remove this and change line 81 to:
```
  s = strjoin([s1 cmd s3], newl);
```
??
or if you like s3 to s2??

cbm755:
Thanks for catching those.  For some reason I don't understand I had to change the `-1` fudge factor to `-2`...

cbm755:
I will merge if tests pass

cbm755:
oh I see why.  The error happens when we run the code, not when we create the function!

Compare:
````
python_cmd({'a=1' 'b=1' 'raise ValueError' 'c=1' 'd=1'});
````

To:
````
python_cmd({'a=1' 'b=1' 'raise ValueError'});
````

So`-2` is not the correct fix, but I may need to add a test like this.

cbm755:
So maybe this approach is not compatible with the error line number stuff.  If so, we'll need to go back to the blank line after function definition approach.

Issue #352, opened 2016-02-15
breaking octsympy/octave with double octoutput_drv


latot:
Hi again, well testing i think i found how we can broke this, i mean when we "leave a mess", basically its double return output with octoutput_drv like:

```
octoutput_drv("1")
octoutput_drv("2")
```

don't need to be consecutive.
a little question, how we can stop the execution of the script, whitout kill the pipe?

latot:
maybe with ```import pdb; pdb.set_trace()``` testing

latot:
the function myerr its breaking this too, i test this https://github.com/cbm755/octsympy/issues/118 without my err and works fine, what do you think remove  it? (we can just let python raise exceptions)

cbm755:
> how we can stop the execution of the script, whitout kill the pipe?

I'm not sure that's possible.  E.g., I think `sympref reset` just discards it and starts a new one: see top of `system_ipc_open2`.

cbm755:
I'm happy enough to remove `myerr` as long as everything still works.  We need to make sure the exceptions reliably make it back to Octave.

latot:
check https://github.com/cbm755/octsympy/pull/353 why i think we can't remove myerr

latot:
what exactly happends when you return two values?

cbm755:
> what exactly happends when you return two values?

`readblock` reads the first one.  The second one just sits there on the stdout.  Then the *next command* fails: see my test in 701e6bba2d302ea92c52573a9f5c762754381199.

cbm755:
fixed in #357

Issue #118, opened 2014-10-01
`sym(true) + 1` leaves a mess


cbm755:
````
AttributeError: 'BooleanTrue' object has no attribute 'as_coeff_Mul'
````
Need a `sympref reset` to recover.

This is basically issue #79.

**Comment by latot** [comment]
mm, this is weird check this, two differents outputs:
```
>> sym(true)+1
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 12, in octoutput_drv
  File "<stdin>", line 52, in octoutput
  File "<stdin>", line 33, in octoutput
  File "/usr/lib64/python2.7/site-packages/sympy/printing/pretty/pretty.py", line 2030, in pretty
    return pp.doprint(expr)
  File "/usr/lib64/python2.7/site-packages/sympy/printing/pretty/pretty.py", line 58, in doprint
    return self._print(expr).render(**self._settings)
  File "/usr/lib64/python2.7/site-packages/sympy/printing/printer.py", line 257, in _print
    return getattr(self, printmethod)(expr, *args, **kwargs)
  File "/usr/lib64/python2.7/site-packages/sympy/printing/pretty/pretty.py", line 1234, in _print_Add
    terms = self._as_ordered_terms(expr, order=order)
  File "/usr/lib64/python2.7/site-packages/sympy/printing/printer.py", line 271, in _as_ordered_terms
    return expr.as_ordered_terms(order=order)
  File "/usr/lib64/python2.7/site-packages/sympy/core/expr.py", line 889, in as_ordered_terms
    terms, gens = self.as_terms()
  File "/usr/lib64/python2.7/site-packages/sympy/core/expr.py", line 918, in as_terms
    coeff, _term = term.as_coeff_Mul()
AttributeError: 'BooleanTrue' object has no attribute 'as_coeff_Mul'
error: sq_string cannot be indexed with {
error: called from
    python_cmd at line 144 column 3
    axplusy at line 40 column 5
    plus at line 39 column 5
error: evaluating argument list element number 1
error: called from
    python_cmd at line 144 column 3
    axplusy at line 40 column 5
    plus at line 39 column 5
```
second:
```
>> python_cmd('return S.true + 1')
error: TypeError: unsupported operand type(s) for +: 'BooleanTrue' and 'int'
error: called from
    python_cmd at line 145 column 5
```
if its an upstream error i think the output should be equal..

cbm755:
ok, noted in upstream bug.

[...]

PR #357, opened 2016-02-17
Rebase353

cbm755:


inst/private/python_header.py
```diff
@@ -11,16 +11,12 @@
...
-    print(str(e[1]).replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;"))
-    print("</f>\n</item>")
-    print("</output_block>\n")
-
+def myerr(error, vital=0):
+    if vital:
+        raise
+    else:
+        octoutput_drv(("INTERNAL_PYTHON_ERROR", error))
+        raise
```
latot: update this plis, check new code in latests commits

inst/private/python_header.py
```diff
@@ -42,8 +38,7 @@ def myerr(e):
     import xml.etree.ElementTree as ET
     from distutils.version import LooseVersion
 except:
-    myerr(sys.exc_info())
-    raise
+    myerr("Python Header: Error doing imports (perhaps no SymPy?)", 1)
```
latot: remove ", 1"

inst/private/python_header.py
```diff
@@ -69,8 +64,7 @@ def Version(v):
         # short but not quite right: https://github.com/cbm755/octsympy/pull/320
         return LooseVersion(v.replace('.dev', ''))
 except:
-    myerr(sys.exc_info())
-    raise
+    myerr("Python Header: Error setting functions, block 1", 1)
```
latot: remove ", 1" plis

inst/private/python_header.py
```diff
@@ -141,9 +135,7 @@ def my_srepr(expr, **settings):
             """return expr in repr form w/ assumptions listed"""
             return _ReprPrinter_w_asm(settings).doprint(expr)
     except:
-        myerr(sys.exc_info())
-        raise
-
+        myerr("Python Header: Error setting functions, block 2", 1)
```
latot: same

inst/private/python_header.py
```diff
@@ -169,8 +161,7 @@ def octoutput_drv(x):
         else:
             print(DOM.toprettyxml(indent="", newl="\n", encoding="utf-8"))
 except:
-    myerr(sys.exc_info())
-    raise
+    myerr("Python Header: Error setting functions, block 3", 1)
```
latot: same

inst/private/python_header.py
```diff
@@ -264,8 +255,8 @@ def octoutput(x, et):
             dbout("type: " + str(type(x)))
             octoutput("python does not know how to export type " + str(type(x)), et)
 except:
-    myerr(sys.exc_info())
-    raise
+    myerr("Python Header: Error setting functions, block 4", 1)
```
latot: same

cbm755:
Thanks.  (I did try to `git cherry-pick` from your branch that commit but there were too many conflicts).

latot:
https://github.com/cbm755/octsympy/pull/358 to merge changes, don't need cherry-pick

cbm755:
how about this?  I haven't tried to make `python_ipc_system` use these ideas yet.

But popen2 should be much safer now.

I put the harded-coded xml function back: previously called `myerr`: I think this should work fine for strings.

latot:
mm, i like some parts and others not, why we can't just call raise from myerr (echo_exception_stdout)?? use it can help to do more easy use this code, and don't call raise everywhere, and don't call octoutput_drv in the myerr stall the system when some error happends D:
i need test some parts because the reason y remove the output of myerr was breaks octave/octsympy

inst/private/python_copy_vars_to.m
```diff
@@ -28,13 +28,13 @@
     %% put inside try-except
     L = do_list(4, in, varargin);
     L = { 'try:' ...
-      sprintf('    %s = []', in) ...
-      L{:} ...
-      '    octoutput_drv("PYTHON: successful variable import")' ...
-      'except:' ...
-      '    octoutput_drv("PYTHON: Error in variable import")' ...
-      '    myerr(sys.exc_info())' ...
```
cbm755: The double output here is the bit that was causing #352 (and #118).  Basically I just remove one of them...

cbm755:
> why we can't just call raise from myerr (echo_exception_stdout)?? 

Could, I just think the `raise` looks better near the `except:`...  don't feel that strongly...

latot:
okisdokis

cbm755:
Shall I merge this?  Then we can decide if

a)  use your try: `octoutput_drv except: echo_exception_stderr`

b) re-arrange `python_header.py` as you do other places

c) drop the hardcoded xml (you wanted I think).

d) put all of `python_header.py` inside one big `try: except` (instead of blocks).  This is because with popen2, the other blocks get pasted into the stdin even if first one raises an exception...  Maybe I should file another issue about this...

latot:
About d) i'm agree, i think we can order to, vital functions, imports, and others definitions (in that order).

to do that ideas i think should do b and then a
i think can't remove all hardcoded, because a, if fails octoutput_drv we need some way to call.

and the main issue, merge this :D

cbm755:
Ok, fixed the `system` ipc.  Will merge when tests pass.

re: (d)  if just one big try-except block then no reason to re-order: either it all works or it all fails...

PR #364, opened 2016-03-03
working on 2d expansions

trueutkarsh:
added new file mtaylor.m which has code for 2d expansions .
Issue with python code in cmd block.


cbm755:
1. First of all, I think we should try to extend `taylor` rather than a new function.
2. You've checked in a octave-core file.
3. why diff etc?   does multiple calls to `.series` not work?   (see the stackoverflow page I linked to earlier).

inst/@sym/mtaylor.m
```diff
@@ -0,0 +1,200 @@
...
+    n = int32(n);
+  end
+
+  if (numel(x) == 1)
+    warning('FIXME: Issue #31 multivar Taylor expansions not implemented')
+  end
+
+  cmd = { '(f, x, a, n) = _ins'
+          '#s = f.series(x, a, n).removeO()'
+          'import math as ms'
```
cbm755: delete this line

inst/@sym/mtaylor.m
```diff
@@ -0,0 +1,200 @@
...
+  end
+
+  if (numel(x) == 1)
+    warning('FIXME: Issue #31 multivar Taylor expansions not implemented')
+  end
+
+  cmd = { '(f, x, a, n) = _ins'
+          '#s = f.series(x, a, n).removeO()'
+          'import math as ms' 
+          'from sympy import *'
```
cbm755: delete this line

inst/@sym/mtaylor.m
```diff
@@ -0,0 +1,200 @@
...
+          'expr=f'
+          'var=x'
+          'i=a'
+          'expn=expr.subs([(var[x],i[x]) for x in range(len(var))]) # first constant term'
+          'for x in range(1,n,1) :'
+          '  bin=[binomial(x,m) for m in range(x+1)] #binomila constands'
+          '  x1=x'
+          '  x2=0'  
+          '  fact=ms.factorial(x)'
```
cbm755: `fact=sp.factorial(x)`

cbm755:
I tried your code, after I fixed the above, it seems to sort of work:
````
>> syms x y
>> f = exp(x) + exp(y)
f = (sym)

   x    y
  ℯ  + ℯ 

>> mtaylor(f, [x y], [0 0])
ans = (sym)

   3    2        3    2        
  x    x        y    y         
  ── + ── + x + ── + ── + y + 2
  6    2        6    2         
````

But it didn't work for `exp(x*y)`.

I suggest looking at using `.series` instead of trying to do it yourself.

**Comment by trueutkarsh** [comment]
@cbm755 please check the latest commit ,I've resolved previous issues.
The reason I did'nt use .series earlier was that the output is simplified too much that its hard to check manually. You'll see in this commit ,if you change point of evaluation something other than [0,0..0]
it becomes a little mess.

**Comment by cbm755** [comment]
Might be good to start thinking about and implementing some test cases.
1.  use `isAlways`, then your expected output does not need to be in the exact same form as the taylor series.
2.  does it create the correct number of terms?

**Comment by trueutkarsh** [comment]
@cbm755  order is creating problem . Using .series we're gonna have some terms of order greater than n-1 (upto 2n-1) because our operation mulitplies the term from both individual series. 
Hence the terms are not gonna remain same for a given order.
I have'nt been able to find a direct function to remove terms of order greater than something in sympy or octave ?

**Comment by cbm755** [comment]
I see.

I wonder about trying `series(x, n=None)`.  Then using an *appropriate* value for `n` in the `y` series---i.e., decreasing `n` as we iterate through the terms of the `x` series.  Note I haven't tried this.

**Comment by trueutkarsh** [comment]
@cbm755 I think it will boil down to using diff() since now we need to work on each term individually . 
I guess the first soln where each term was computed separately  would'nt have this problem.

**Comment by cbm755** [comment]
Perhaps.  I'd still like to try an approach based on series as I outline above.  But you can do with `diff` for now if you want (maybe your original implementation?).

Most important is to build a nice test set---easy to try other implementations after that.

**Comment by trueutkarsh** [comment]
@cbm755  in console all the checks have passed but here it is'nt the case.
Currently all 1d and 2d cases work (is uses diff). For series with more than three variables the series can be evaluated but without correct order(it is .series) . I am working on it.

**Comment by cbm755** [comment]
FYI: the doctests are separate: you can run them by first doing `pkg install -forge doctest` then `pkg load doctest` then `doctest @sym/taylor`.

**Comment by trueutkarsh** [comment]
@cbm755  this function is capable of doing single variable and 2d expansions .
for (n>2) order is still an issue .But for rest its pretty good .

**Comment by cbm755** [comment]
OK I will review.... Did you drop mtaylor yet?  It should just extend taylor.m

**Comment by trueutkarsh** [comment]
@cbm755  extended taylor,removed mtaylor.

File: inst/@sym/taylor.m [hunk 4]
```diff
@@ -30,14 +30,14 @@
...
-%%          ─── + ── + ── + ── + x + 1
-%%          120   24   6    2
+%% >> syms x y 
+%% >> f = exp(x*y);
+%% >> taylor(f, [x,y] , [0,0], 'order', 6)
+%%    @result{}  (sym)
+%%                   2  2          
+%%                  x ⋅y           
+%%                  ───── + x⋅y + 1
+%%                    2
```
cbm755: Leave the existing examples please.  You can (and should) add new ones

inst/@sym/taylor.m
```diff
@@ -56,37 +58,50 @@
...
-    x = symvar(f,1);
-    a = sym(0);
-    i = 1;
-  elseif (nargin == 2)  % taylor(f,x)
-    x = varargin{1};
-    a = sym(0);
-    i = 2;
-  elseif (~ischar(varargin{1}) && ~ischar(varargin{2}))
-    % taylor(f,x,a,...)
+    x = symvar(f);
```
cbm755: This should stay as `symvar(x,1)`: by default it does single variable.

inst/@sym/taylor.m
```diff
@@ -120,9 +167,10 @@
 %! syms x
 %! f = exp(x);
 %! expected = 1 + x + x^2/2 + x^3/6 + x^4/24 + x^5/120;
-%! assert (isequal (taylor(f), expected))
-%! assert (isequal (taylor(f,x), expected))
-%! assert (isequal (taylor(f,x,0), expected))
+%! assert (isAlways(taylor(f)== expected))
+%! assert (isAlways(taylor(f,x)== expected))
+%! assert (isAlways(taylor(f,x,0)== expected))
+
```
cbm755: If tests are working, don't change them.  Nothing 1D should change here....

cbm755:
See comments above.  Also: please remove junk workspace files etc.  And delete mtaylor.m?

trueutkarsh:
@cbm755  please review.

cbm755:
You are asking for more review but you have not yet addressed all my comments from my previous reviews:
1. There are still junk files (e.g., octave-workspace).
2. There is still a copy of mtaylor.m.
3. Many of the existing 1D tests have been changed for no obvious reason.
I think I mentioned all of these at least once above!

As to the actual code:
* Do not `import math`
   * use a symbolic factorial from sympy (`sp.factorial?`)
   * There is a `1.0` in the code: probably incorrect.  Either `1/...` or `S.One/...` I guess?
   * Don't use `nsimplify`.  Basically, this algorithms seems to do stuff in floating point and should not.
* Please spell check: `binomila constands`, `upto` is not a word, etc.

inst/@sym/taylor.m
```diff
@@ -22,22 +22,30 @@
 %% @deftypefnx {Function File} {@var{g} =} taylor (@var{f}, @var{x})
 %% @deftypefnx {Function File} {@var{g} =} taylor (@var{f}, @var{x}, @var{a})
 %% @deftypefnx {Function File} {@var{g} =} taylor (@dots{}, @var{key}, @var{value})
-%% Symbolic Taylor series.
+%% Symbolic Two dimesional  Taylor series.
```
cbm755: I prefer previous version

inst/@sym/taylor.m
```diff
@@ -1,4 +1,4 @@
-%% Copyright (C) 2014, 2015 Colin B. Macdonald
+%% Copyright (C) 2014, 2015 Utkarsh Gautam
```
cbm755:
````
%% Copyright (C) 2014, 2015 Colin B. Macdonald
%% Copyright (C) 2016 Utkarsh Gautam
````
And then anyone else who modifies it in 2016 adds her name next to yours.

trueutkarsh:
@cbm755 sorry about last few commits , I've encorporated everything you said in the latest commit .

cbm755:
Well, not quite everything I said is incorporated.  But I can patch a bit before I merge...

cbm755:
Please see e0571afc9126ca7230f1bd24e463c96e6f460757 for future reference.  Generally, one should have minimal patches which make only changes related to the issue at hand.

Thanks for this, I'll merge #390 in a bit.  It has all your work here squashed (to avoid the binary files appearing in the history, etc).

cbm755:
Ok this is in!  thanks again.

PR #365, opened 2016-03-04
Use carat notation for xor on newer sympy

### LINKED ISSUE #340: setxor could prepare for the future
````
  %%FIXME: In future version of SymPy, replace (A - B) | (B - A) with A ^ B
  % Version(spver) >= Version("0.7.7.dev")
````
With @latot 's nice `Version()` work, this doesn't have to be a comment, the code could actually do this.

### PR DESCRIPTION
**Pull request by cbm755**
Fixes #340.

### DIFF
--- a/inst/@sym/setxor.m
+++ b/inst/@sym/setxor.m
@@ -60,11 +60,16 @@ function r = setxor(a, b)
   cmd = {
          'a, b = _ins'
          'if isinstance(a, sp.Set) or isinstance(b, sp.Set):'
-         '    return (a - b) | (b - a),'
+         '    if Version(spver) < Version("0.7.7"):'
+         '        return (a - b) | (b - a)'
+         '    return a ^ b'
          ''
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cbm755** [comment]
@latot does this look ok?

**Comment by latot** [comment]
mmmmmm, i don't know, actually sympy call exactly the same code overloading the operator, so maybe use the the if only will affect the performance (lower), i think the comment FIXME and add this to https://github.com/cbm755/octsympy/issues/315 should be enough.

File: inst/@sym/setxor.m [hunk 1]
```diff
@@ -60,11 +60,16 @@
   cmd = {
          'a, b = _ins'
          'if isinstance(a, sp.Set) or isinstance(b, sp.Set):'
-         '    return (a - b) | (b - a),'
+         '    if Version(spver) < Version("0.7.7"):'
```
latot: if you wanna merge this its posible use 0.7.7.dev instead?

cbm755:
ok

PR #366, opened 2016-03-05
For large integers, enable the "rat()" heuristics warning

### LINKED ISSUE #361: big numbers
Hi, well othe problem in octave affecting here, ex:

```
sym(100000000000000000000000)
ans = (sym) 99999999999999991611392
```

two options of temporary solve:

1)use exp
```
sym(10)^23
```

2) use ""
```
sym("100000000000000000000000")
```

cbm755:
This is because large integers represented in double-prec, cannot
be reliably identified as integer, nor is their value precise
enough (for example 1e16 + 1 == 1e16).  We give the same warning
that `sym(10.3)` gives.  The best way to enter large integer values
is `sym(10)^16`.

Fixes #361

### DIFF
--- a/inst/@sym/private/magic_double_str.m
+++ b/inst/@sym/private/magic_double_str.m
@@ -1,4 +1,4 @@
-%% Copyright (C) 2015 Colin B. Macdonald
+%% Copyright (C) 2015, 2016 Colin B. Macdonald
 %%
 %% This file is part of OctSymPy.
 %%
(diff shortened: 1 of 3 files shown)


cbm755:
CC @latot

inst/@sym/private/magic_double_str.m
```diff
@@ -48,7 +48,8 @@
     s = '-inf';
   elseif (isnan(x))
     s = 'nan';
-  elseif (isreal(x) && (mod(x,1) == 0))   % is integer
+  elseif (isreal(x) && (abs(x) < 1e15) && (mod(x,1) == 0))
```
latot: should be ```abs(x) <= 1e15```???

cbm755:
True.  But I was also worried about values where `mod(x,1)==0` was reliable...

We could also just use something arbitrary like 10000 or 1000000.  I'm pretty sure if anyone is using a really big number without quotes than they don't really know what they are doing and the warning is probably appropriate....

cbm755:
I'll merge for now, but of course we can change change this value later if desired.

Issue #254, opened 2015-07-03
adopt the maintainers makefile


cbm755:
See discussion in #238.

cbm755:
We basically have, although probably room for improvement.

Issue #82, opened 2014-08-20
popen2 on windows is sensitive to stderr output (e.g., by python)


cbm755:
Upstream Octave:

https://savannah.gnu.org/bugs/?43036

FIXME: workaround this.

cbm755:
We have a workaround since long time ago.  Only thing needed is fixing this upstream.

okejokej:
Hi! Just wanted to note that it'd be helpful if it'd be more explicitly(that is, somewhere outside of the file itself) stated that winwrapy.bat might require editing before the package actually begins functioning.
It seems to apply to more than just old systems as both my XP and Win 7 machines just flip the table if there's an attempt of using symbolic package without editing the bat file(to leave only the necessary minimum).

Cheers!

cbm755:
thanks @okejokej 

Perhaps you could [edit part of the wiki](https://github.com/cbm755/octsympy/wiki/Notes-on-Windows-installation) with your experiences on XP?  (I don't really know what edits to the wrapper that you're referring to)

okejokej:
Alright, added a note about it at the bottom - I completely missed that Wiki ^^

cbm755:
thanks!  And I know I know what changes you're talking about too :-)

Where do you suggest documenting this where you would have found it?

okejokej:
I think it would be good to put a link to that wiki in package's description on Octave-Forge, that is if it allows links, of course :P If not, some mention of Wiki's existence, perhaps.

cbm755:
Added to the README for now, thanks for the suggestions.

cbm755:
@mtmiller has proposed a patch at https://savannah.gnu.org/bugs/?43036 and needs testing.

cbm755:
Many thanks to @mtmiller and Tatsuro MATSUOKA, this is fixed upstream!

PR #370, opened 2016-03-06
Forced use of bundled python on windows.

### LINKED ISSUE #368: windows "bundle" pkg uses system python (instead of bundled py.exe)
I tried the following to install symbolic on octave 4.0
```
>>pkg install symbolic-win-py-bundle-2.2.4.zip
For information about changes from previous versions of the symbolic package, run 'news symbolic'.
>> pkg load symbolic
>> syms x
```
And i got the following error message after the last command
```
OctSymPy v2.2.4: this is free software without warranty, see source.
Detected Windows: using "winwrapy.bat" to workaround Octave bug #43036
Some output from the Python subprocess (pid 6304) might appear next.
extractblock: read an error back from python
str1 =
<class 'ImportError'>

str2 =
No module named 'sympy.logic.boolalg'

error: error reading back from python
error: called from
    extractblock>process_item at line 166 column 7
    extractblock>helper at line 51 column 14
    extractblock at line 13 column 8
    python_ipc_popen2 at line 82 column 7
    python_ipc_driver at line 30 column 13
    python_cmd at line 138 column 9
    sym at line 361 column 5
    syms at line 158 column 9
error: called from
    syms at line 158 column 9
```
Any workaround? Am I missing something?

genuinelucifer:
Fixes #368

### DIFF
--- a/bin/winwrapy.bat
+++ b/bin/winwrapy.bat
@@ -4,7 +4,7 @@ REM This bat file discards stderr from python.exe.
 REM This is needed as Octave for Windows cannot tolerate stderr from a popen2
 REM subprocess: see [this bug report](https://savannah.gnu.org/bugs/?43036).
 
-REM We look for python.exe and use that if found.  Otherwise we try py.exe.
+REM We look for python.exe and drop the stderr output in that if found.  Otherwise quit.
 REM You'll need to edit this script if your python is called something else.
 
 REM This batch file uses the "where" command, which is present only for
@@ -13,14 +13,11 @@ REM as Windows XP) only lines 17 or 20 should be kept in the next code block.
 
 REM returns 0 if python.exe is found
 where /Q python.exe
-IF ERRORLEVEL 1 GOTO else
+IF ERRORLEVEL 1 GOTO endif
 python.exe -i 2> NUL
-goto endif
-:else
-py.exe -i 2> NUL
 :endif
 
-REM If python is not installed in your system, "py.exe" will be used.
+REM If you are using the windows bundle for octsympy then "octpy.exe" will be used.
 REM It is [available online](http://www.orbitals.com/programs/pyexe.html)
 REM You may need to include its folder (e.g.,
 REM    C:\Octave\Octave-3.8.2\share\octave\package\octsympy-2.1.1\bin)
@@ -29,13 +26,13 @@ REM in the windows environment variable %PATH
 REM Other options for the stderr:
 
 REM redir stderr to stdout
-REM py -i 2>&1
+REM octpy -i 2>&1
 
 REM drop stderr
-REM py -i 2> NUL
+REM octpy -i 2> NUL
 
 REM log stderr
-REM py -i 2> stderrlog.txt
+REM octpy -i 2> stderrlog.txt
 
 REM Fancier logging with busybox/tee
 REM del /Q inlog.txt
--- a/make_windows_package.sh
+++ b/make_windows_package.sh
@@ -47,6 +47,7 @@ cp -R octsympy ${WINDIRTMP}
   ./bootstrap
   ./configure
   make )
+cp bin/winwrapy.bat ${WINDIRTMP}/bin/
 
 # copy things to the package
 mkdir ${WINDIR}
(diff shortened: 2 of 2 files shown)


cbm755:
Looking good.

But we need winwrap.py on standard installs too (its the workaround for stderr bug #82).  So I don't think we can always use octpy.exe...  Maybe should ship two different winwrap.py, one for standard installs and one for the bundle?  Or we can `sed` to fix it in the make bundle script...

cbm755:
#82 is annoying: at the time, I could find no other way to discard the stderr except using this bat file...

Maybe we should always assume `python.exe` for the standard package and `sed` it to `octpy.exe` for the bundle...

genuinelucifer:
By standard install, you mean installs on windows only, right?
If standard installs work too, then yes ```sed``` would probably be the best way.
Should i do it?

cbm755:
"standard install" means `pkg install -forge symbolic` (i.e., the user installs her own python and sympy).

Maybe we do not need to rename `py.exe`: maybe we just need to drop the `WHERE` part `winwrapy.bat`?

> Should i do it?

Yes please.  I cannot merge this PR as it is right now.

genuinelucifer:
I couldn't think how to solve this after removing the ```where``` part. So, I made it to look for the bundled python first. I didn't actually release and test it though. But, I guess it should work.

genuinelucifer:
Probably checking in the current directory would be better. If it's not desired, please tell me and i'll revert it.

bin/winwrapy.bat
```diff
@@ -12,12 +12,13 @@ REM Windows Server 2003 and later versions. So, for older windows systems (such
 REM as Windows XP) only lines 17 or 20 should be kept in the next code block.
 
 REM returns 0 if python.exe is found
-where /Q python.exe
-IF ERRORLEVEL 1 GOTO else
+REM first try bundled python executable, if it is
+REM not found then use system python
+IF exist "./octpy.exe" GOTO bundledPython
```
cbm755: I'm not sure about `./octpy.exe`: maybe the cwd is somewhere else and this dir is merely in our path?  Have you tested this with `python.exe` installed?

cbm755:
BTW, please don't rebase during review.

bin/winwrapy.bat
cbm755: Maybe when a `.bat` file runs, the cwd refers to the directory where the `.bat` lives rather than where it was run from?  citation needed.

cbm755:
Given the bit about `where` failing on Win XP, maybe just putting `python.exe` and then doing `octpy.exe` with `sed` in the bundle script is simpler...

bin/winwrapy.bat
genuinelucifer: I tried it and it doesn't work. Probably have to change directory first!

genuinelucifer:
Sorry for the rebase. I ran one of the scripts and it cloned one more octsympy and my last commit added so many extra files into the branch. Instead of revert, I just removed everything and made a fresh branch and force push. (bad habit) Will remember in future.

bin/winwrapy.bat
genuinelucifer: Even changing directory didn't work! I used 
```cd /D %0\..```
didn't work...

**Comment by genuinelucifer** [comment]
@cbm755 I tried doing many other things to look for the octpy.exe and then going for system python but sadly nothing seems to work. And I could not find official documentation of how to run the bat file in the folder it is. Whatever I found via google didn't work.
.
The only solution (which you suggested) that seems will work is having 2 different bat scripts for the different bundles. Should I do that (modify scripts to copy different bat files for different packaging)?

**Comment by cbm755** [comment]
> The only solution (which you suggested) that seems will work is having 2
> different bat scripts for the different bundles. Should I do that
> (modify scripts to copy different bat files for different packaging)?

Ok.  Let's just do one .bat file with python.exe in it and a `sed`
command (in the `build_windows_bundle.sh`) to change `python.exe` to
`octpy.exe` when building the bundle.

**Comment by cbm755** [comment]
what is the status of this?

**Comment by genuinelucifer** [comment]
Oops. Had forgot about this. I've been using octsympy on linux.
I guess we should have 2 different bat files. If that's okay, I'll push
some changes tomorrow.

**Comment by cbm755** [comment]
See my previous comment for what I think should be done

**Comment by genuinelucifer** [comment]
@cbm755 I have verified it on my windows. When I have python 3.4 in my ```PATH``` then too the ```octpy.exe``` is found and used.
.
Please review.

File: make_windows_package.sh [hunk 2]
```diff
@@ -47,6 +47,7 @@ cp -R octsympy ${WINDIRTMP}
   ./bootstrap
   ./configure
   make )
+cp bin/winwrapy.bat ${WINDIRTMP}/bin/
```
cbm755: what is this line for?

cbm755:
Merged except for the line highlighted above.

make_windows_package.sh
genuinelucifer: > > +cp bin/winwrapy.bat ${WINDIRTMP}/bin/
>
> what is this line for?
>
> —

Well, whenever i ran 'make_windows_package.sh' it always downloaded a fresh
copy of octsympy from git and packaged it. It did not include my
winwrapy.bat but the one from github. So, in order to include my edited
winwrapy.bat, i used this.

cbm755: Ok, I see.

I'm still dropping that line.  These scripts probably need some work: they are intended to run from *another directory*, checkout a clean copy and package it.  I agree that's not ideal for testing.  Guess they could be improved as part of #254.  Probably incorporated into the makefile itself.

Although, if memory serves the @oheim's Maintainer's Makefile still needs changed committed (although not pushed to remote repo).

Thanks again for this.

PR #372, opened 2016-03-08
basic sorting function

trueutkarsh:
Basic sorting function for symbols.


inst/@sym/sort.m
```diff
@@ -0,0 +1,72 @@
+%% Copyright (C) 2014, 2015 Colin B. Macdonald
```
cbm755: your name here

inst/@sym/sort.m
```diff
@@ -0,0 +1,72 @@
...
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @deftypefn  {Function File} {@var{g} =} sort (@var{f})
+%% Symbolic Two Variable Taylor series.
```
cbm755: taylor?

inst/@sym/sort.m
```diff
@@ -0,0 +1,72 @@
...
+
+%!test 
+%! f = [sym(1),sym(0.5)]
+%! sort(f)	
+%! expected = [1/2 ,1]
+%! assert (isequal (sort(f), expected))
+
+%! sym x
+%! f=[sym(1) x]
+%!	error <Invalid Input> sort (f)  # test that throws specific error
```
cbm755:
````
%!error <foo>
%! syms x
%! ...
%! sort (f)
````
You can run these tests locally with `test @sym/sort`

**Comment by cbm755** [comment]
Looks like a good start.

You'll need to clean up the comment header.

The tests are too verbose, need some semicolons.  See above for how to write the last test.

I think the `expected` stuff should be `sym([...])` although they might pass without.

**Comment by cbm755** [comment]
Since you're just calling Octave's core `sort`, it should be pretty easy to make `[S, I] = sort(foo)` work properly, then add some tests and document it.

**Comment by cbm755** [comment]
you might also test that `sort(sym([]))` and `sort(sym(2))` work correctly.

File: inst/@sym/sort.m [hunk 4]
```diff
@@ -0,0 +1,82 @@
...
+%!test 
+%! f = [sym(1),sym(0.5)] ;
+%! sort(f)	;
+%! expected = sym([sym(1/2) ,sym(1)]) ;
+%! assert (isequal (sort(f), expected))
+
+%! test 
+%! sym x ;
+%! f=[ sym(1) x];
+%! error <Invalid Input> sort (f)  # test that throws specific error
```
cbm755: I think I told you exactly what to type for this test: this isn't it.

cbm755:
Those binary and other junk files have got to go.  I suggest looking at `git add -p -i` and never using `git commit -a`.

inst/@sym/sort.m
```diff
@@ -0,0 +1,82 @@
...
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @deftypefn  {Function File} {@var{g} =} sort (@var{f})
+%% Sorting symbols .
+%% Author: Utkarsh Gautam
```
cbm755: not a valid deftype block:   try `help @sym/sort` and look at other files.

inst/@sym/sort.m
```diff
@@ -0,0 +1,82 @@
...
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @deftypefn  {Function File} {@var{g} =} sort (@var{f})
+%% Sorting symbols .
```
cbm755: extra space before full stop and maybe not the best one-liner: it actually can only sort Numbers.

Maybe (read this)[https://www.gnu.org/software/octave/doc/interpreter/Documentation-Tips.html]

inst/@sym/sort.m
```diff
@@ -0,0 +1,82 @@
...
+
+%!test 
+%! f = [sym(1),sym(0)] ;
+%! sort(f)	;
+%! expected = sym([0 , 1]) ;
+%! assert (isequal(sort(f), expected));
+
+%!test 
+%! f = [sym(1)] ;
+%! sort(f)	;
```
cbm755: no spaces before semicolon.  Its customary to copy the style of other functions when contributing to a Free Software project.

inst/@sym/sort.m
```diff
@@ -0,0 +1,82 @@
...
+%! expected = sym(f) ;
+%! assert (isequal(sort(f), expected));
+
+%!test
+%! f = [sym(3),sym(2),sym(6)] ;
+%! [s,i]=sort(f)	;
```
cbm755: `[s, i] = sort(f);`
(note spaces)

inst/@sym/sort.m
```diff
@@ -0,0 +1,82 @@
...
+%! expected_s = ([sym(1) ,sym(1);sym(2) ,sym(2);sym(3) ,sym(3)]) ;
+%! expected_i = ([1,3;2,1;3,2]) ;
+%! assert (isequal(s,expected_s) && isequal(i,expected_i));
+
+%!test 
+%! f = [] ;
+%! sort(f)	;
+%! expected = sym([]) ;
+%! assert (isequal(sort(f), expected));
```
cbm755: you want `sort(sym([])`, this can probably just be one line
````
%!assert (isequal (sort(sym([])), sym([])))
````

File: inst/@sym/sort.m [hunk 10]
```diff
@@ -0,0 +1,82 @@
...
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @deftypefn  {Function File} {@var{g} =} sort (@var{f})
+%% Sorting symbols .
+%% Author: Utkarsh Gautam
+%% Keywords: symbolic, sort
+
+function [s,i] = sort(f)
+	if isempty(f)
+		[s,i]=sort(f)
```
cbm755: Either this is recursive or this branch is never taken, or perhaps both ;-)

`sort([])` will not call `@sym/sort` so you don't need to worry about that: one of the inputs must be `sym` to call `@sym/<method>`

cbm755:
See above for a partial review.   Also, still a bunch of files committed here that should not be.

Getting there I think :)

cbm755:
Weee passing ;-)

Something to think about: does it work if `f` contains a noninteger such as `sqrt(sym(2))`?

trueutkarsh:
Since we are converting every element  to double it works .But it converts all of them to rationals.
Can we do something about it ? or do we need to ?

cbm755:
Yes I think we have to worry about that. I think you'll have to do the sorting in Python.

trueutkarsh:
@cbm755  I had the idea of first computing 
`[l,i]=sort(double(f))`
then using this 'i' matrix we could form new matrix 's' .


```
function [s,i] = trialsort(f)
	if isempty(f)
		s=f;
		i=f;
	else 
	  if isallconstant(f) 
	    [k,i]=sort(double(f));
	    s=sym(zeros( rows(f),columns(f) ) );
	    i
	    if rows(f)==1 
		    for l=1:columns(f)
		    	l
		    	%sym(f(1,int16(i(1,l))))
		    	s(1,l)=f(1,i(1,l))
		    end
		  else
		  	for l=1:columns(f)
		  		for m=1:rows(f)
		  			s(m,l)=sym(f(int16(i(m,l)),l));
		  		end
		  	end	
		  end  	  	
	  else
	    error('Invalid Input')
	  end
	end    
end
```
Logically this should work ,but I always get indexing out of range error (which i dont think should arise.)

trueutkarsh:
@cbm755 now it can sort  sqrt(sym(2)), pi and stuff like that as well .  
But since sym(sqrt(2)) is  automatically converted to a rational( rat() heuristics),I cannot to anything about it .

cbm755:
* as noted above, you still have other junk files checked in: can you remove these?

* you'll need to do the sorting in python (inside `cmd  = ` block).

trueutkarsh:
@cbm755 there isn't a way to compare symbols in sympy so I dont think sympy would be of help there.
By the way I guess current function does sort symbols integers or non integers in octave  . The rationalizing issue can only be solved if sym(1/3) are represented as 1/3 not 0.33 by sym function (line 214 sym.m) .
This is an issue of sym(sqrt(2)) being not equal to sqrt(sym(2)) symbolically in octave.

cbm755:
> This is an issue of sym(sqrt(2)) being not equal to sqrt(sym(2)) symbolically in octave.

That can never be fixed, but we can avoid converting existing sym objects back to doubles (as I have been trying to suggest).

Its easy to sort a list of SymPy objects in Python:
````
In [17]: L = [sqrt(S(3)), S(1)]

In [18]: sorted(L)
Out[18]: [1, sqrt(3)]
````
or you can use `L.sort()`.

**Comment by trueutkarsh** [comment]
@cbm755 sorting in python doesn't seem to work. Still it is converting it to rationals.
This is because even if one element in matrix is symbol, it call sym() constructor for every element in matrix as shown in below example.
```
octave:52> l=[sym(2),1,1]
l = (sym) [2  1  1]  (1×3 matrix)
octave:53> l(2)
ans = (sym) 1
```

**Comment by cbm755** [comment]
That example you gave is the correct and intended behaviour...  (previously, your code converted it *back* into a double: this was incorrect).

So this is looking much better I think.

1.  I don't think you should check `isallconstant`---sympy should raise an exception if asked to sort `[1 x 2]`.

2.  There are many tab characters---these will need to be replaced with spaces.

3.  Some formatting problems: e.g., please use a space after `;`.  And one space on each side of `=`.

File: inst/@sym/sort.m [hunk 11]
```diff
@@ -0,0 +1,99 @@
...
+%! assert (isequal (s, expected_s))
+
+%!test
+%! f = [sym(pi), sin(sym(2)), sqrt(sym(6))]; 
+%! s = sort(f); 
+%! expected_s = sym([sin(sym(2)), sqrt(sym(6)), sym(pi)]); 
+%! assert (isequal (s, expected_s))
+
+%!test 
+%! f = [sym(1), sym(2); sym(2), sym(pi); sym(pi), sym(1)];
```
cbm755: ummm, obviously I meant spaces after commas and semicolons within matrices, not at end of lines!

i.e.,
`[1 2; 3 4]`

Trailing whitespace is evil, don't do that.

cbm755:
I trimmed the whitespace and added a few tests.  This has been merged.

Thanks!

trueutkarsh:
@cbm755  I guess I made a mistake here. I had forced push this last commit hence I guess your changes are not visible here. Sorry about that. I was trying to remove the trailing spaces on my own, but by the time I had pushed my changes, yours were gone. :(

cbm755:
> I guess I made a mistake here. I had
> forced push this last commit hence I guess your changes are not visible

No mistake, I just did it on a different branch in the meantime and
merged it from that branch.  Your previous commits are all in.  Sorry
for the confusion.

PR #374, opened 2016-03-10
[wip] experimental PyTave communication

cbm755:
Instead of all the fragile stdin/stdout communication, we might try to use pytave which links directly to Python via C++ (specifically, boost).  Assuming it gets developed and maintained, it should be more robust than current approaches.

- [x] extend pytave to support the python `eval` and `exec` commands.
- [x] add a new `python_ipc_pytave` and `sympref` option.
- [x] for now, we just return strings, and parse our own way (with `extractblock`).
- [x] use pytave to return variables to octave, using pytave for non-sympy objects, and some process TBD for sympy objects.  Might be longer term.
- [x] clean up the debugging output and xml success messages.
- [x] use pytave to copy objects to Python.
- [x] fix *most* [issues labelled pytave](https://github.com/cbm755/octsympy/labels/pytave).
- [x] ~~Travis CI runs PyTave tests.~~  Defer until we have Octave 4.2 on Travius.


cbm755:
For anyone following along: here is the appropriate pytave branch: https://bitbucket.org/macdonald/pytave/branch/execeval

trueutkarsh:
@cbm755 can you elaborate a bit  more what are the goals in this ? Currently I've understood that we need to implement pyexec and pyeval first in pytave and then incormporate it here ..
Am I correct ?

cbm755:
My pytave branch on bitbucket already has rough implementations of pyexec and pyeval.  But more generally, improving pytave is a big part of this.  

There is an upcoming pytave code sprint: search the octave maintainers maillist.

trueutkarsh:
I had a doubt that, how does octsympy handles custom mathematical functions
objects e.g gamma function , airy etc. These functions cannot be written in
closed form
hence have to expressed as expression.
If sympy has a function say"gamma" how I can I use its instance in octsympy
?
Is there some direct way or I need to define it it explicitly in octave.
Similarly since there is no support for sympy in boost library ,which forms
the basis of pytave, other than returning just evaluations of functions , how should I go about
creating their instances and
and then using them .?
My present idea is to create a copy of their custom class (a wrapper) and
for each member function or operation , call the corresponding python
function. ? Can anything better be done.?

cbm755:
@genuinelucifer I committed a fix for zero-size matrices

cbm755:
@genuinelucifer I didn't do proper PR (sorry), but please review commits 5b47937 and defe9b2 (see also #483) and confirm you are happy with those.

genuinelucifer:
@cbm755 
I guess these changes may fix something in `pytav_ex` too.
To get these changes into pytave_ex (only the last 2 commits), should I do git cherry-pick or git merge? Which should be the preferred choice?

cbm755:
I think I would try to merge pytave branch onto pytave_ex: sometimes I first create a temporary branch and try to merge there (as I don't really know how to undo a merge, even locally).  Then if it works, I do it again on `pytave_ex`.

cbm755:
Thanks @genuinelucifer for your work on this.  I'm going to merge this.  Further improvements can happen on master.

PR #379, opened 2016-03-10
Improve portability of shell scripts and make rules

mtmiller:
Eliminate bashisms in shell scripts and options that are GNU extensions.

### DIFF
--- a/make_release_packages.sh
+++ b/make_release_packages.sh
@@ -15,19 +15,20 @@ MLDIR=$MLPKG
 
 echo "Making packages for octsympy-$VER."
 
-read -p "Press [Enter] to git clone and make packages..."
+printf "Press [Enter] to git clone and make packages..."
+read dummy
 
 
 # checkout a clean copy
 rm -rf octsympy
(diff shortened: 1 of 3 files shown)


cbm755:
Thanks!  The `make` stuff is most important. 

The other scripts were never really intended to be platform independent---just to automate my workflow.  `s|/bin/sh/|/bin/bash|` might've been easier... ;-)  

Have you tested?  I suppose I can just merge and fix up if necessary when I do the next release.

mtmiller:
Btw, this need comes out of the `make_release_packages.sh` script failing horribly for me with `/bin/sh` set to `/bin/dash` on my system (as on most Debian and Ubuntu systems).

I tested both scripts on `master` with bash and compared the output to this branch tested with both the `bash` and `dash` shells and I get identical output from all runs.

cbm755:
very thorough!  I shall personally try to be more careful and use `/bin/bash` by default.

mtmiller:
I actually prefer `/bin/sh`, keeps me honest, but the choice is yours :) Thanks!

PR #380, opened 2016-03-10
bootstrap: Include m4 macros directory, exit on error

mtmiller:


cbm755:
I don't know anything about this stuff so ok :-)

Issue #376, opened 2016-03-10
Support saving/loading `sym` objects from files


cbm755:
I've been meaning to look at saving `sym` objects to disk.  They are really just a structure with a few fields.  Only `x.pickle` (currently `char(x)` but #375 might change that) is actually important as far as SymPy is concerned.

1.  I've never promised API stability of that structure, might need to stabilize the structure a bit too.  Or at least read about loading/saving `@` classes.

2.  Should we rename `x.pickle` to `x.srepr`?  It was originally a pickle but has not been for a long time.  See also #269.

cbm755:
How to start: look at "saveobj" and "loadobj" methods.

genuinelucifer:
@cbm755 Should I develop 2 new functions ```saveobj``` and ```loadobj``` which would save the object (passed as argument) using python code?
else,
To let users save it using inbuilt ```save``` and ```load```, I'll have to implement saveobj function in the sym class. How do I do that?

cbm755:
> I'll have to implement saveobj function in the sym class. How do I do that?

Create a `inst/@sym/saveobj.m` file.  And `inst/@sym/loadobj.m`.

(See the GNU Octave documentation for what they are supposed to do.)

A `sym` objection is basically a `struct`.

For `loadobj`, there is a special "magic" form of calling `@sym/sym.m`
which activates the constructor (see the source of that file).

genuinelucifer:
Hi @cbm755 
I almost completed the functions.
I also installed octave on my linux machine so as to test the files before submitting a PR. I get the following error when I try to use any command (for any file in inst/@sym/)

cbm755:
Are you saying you cannot run tests on some other OS?  If so, please file a separate issue for that.

cbm755:
This error looks familiar...  Where are you running octave from?  should be `inst` (not `inst/@sym`).  Does that help?

genuinelucifer:
Thanks. It worked from inst directory :)

cbm755:
what about tests on Windows?  Can you run them or not?

genuinelucifer:
I guess yes. Only objective C tests are shown undefined. Other tests are
passing.

cbm755:
Closing

PR #393, opened 2016-03-31
Adding save and load functionality for sym objects

genuinelucifer:
Fixing #376 

- [x] Complete the functions and tests
- [x] Fix The doctests

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -66,6 +66,8 @@ octsympy 2.3.0-dev
 
   * Documentation improvements.
 
+  * Saving/Loading sym objects is now supported (it previously worked, by accident).
+
 
 
 octsympy 2.2.4 (2015-12-14)
--- a/inst/@sym/sym.m
+++ b/inst/@sym/sym.m
@@ -145,6 +145,9 @@
 %% @end group
 %% @end example
 %%
+%% It is also possible to save sym objects to file and then load them when
+%% needed in the usual way with the @code{save} and @code{load} commands.
+%%
 %% The underlying SymPy “srepr” can also be passed directly to
 %% @code{sym}: @pxref{char} for details.
 %%
@@ -673,3 +676,16 @@ end
 %! n = sym('n', 'negative', 'even');
 %! a = assumptions(n);
 %! assert(strcmp(a, 'n: negative, even') || strcmp(a, 'n: even, negative'))
+
+%!test
+%! % save/load sym objects
+%! syms x
+%! y = 2*x;
+%! a = 42;
+%! myfile = tempname();
+%! save(myfile, "x", "y", "a")
+%! clear x y a
+%! load(myfile)
+%! assert(isequal(y, 2*x))
+%! assert(a == 42)
+%! assert(unlink(myfile) == 0)


genuinelucifer:
@cbm755 Please review
.
Also, although I have worked with functions in octave but I was really shocked to see I am unable to define a function with 1 argument and then call it by passing an string as the argument. I always get the error that I mentioned. Any suggestions?

cbm755:
I will take a look.  Re: string argument: I'm not sure what you mean.  But please note any method defined as `@sym/method.m` must have at least one `sym` argument: that is how Octave identifies it as a method for `sym` object.  You'll note Symbolic has some functions in `inst/`: this is why.

Does that help?

cbm755:
Have you read http://www.gnu.org/software/octave/doc/v4.0.1/Manipulating-Classes.html?

I'm not sure saveobj is supposed to actually create a file: looks to me more like some preprocessing...  But I admit that I have not looked carefully at it.

Anyway, what we need to support is:
````
>> syms x
>> a = 42;
>> save myfile x a
>> clear
>> load myfile
>> x
  x = (sym) x
>> a
  a = 42
````

**Comment by genuinelucifer** [comment]
> any method defined as @sym/method.m must have at least one sym argument: that is how Octave identifies it as a method for sym object
Oh that's why I couldn't define it. Thanks.
.

> Have you read http://www.gnu.org/software/octave/doc/v4.0.1/Manipulating-Classes.html?
Sorry. I read the matlab documentation for the same. Now it is more clear. I will implement the functions and try again.

**Comment by genuinelucifer** [comment]
@cbm755 Please review
.
It works correctly now. And all the tests pass.
.
Also, I saw that there was no need to create these 2 files (except that these files provide some tests to check). The ```save myfile x``` and ```load myfile``` worked perfectly without doing anything...

**Comment by genuinelucifer** [comment]
Also, I have to rely on buildbot to check the doctests. (hence a few more pushes per PR).
I tried to install doctest but many errors arise.
I solved some of them but am stuck with one for now.
.
Whenever I run ```doctest @sym/somefile.m``` it always shows fail with the ```got``` part as:
```
got     : ??? 'doctest_evalc' undefined near line 140 column 29
```

cbm755:
I suspect that means you did not install doctest correctly (that file needs to be compiled).

`pkg install -forge doctest` is the recommended way to install packages.

I had no idea save/load was already working! :)  I will review this...

genuinelucifer:
I did install from forge only!
.
Well, here's how I did it:
- I ran ```sudo apt-get install doctest``` on linux terminal. After it got installed I remembered that it is not what I have to install. SO ,
- I ran ```pkg install -forge doctest``` in octave terminal. It showed many errors while installation and the installation did not complete.
- I ran doctest on linux terminal and found it complained about ```ghc``` and hence
- I ran ```sudo apt-get install ghc``` on linux terminal
- Then I went back to octave terminal and did ```pkg install -forge doctest``` and this time it installed perfectly!
- But now I am stuck with the error for ```got``` section!

cbm755:
doctest and octave-doctest not the same thing!  ghc should be unrelated.  Probably you dev environment (gcc, etc) is/was incomplete.

Please report the errors you got with `pkg install -forge doctest` at https://github.com/catch22/octave-doctest

cbm755:
> and this time it installed perfectly!

Seems unlikely (unless `ghc` just happened to have the same deps).  Can you `pkg uninstall doctest` and then try `pkg install -forge doctest`?

genuinelucifer:
@cbm755 It's late night for me. I'll surely check the errors and report them to the doctest repo tomorrow. :)

cbm755:
So, maybe we should not have saveobj/loadobj at all?  And put the tests somewhere else.  OTOH, if we every change the fieldnames then we'd need `loadobj` for backwards compatibility.  Not a review, just thinking out loud...

cbm755:
One review item: the tests will need to delete `myfile` after use.  Also, should check if Octave has a function to request a temp file name form the OS.

We probably don't need many tests here, just one should be enough.

Another thought: we could save only the `.pickle` (in `saveobj`) and then re-create the other fields in `loadobj`.  Advantages/disadvantages...?

genuinelucifer:
> So, maybe we should not have saveobj/loadobj at all? And put the tests
somewhere else.

Yes. That seems a better choice. Is there a common file for tests?

> OTOH, if we every change the fieldnames then we'd need loadobj for
backwards compatibility. Not a review, just thinking out loud...

Hmm :)

genuinelucifer:
> One review item: the tests will need to delete myfile after use. Also,
should check if Octave has a function to request a temp file name form the
OS.

Hmm. Will check for such functionality.

> We probably don't need many tests here, just one should be enough.

Will do.

> Another thought: we could save only the .pickle (in saveobj) and then
re-create the other fields in loadobj. Advantages/disadvantages...?

Advantage probably to reduce the amount of data to be stored (will have to
do some tests to see how much)
Disadvantage mostly perfect(probably?!) recreation of the object?!

cbm755:
> Is there a common file for tests?

Some stuff is in `misc.tst`.  Although I think at one point I was planning to move everything out of `misc.txt`

Another possibility is in `@sym/sym.m` (as the ctor is the main entry point of the class).

cbm755:
`NEWS` should have an new bullet point too: `* saving/loading sym objects is now supported (it previously worked, by accident).`

genuinelucifer:
@cbm755 I added the line in NEWS and also shifted the tests to ```@sym/sym.m```.
Now the tests use temp file names and also delete them after use.
Please review.
Should I squash the commits into one?

cbm755:
I think you need `save(myfile, 'x')` or something like that.  I think currently it will save it to literally `myfile`?  But strange that `unlink` doesn't fail and then make the test fail!

Maybe we don't need a doctest.  Just a sentence in the docs something like you have already:
`It is also possible to save sym objects to file and then load them when needed in the usual way with the @code{save} and @code{load} commands`.

+1 for squashing.

genuinelucifer:
> I think you need save(myfile, 'x') or something like that. I think
currently it will save it to literally myfile? But strange that unlink
doesn't fail and then make the test fail!

Strange indeed. I didn't thunk about that earlier. Will change it

> Maybe we don't need a doctest. Just a sentence in the docs something like
you have already:
> It is also possible to save sym objects to file and then load them when
needed in the usual way with the @code{save} and @code{load} commands.

Ohk will remove the doctest.

> +1 for squashing.

Will squash and push the changes in a couple of hours.

cbm755:
`help unlink`: test should fail if ERR is nonzero.

genuinelucifer:
@cbm755 Done.

cbm755:
Great!  This is in.

FYI: I try to maintain matlab compatibility so I avoid using `"` for strings (e.g., in your test).

genuinelucifer:
> Great! This is in.
>
> FYI: I try to maintain matlab compatibility so I avoid using " for
strings (e.g., in your test).
>

Oh. I'll try to remember that for future works.
Would you please point me to some other issue to work on?

PR #398, opened 2016-04-08
import some functions from specfun

cbm755:
This as numerical (`@double/@single`) functions for some functions, currently hosted by the [`specfun` pkg](http://octave.sourceforge.net/specfun/).

This is *not* a hostile takeover ;-) see https://savannah.gnu.org/patch/?8644

For this PR, I will do:
- [x] laguerreL
- [x] heaviside
- [x] dirac
- [x] lambertw


cbm755:
The implementation of `@double/laguerreL` was numerically unstable, so I fixed that.

PR #410, opened 2016-04-13
added yn and jn 

trueutkarsh:
more test cases to be added.

### DIFF
--- a/inst/@sym/besselyn.m
+++ b/inst/@sym/besselyn.m
@@ -18,28 +18,28 @@
 
 %% -*- texinfo -*-
 %% @documentencoding UTF-8
-%% @deftypefn  {Function File} {@var{Y} =} yn (@var{alpha}, @var{x})
+%% @deftypefn  {Function File} {@var{Y} =} besselyn (@var{alpha}, @var{x})
 %% Symbolic Spherical Bessel function of the second kind.
 %%
 %% Example:
 %% @example
 %% @group
 %% syms n x
-%% A = 2*yn(n, x)
-%%   @result{} A = (sym) 2⋅yn(n, x)
+%% A = besselyn(n, x)
+%%   @result{} A = (sym) yn(n, x)
 %% diff(A)
 %%   @result{} ans = (sym)
 %%   
-%%                      2⋅(n + 1)⋅yn(n, x)
-%%     2⋅yn(n - 1, x) - ──────────────────
-%%                           x   
+%%                   (n + 1)⋅yn(n, x)
+%%    yn(n - 1, x) - ────────────────
+%%                          x 
 %% @end group
 %% @end example
 %%
-%% @seealso{besselj, besseli, besselk, jn}
+%% @seealso{besselj, besseli, besselk, besseljn}
 %% @end deftypefn
 
-function Y = yn(n, x)
+function Y = besselyn(n, x)
   if (nargin ~= 2)
     print_usage ();
   end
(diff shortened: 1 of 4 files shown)


trueutkarsh:
@cbm755  could you suggest more test cases that can be added ?

cbm755:
These won't pass: roundtrip not possible AFAIK b/c jn not implemented for double.

So test against some numerical value: see sinint/cosint in my other branch.  Or maybe better to lookup the formula in terms of non spherical Bessel.

cbm755:
How about we call these "besseljn" (for now)

inst/@sym/jn.m
```diff
@@ -0,0 +1,61 @@
...
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @deftypefn  {Function File} {@var{Y} =} jn (@var{alpha}, @var{x})
+%% Symbolic Spherical Bessel function of the first kind.
+%%
+%% Example:
+%% @example
+%% @group
+%% syms n x
+%% A = 2*jn(n, x)
```
cbm755: drop this 2 (I sometimes use that to avoid a divide by 2 in the result, but here it doesn't seem its needed.

cbm755:
> These won't pass: roundtrip not possible AFAIK b/c `jn` not implemented for double.

I apologize, that is not quite true.  SymPy generates code with the` sqrt(x)` factor, so function_handle will work correctly and should be tested.  But there is no double `yn(2,10)` in Octave so some changes to the test are needed.

trueutkarsh:
@cbm755  function_handle did'nt worked on my machine explicitly when I had tried to run each line of test case one by one, still the test cases were passed. This should'nt happen right ?

cbm755:
It should not.  Unless your sympy is earlier than 1.0 (as the test is skipped in that case)...

trueutkarsh:
@cbm755  yeah sympy version was the case. Now the test fails. abs(A-B) is of order 10^-9 and eps is itself of order 10^16. Should I put a custom 10^-8  for  checking  double(A-B) for the test case to pass?
Or something else can be done ?
There's nothing to be done in terms of changing the function besselyn other than representing yn in terms of besselj and bessey ?

inst/@sym/besselyn.m
```diff
@@ -0,0 +1,62 @@
...
+  Y = binop_helper(n, x, 'yn');
+end
+
+
+
+%!test
+%! % roundtrip
+%! if (python_cmd ('return Version(spver) >= Version("1.0")'))
+%! syms x
+%! A = besselyn(sym(2), sym(10));
```
cbm755: Try: `A = double(besselyn(...))`

cbm755:
I don't think A is a double, see comment above.

(its possible that the generated code is not the best numerical implementation of yn: may need to worry about that later, not for this PR)

cbm755:
I got:
````
>> h(10) - double(A)
ans =   -2.7756e-17
````

File: inst/@sym/besselyn.m [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+%% @group
+%% syms n x
+%% A = besselyn(n, x)
+%%   @result{} A = (sym) yn(n, x)
+%% diff(A)
+%%   @result{} ans = (sym)
+%%   
+%%                   (n + 1)⋅yn(n, x)
+%%    yn(n - 1, x) - ────────────────
+%%                          x
```
cbm755: This is kind of picky, but please indent the `yn(n - 1, x) - ....` bit to line up under the `u` of `@result{}`

trueutkarsh:
@cbm755 
the "double(besselyn(...))" works for besselyn and besseljn.
But for besseljn for "A=double(besseljn(sym(2), sym(9)))" and "B=h(9)" the test cases are passed but
but not for "A=double(besseljn(sym(2), sym(10)))" and "B=h(10)". since both are of order of 10^-16
hence this is possible.

This leads to doubt that test case may fail for besselyn also for some other value of evaluation other than (2,10). 
what can we do about it ?

cbm755:
OK so use some value other than 10 where the quantity is O(1)

cbm755:
Looks good, merging.  We can always add more tests later.  And we can always rename later, especially easy if done before next release.

Issue #394, opened 2016-04-03
tests that create temp files should use "tempname"


cbm755:
ccode, fcode may have hardcoded filenames.  Maybe others?

@genuinelucifer interested?

genuinelucifer:
>
> @genuinelucifer interested?
>
>

Will do.

genuinelucifer:
Hi @cbm755 .
I got busy with some other project and hence the delay in following up.
.
I looked up ```ccode``` but it seems that it doesn't create a file if no name is passed. It just returns the contents in 2 variables. So, I don't think it would be good idea to first create a temp file and then read it back.
Also, I could not find any file for ```fcode```...

cbm755:
> Also, I could not find any file for |fcode|...

Maybe it was "function_handle"?  Look at the output from
`octsympy_tests` and you'll see something writing temp files.

genuinelucifer:
Thanks. Found it.

genuinelucifer:
@cbm755 Probably #411 completes this...

cbm755:
thanks!

PR #411, opened 2016-04-14
Make functions use tempname to create new files

genuinelucifer:
Fixing #394 
Had to change how the file is written in order to properly load a files from ```/tmp/```

- [x] Make fcode use tempname

### DIFF
--- a/inst/@sym/function_handle.m
+++ b/inst/@sym/function_handle.m
@@ -117,7 +117,6 @@ function f = function_handle(varargin)
 
     [fcnpath, fcnname, fcnext] = fileparts(param.fname);
     fname2 = fcnname;
-    %fname2 = param.fname; fcnname = param.fname;
     % was old note about findsymbols vs symvar: not relevant
     [worked, out] = python_cmd (cmd, varargin(1:Nout), fcnname, fname2, param.show_header, inputs);
 
--- a/inst/@sym/function_handle.m
+++ b/inst/@sym/function_handle.m
@@ -115,7 +115,10 @@ function f = function_handle(varargin)
       param.fname = param.fname(1:end-2);
     end
 
-    fname2 = param.fname; fcnname = param.fname;
+    [fcnpath, fcnname, fcnext] = fileparts(param.fname);
+    fcnname (fcnname == '-') = '_';
+    fname2 = fcnname;
+    %fname2 = param.fname; fcnname = param.fname;
     % was old note about findsymbols vs symvar: not relevant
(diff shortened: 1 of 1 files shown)


genuinelucifer:
@cbm755 Please review. I'm looking into other files as well. Will add them soon.

genuinelucifer:
I've gone through all the files and no other file is creating any new file. So, I guess this PR is complete.
.
Please let me know if I missed something.

inst/@sym/function_handle.m
```diff
@@ -115,7 +115,10 @@
       param.fname = param.fname(1:end-2);
     end
 
-    fname2 = param.fname; fcnname = param.fname;
+    [fcnpath, fcnname, fcnext] = fileparts(param.fname);
+    fcnname (fcnname == '-') = '_';
```
cbm755: let's not do this here: if user wants `-` let her do so...  (We'll fix the tempname thing elsewhere...)

inst/@sym/function_handle.m
```diff
@@ -246,23 +253,27 @@
 %!test
 %! % output to disk
 %! fprintf('\n')
-%! f = function_handle(2*x*y, 2^x, 'vars', {x y z}, 'file', 'temp_test_output1');
+%! temp_file = tempname();
+%! temp_file(temp_file =='-') = '_';
```
cbm755: `temp_file = tempname('', 'oct_')`

I'm not sure its guaranteed to have to `-` but at least it doesn't by default ;-)

inst/@sym/function_handle.m
```diff
@@ -130,11 +133,15 @@
...
+    [fid,msg] = fopen(file_to_write, 'w');
     assert(fid > -1, msg)
     fprintf(fid, '%s', M.code)
     fclose(fid);
-    fprintf('Wrote file %s.\n', M.name);
+    fprintf('Wrote file %s.\n', file_to_write);
+    if (exist(fcnpath))
+        addpath(fcnpath);
+    end
     f = str2func(fcnname);
```
cbm755: this is scary: I sort of think its the user's responsbility, if she wants.  Plus added `/tmp` to the path is bound to be some sort of code-injection security bug (Bob create `/tmp/ls.m` which contains bad code for Alice to run etc).

genuinelucifer:
@cbm755 I did the changes. Please review again.

cbm755:
How do we know `tempname` will be in `/tmp`?

genuinelucifer:
> How do we know tempname will be in
> /tmp?

Well, i always found them there :P
.
I changed it now. If its okay, i could squash the commits.

cbm755:
I think this is ok as-is.  I'll merge it

PR #414, opened 2016-04-19
configure: only warn on missing Python/SymPy

cbm755:
Reasons include: 1. its only needed at runtime; 2. not all our
features need Python (double-only special functions implementations
for example); 3. make it easer for folks on alternate OSes to debug.


cbm755:
error is doctest I've broken elsewhere.  Will merge.

Issue #412, opened 2016-04-18
Mac OS X installation is difficult


cbm755:
on Octave help list Ben Abbott said:
 
> Octave’s symbolic toolbox is tricky to install on OS X. I’m confident it
> can be done, but haven’t yet done it myself. My difficulty is due to
> Fink’s python which is named “python2.7”. In order for symbolic-2.3.0
> toolbox’s configure process to use Fink’s python2.7, it is necessary to
> modify configure.ac …
> 
> 16 
> -17 AC_PATH_PROG(PYTHON, python)
> +17 AC_PATH_PROG(PYTHON, python2.7)
> 18 if test -z "$PYTHON"; then
> 19   AC_MSG_ERROR(Failed to find Python)
> 20 fi
> 
> You’ll need to replace “python2.7” with the command needed to run python
> in the Anaconda environment.
> 
> Next, run ./bootstrap in order to create the configure script that uses
> the desired python command.
> 
> Then create a tarball for symbolic-2.3.0 using the corrected configure
> script.
> 
>    tar -z -c -f symbolic-2.3.0.tar.gz symbolic-2.3.0
> 
> The symbolic toolbox requires sympy >=0.7.5. Fink provides version
> 0.7.3. Thus, I created a local Fink package for version 0.7.6.1 and
> installed it. 
> 
> I then installed a local version of the Symbolic toolbox
> 
>    pkg install -local symbolic-2.3.0.tar.gz

We should have a better story than that!

carandraug:
If there is already both a python and a python2.7 installed on the system (am I correct in assuming that this is what is happening here), how does the symbolic package then picks the right python during runtime?

Also, there is no need to change the sources. `AC_PATH_PROG` will check if the variable already exists and if so, skip the test. So one should be able to do something like:

    export PYTHON=/path/to/my/python2.7
    octave --eval 'pkg install symbolic-2.3.0.tar.gz'

or using `setenv` from within Octave (note that it expects `PYTHON` to be a path so if you set `PYTHON` to just the bin name, it will not work and look for python only in the path anyway).

bpabbott:
Ok. I tried "export PYTHON=python2.7", but that did not work. Is it necessary to include the entire path?

carandraug:
> Ok. I tried "export PYTHON=python2.7", but that did not work. Is it necessary to include the entire path?

[...]

Issue #415, opened 2016-04-19
Allow using env vars to specify python executable


cbm755:
See #412.  Specifically https://github.com/cbm755/octsympy/issues/412#issuecomment-211596214.

I'd like this to work at:

1. at run-time: either replacing or working with `sympref python /my/path/python`.

2. during any install-time time configuring.

cbm755:
2. seems to be already working, according to comments in #412.

PR #418, opened 2016-04-20
Support PYTHON env variable

cbm755:
Fix for #415.

@genuinelucifer are you able to test this on windows?  The python executable name is passed to `winwrapy.bat` as a command line argument: should work but I've not tested it!

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -29,6 +29,9 @@ octsympy 2.4.0-dev
   * Add spherical Bessel functions: these are called `besseljn` and
     `besselyn` (for now, subject to change).  Thanks to Utkarsh Gautam.
 
+  * The environment variable `PYTHON` can control which Python executable
+    is used.  Changing the Python executable should work on Windows.
+
 
 
 octsympy 2.3.0 (2016-04-08)
--- a/bin/winwrapy.bat
+++ b/bin/winwrapy.bat
@@ -5,7 +5,7 @@ REM This is a workaround as Octave for Windows cannot tolerate stderr from a pop
 REM subprocess: see [this bug report](https://savannah.gnu.org/bugs/?43036).
 REM You'll need to edit this script if your python is called something else.
 
-python.exe -i 2> NUL
+%1 -i 2> NUL
 
 REM If you are using the windows bundle for octsympy then "octpy.exe" will be used.
 REM It is [available online](http://www.orbitals.com/programs/pyexe.html)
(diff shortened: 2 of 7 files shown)


cbm755:
If I've understood correctly, this also fixes #412.

genuinelucifer:
How do I test it?
.
If i run make_windows_package then it always downloads master branch.
Should I zip the files myself?
Sorry, I haven't yet quite got the hang of how the package is made.

cbm755:
hmm, good point.  Really need to improve that build script I guess!  I think for now you'll have to hack the script to use this branch instead: must be a way to pass something like `-b fix415` to git somewhere in the script.

cbm755:
On GNU/Linux, one can just run Octave in the `inst/` directory of your developement directory.  Does that work on Windows?

genuinelucifer:
> On GNU/Linux, one can just run Octave in the inst/ directory of your
developement directory. Does that work on Windows?

I'll try and let you know (hopefully within half an hour)..

cbm755:
Wait, also we don't need to test with the "bundle": should be easier not to.  Although I guess we'll need to test the bundle too before next release.

genuinelucifer:
> we don't need to test with the "bundle":

Hmm. Sounds good.

genuinelucifer:
I uninstalled symbolic from octave and then changed the current directory to `/inst' but still got the error... (I pulled the branch `fix415` and switched to it before doing this).
```
>> pwd
ans = E:\My_Repos\octsympy\inst
>> syms x
OctSymPy v2.4.0-dev: this is free software without warranty, see source.
Detected Windows: using "winwrapy.bat" to workaround Octave bug #43036
error: popen2: process creation failed
error: called from
    python_ipc_popen2 at line 84 column 21
    python_ipc_driver at line 57 column 13
    sym at line 365 column 5
    syms at line 162 column 9
error: called from
    syms at line 162 column 9
```

genuinelucifer:
I couldn't try doing much changed today. I'll definitely try to find some workaround for this tomorrow. And post any findings.

cbm755:
Maybe you have to move winwrapy into that dir too?  Or put it in path or something

genuinelucifer:
> Maybe you have to move winwrapy into that dir too? Or put it in path or something

I had to copy paste winwrapy.bat into `/inst` and it worked as U got this output:
```
>> syms x
OctSymPy v2.4.0-dev: this is free software without warranty, see source.
Detected Windows: using "winwrapy.bat" to workaround Octave bug #43036
Some output from the Python subprocess (pid 2812) might appear next.

Closing the Python pipe...
error: Python exception: ImportError: No module named 'sympy'
    occurred in python_header import block
error: called from
    python_cmd at line 178 column 5
    sym at line 365 column 5
    syms at line 162 column 9
error: called from
    syms at line 162 column 9
```

cbm755:
I'll merge this for now.

I filed #419.

@mtmiller @bpabbott feel free to continue to discuss saving of configuration from the `./configure` stage if you wish.  I don't think I have anything to add to that discussion at the moment.

PR #420, opened 2016-04-21
update texinfo documentation to be more precise

cbm755:
@oheim can you review these changes before I make them more widely?

I haven't dropped some output variables yet (e.g., `times`) but I might do so.

Questions:

1. for `class` in `@defop`, why use `{@@infsup}` instead of just `@@infsup`?

2. catalan: would you rather use category `Constant` here?  It is not a constant in the programming sense of the word: its a function that returns a mathematical constant.


cbm755:
(sorry there are some other unrelated doc changes mixed in here---just ignore those)

oheim:
Looks good. You have read my posting one the mailing list, haven't you ;-)

1. Since it is a single word, you probably don't need the curly brackets. Also I was unsure about not putting curly braces around the argument list (e. g. in ```diff```), but Texinfo manual says you can use a list of arguments (until the end of the row), so this is probably fine as well.
2. I'd use category ```Constant``` for functions that return a constant (since we have no true constant in Octave otherwise)

cbm755:
> You have read my posting on the mailing list, haven't you ;-)

I shore 'av pardner!  Thanks.

cbm755:
Merging for now, lots more to be done here but I'll file another bug for that.

PR #422, opened 2016-04-21
Check that we have Python before starting the IPC

### LINKED ISSUE #417: check if python actually exists before system/popen2 calls
At run-time, we could/should check that `sympref('python')` is something we can run before opening the pipe.  I.e., it exists, or we could call some simple test.

See https://github.com/cbm755/octsympy/issues/413#issuecomment-212169061

cbm755:
Fixes #417.  For the system-based mechanism, we only do this once
to avoid overhead with each call.

### DIFF
--- a/inst/private/assert_have_python_and_sympy.m
+++ b/inst/private/assert_have_python_and_sympy.m
@@ -16,7 +16,7 @@
 %% License along with this software; see the file COPYING.
 %% If not, see <http://www.gnu.org/licenses/>.
 
-function assert_we_have_python(pyexec)
+function assert_have_python_and_sympy(pyexec)
 %private function
 
   [st, out] = system([pyexec ' -c "a = 42"']);
@@ -27,4 +27,22 @@ function assert_we_have_python(pyexec)
            '    Is Python installed?  Is your "path" configured correctly?'], ...
           pyexec)
   end
+
+
+  minsympyver = '0.7.6';
+
+  [st, out] = system([pyexec ' -c "import sympy; print(sympy.__version__)"']);
+
+  if (st ~= 0)
+    error('OctSymPy:nosympy', ...
+          'Python cannot import SymPy: have you installed SymPy?')
+  else
+    spver = strtrim(out);
+    if (compare_versions (spver, minsympyver, '<'))
+      error('OctSymPy:oldsympy', ...
+            'SymPy is installed but is too old (sympy-%s found, %s required)', ...
+            spver, minsympyver)
+    end
+  end
+
 end
--- a/inst/private/assert_we_have_python.m
+++ b/inst/private/assert_we_have_python.m
@@ -1,0 +1,30 @@
+%% Copyright (C) 2016 Colin B. Macdonald
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+function assert_we_have_python(pyexec)
+%private function
+
+  [st, out] = system([pyexec ' -c "a = 42"']);
+  if (st ~= 0)
+    error('OctSymPy:nopython', ...
+          ['Cannot run the Python executable "%s"\n' ...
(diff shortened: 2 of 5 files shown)


cbm755:
@mtmiller @genuinelucifer can you each review this?  @genuinelucifer would be awesome if you could test in Windows.

In particular, let me know if either of you have a better idea.

1. `exist` in Octave is the wrong tool: $PYTHON could be `/usr/bin/python` or just `python`.  Also `python` itself could be a shell alias.

2. In bash (and others shells I think), I can do `command -v python`.  But Windows... and we'd have to `system` anyway...

cbm755:
Here's how it works:
````
>> setenv PYTHON /usr/loco/wtf/is/python42
>> sympref reset
Resetting the communication mechanism
>> syms x
OctSymPy v2.4.0-dev: this is free software without warranty, see source.
sh: /usr/loco/wtf/is/python42: No such file or directory
error: Cannot run the Python executable "/usr/loco/wtf/is/python42"
    Python and SymPy are needed for most Symbolic features.
    Is Python installed?  Is your "path" configured correctly?
error: called from
    assert_we_have_python at line 24 column 5
    python_ipc_popen2 at line 78 column 5
    python_ipc_driver at line 57 column 13
    sym at line 365 column 5
    syms at line 162 column 9
>> 
````

**Comment by mtmiller** [comment]
@cbm755 I'll test some weird cases I came up with to see how it performs. At a glance looks right to me, you are right that invoking the interpreter is the best test rather than testing for file existence.

Is it overkill to do a test script that makes sure it's actually Python, something a little more than `a = 42`? Maybe something like `import sys; assert(type(sys.version) == str)`?

**Comment by cbm755** [comment]
Thanks Mike.  Also, maybe we could make another separate `system` call to check that `import sympy` works too...

**Comment by genuinelucifer** [comment]
@cbm755 It seems to be working correctly! :)
```
>> setenv PYTHON e:/some/dir/python45
>> sympref reset
Resetting the communication mechanism
>> syms x
The system cannot find the path specified.
OctSymPy v2.4.0-dev: this is free software without warranty, see source.
error: Cannot run the Python executable "e:/some/dir/python45"
    Python and SymPy are needed for most Symbolic features.
    Is Python installed?  Is your "path" configured correctly?
error: called from
    assert_we_have_python at line 24 column 5
    python_ipc_popen2 at line 78 column 5
    python_ipc_driver at line 57 column 13
    sym at line 365 column 5
    syms at line 162 column 9
error: called from
    syms at line 162 column 9
```

**Comment by genuinelucifer** [comment]
> maybe we could make another separate system call to check that import sympy works too...

 :+1:

**Comment by cbm755** [comment]
Thanks @genuinelucifer for testing on Windows.  Compared with my output on Fedora, it looks like the standard error might be hidden with the `system` command.  But that shouldn't matter: we tell the user enough in the error message.

**Comment by cbm755** [comment]
It now checks for SymPy:
````
OctSymPy v2.4.0-dev: this is free software without warranty, see source.
  File "<string>", line 1, in <module>
ImportError: No module named sympy
error: Python cannot import SymPy: have you installed SymPy?
error: called from
    assert_have_python_and_sympy at line 37 column 5
    python_ipc_popen2 at line 78 column 5
    python_ipc_driver at line 57 column 13
    sym at line 365 column 5
    syms at line 162 column 9
````

And the version is checked as well (fake version used for demo):
````
OctSymPy v2.4.0-dev: this is free software without warranty, see source.
error: SymPy is installed but is too old (sympy-1.0.1.dev found, 10.0.0 required)
error: called from
    assert_have_python_and_sympy at line 42 column 7
    python_ipc_popen2 at line 78 column 5
    python_ipc_driver at line 57 column 13
    sym at line 365 column 5
    syms at line 162 column 9
````

If either of you have suggestions for better text, let me know.  One thought is to point to a FAQ or wiki entry.

**Comment by mtmiller** [comment]
LGTM. I get similar output as @cbm755.

One interesting case:

    >> setenv PYTHON true
    >> syms x y z
    OctSymPy v2.4.0-dev: this is free software without warranty, see source.
    error: compare_versions: version numbers V1 and V2 must be a single row
    error: called from
        compare_versions at line 84 column 5
        assert_have_python_and_sympy at line 41 column 5
        python_ipc_popen2 at line 78 column 5
        python_ipc_driver at line 57 column 13
        sym at line 365 column 5
        syms at line 162 column 9

At least it doesn't hang like it used to, but error is not clear in this case (`PYTHON` is set to some program that doesn't error, but also doesn't produce meaningful output that looks like a working Python).

**Comment by cbm755** [comment]
Maybe something like:
````
  [st, out] = system([pyexec ' -c "print(\"hello world\")"']);
  if (st ~= 0)
    error(...)
  end
  if (~ strcmp(strtrim(out), 'hello world'))
    warning('unexpected output, perhaps something is wrong with your python?')
  end
```

cbm755:
(I'm not going to try to protect against malicious $PYTHON but good to prevent accidents)

Or:
````
if (st ~= 0) || (~ strcmp(strtrim(out), 'hello world')))
  error(...)
````

mtmiller:
Yeah that looks good.

(I'm also not suggesting to go overboard with protective checking, just basic accidental usage.)

cbm755:
Done in 674dfc1aa915bded81d8cc389a0ca9a6033f5d4e

PR #425, opened 2016-04-25
Sinint

cbm755:
This isn't very good (not accurate enough), its just an import from specfun package with (failing) tests added.


latot:
maybe this can be closed? actually exist a Sinint in ```@sym/sinint``` or this pr have other objetive?

cbm755:
Other objective.  This was a `@double/sinint` implementation in pure Octave (as we are merging stuff from the old `specfun` package).  But its poor code.  Instead I did a `@double/sinint` that calls sympy (actually mpmath) which is kinda slow.

Longer-term, someone might want a "proper" Octave-native sinint.  I'll leave this open for a bit, as I'm not sure this story is

Issue #419, opened 2016-04-21
deprecate `sympref python meh` in favour of $PYTHON


cbm755:
@mtmiller:

> But yes, now that you are using the value of `$PYTHON` from the environment, you could drop
> `sympref python` completely (or have it display a short note to the user that they should use
> `setenv PYTHON` now).

bpabbott:
I had assumed that $PYTHON was only respected by ./configure. In which case, it would be great if the default preference was set with $PYTHON used by configure.

cbm755:
The proposal here is to have no preference: just to respect $PYTHON.

bpabbott:
Understood. I'm concerned that such a change will prevent a package manager from installing a functional symbolic package.

cbm755:
Hmm that sounds bad.  Can you clarify?  Would it be worse than what we have now?

There is (now) a line in `sympref.m` that sets: `DEFAULTPYTHON = 'python'`.  A sufficiently advanced configure script (i.e., not what we have) could presumably change that at install time to `/opt/sw/weird/path/for/packages/python` or whatever.

One example is the "bundle" windows package: preparing that invokes `sed` to change the `DEFAULTPYTHON` to `octpy.exe` (which we ship)---that would still work under this proposal.  
Packagers could do something similar (if necessary).

But if the user sets $PYTHON and that breaks things, isn't that still what she intended?

bpabbott:
I think it would be worse. What we have now could be patched to modify the default. Using <sed> for that is a good solution.

cbm755:
Am I missing something?  Right now, packager can use `sed`.  With this proposal packager can use `sed`.  What is the difference?

Maybe I should do a branch with this change so we have something specific to compare...

bpabbott:
With the present approach, sed can be used to change the preference.. And this change would be private to the symbolic package,

With $PYTHON the shell environment is public.

cbm755:
But sed will still work...

The only difference I can think of is a user who set PYTHON (e.g., for some other program) and then is surprised to find Symbolic uses that python too.

But I think this is a bit like PAGER or EDITOR: I think users generally expect programs to respect those env vars (rather than providing some other configuration mechanism.)

cbm755:
Closed by #427.

PR #427, opened 2016-04-26
replace sympref with setenv for python cmd

cbm755:


cbm755:
Fixes #427.  Although maybe one should deprecate for one release rather than just drop.

cbm755:
Umm, I mean fixes #419

PR #432, opened 2016-04-30
Check for file existence before calling str2func

### LINKED ISSUE #431: test @sym/function_handle failed
```
Wrote file C:\Users\SnD\AppData\Local\Temp\oct_r1ACyg.m.
***** test
 % output to disk
 fprintf('\n')
 if (exist ('octave_config_info', 'builtin'))
   temp_file = tempname('', 'oct_');
 else
   temp_file = tempname();
 end
 % allow loading function from temp_file
 [temp_path, ans, ans] = fileparts(temp_file);
 addpath(temp_path);
 f = function_handle(2*x*y, 2^x, 'vars', {x y z}, 'file', temp_file);
 assert( isa(f, 'function_handle'))
 [a,b] = f(10,20,30);
 assert (isnumeric (a) && isnumeric (b))
 assert (a == 400)
 assert (b == 1024)
 assert (unlink([temp_file '.m']) == 0)
 % remove temp_path from load path
 rmpath(temp_path);
!!!!! test failed
@oct_r1ACyg: no function and no method found
shared variables

x = (sym) x
y = (sym) y
z = (sym) z
```

genuinelucifer:
Fixes #431 

Although I don't think this is a nice solution, it is here just for temporary solution.
I added `fflush(fid)` but that doesn't seem to do the job. So, I had to use the `while`.

Probably we have to reload some sort of global cache for files in octave?

### DIFF
--- a/inst/@sym/function_handle.m
+++ b/inst/@sym/function_handle.m
@@ -134,12 +134,17 @@ function f = function_handle(varargin)
     fclose(fid);
     fprintf('Wrote file %s.\n', file_to_write);
     
-    fprintf('Checking for the function to exist... ');
+    %FIXME: Check upstream to rehash the files correctly once created
+    if(exist('OCTAVE_VERSION') && ispc())
+    fprintf('Due to an upstream bug in octave on windows, we have to wait for the file to be loaded.\n');
+    fprintf('See https://savannah.gnu.org/bugs/?31080 for more information...\n');
+    fprintf('Finding function %s... ', fcnname);
     while (exist(fcnname) == 0)
     end
-    fprintf('Found!\n');
-    f = str2func(fcnname);
+    fprintf('Found!\n\n');
+    end
     
+    f = str2func(fcnname);
 
   else % output function handle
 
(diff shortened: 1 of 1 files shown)


genuinelucifer:
I tried `rehash` but it didn't work either...

mtmiller:
This is a really bad workaround for what may be a Windows filesystem or MinGW problem. `fflush` should be implied by `fclose`, and after a file is closed it should `exist`.

That said, may be necessary even if we find the cause and fix in Octave, if you want to not depend on a future version of Octave.

genuinelucifer:
@cbm755 Please check if the message is how it's required... and anything else that needs to be edited.

cbm755:
1.  check other files for how to do the `OCTAVE_VERSION` check.
2.  Do you think we need to output anything here?  Maybe that text can just be a comment?  Unless there is a good reason to print it: for example, if they delay could be longer than say 10s.

genuinelucifer:
>  if they delay could be longer than say 10s.

Actually the whole `test @sym/function_handle.m` takes 40-45 sec on my PC.
And the output is shown only after the test is complete. Can I flush stdout somehow?

genuinelucifer:
Well, I found the `fflush(stdout)` from stackoverflow.

Also, if I let the while loop be empty then it took approx 102 seconds but if I call rehash in the while then it took only 12.518 seconds!!! That's why I added the rehash...

cbm755:
Ok, I reformatted slightly and put a pause in, see aeabaf8442bbe8943ea2ff179b5b10df48ecacb1

genuinelucifer:
Thanks for the edit.
What should I work on next?

PR #434, opened 2016-05-01
Use canonical test for Octave on OCTAVE_VERSION

mtmiller:
Test for existence of "OCTAVE_VERSION" instead of "octave_config_info",
recommended best practice by upstream, and the latter is deprecated in
Octave 4.1+.


cbm755:
Thanks!

PR #451, opened 2016-05-23
Only use winwrapy.bat on Octave <= 4.0.2

### LINKED ISSUE #449: use winwrap.bat only on older Octave versions
See #82.  This will also fix #316.

* Current (2016-05) dev versions of Octave are fixed (what will become 4.2 I think).

* Hopefully, fix will appear in 4.0.3 as well.

So we need a version check in `python_ipc_popen2.m` to use `winwrap.bat` only when version `< 4.0.3`

cbm755:
Fixes #449.


cbm755:
Briefly tested on Windows 10 with a patched Octave 4.0.2 binary.  This seems to work!

PR #452, opened 2016-05-24
Add prelimnary support to convert to octave types at one place.

genuinelucifer:
I'm putting it here to get your feedback and guidance.
.
I have many questions which I'll put inline as comment.


inst/private/check_and_convert.m
```diff
@@ -0,0 +1,36 @@
...
+  pyexec(str_check_is_sym);
+  is_sym = pyeval('is_sym');
+  
+  if is_sym
+    ascii = pyeval('sp.pretty(_outs[0], use_unicode=False)');
+    unicode = pyeval('sp.pretty(_outs[0], use_unicode=True)');
+    str = pyeval('sympy.srepr(_outs[0])');
+    flat = pyeval('str(_outs[0])');
+    
+    str_eval = strjoin({'temp = Matrix([_outs[0]])',
```
genuinelucifer: I donno why but if I don't cast it to Matrix then I always get the error that 
`'Symbol' object has no property shape`

inst/private/check_and_convert.m
```diff
@@ -0,0 +1,36 @@
+function obj = check_and_convert()
+  newl = sprintf('\n');
+  
+  str_check_is_sym = strjoin({'if temp is None or isinstance(temp, (sp.Basic, sp.MatrixBase)):',
```
genuinelucifer: Why did you check for None in your code? I couldn't replicate a case to get None. I put it here just to follow your code.

inst/private/check_and_convert.m
```diff
@@ -0,0 +1,36 @@
...
+                        '#function call to pyobj with params not implemented yet',
+                        'd0 = _d[0]',
+                        'd1 = _d[1]'}, newl);
+    pyexec(str_eval);
+    _d = [1 1];
+    _d(1) = pyeval('d0');
+    _d(2) = pyeval('d1');
+    
+    obj = sym([], str, [_d(1) _d(2)], flat, ascii, unicode);
+  else
```
genuinelucifer: why else? When/How can we reach this case?
When we issue commands related to symbols then only python_ipc_pytave does something. Else it is never called. What am I missing?

inst/private/python_ipc_pytave.m
```diff
@@ -61,25 +61,28 @@
...
   loc = python_copy_vars_to('_ins', true, varargin{:});
   s = strjoin(loc, newl);
-  pyexec(s)
+  pyexec(s);
 
   s = strjoin(cmd, newl);
-  pyexec(s)
+  pyexec(s);
+  
+  info.raw = pyeval('_outs[0]');
```
genuinelucifer: Why are we storing the value in `info.raw`? I didn't see it being used somewhere...

inst/private/python_ipc_pytave.m
```diff
@@ -61,25 +61,28 @@
...
+  A = check_and_convert();
```
genuinelucifer: I'll change the name from `check_and_convert` to something better.. :P

inst/private/python_ipc_driver.m
```diff
@@ -51,6 +51,8 @@
       which_ipc = 'system';
     end
   end
+  
+  which_ipc = 'pytave';
```
genuinelucifer: did just to test. Is there a better way to do this? (Sorry, I didn't look much into it)

inst/private/check_and_convert.m
cbm755: `None` is a bit of special case: you can probably ignore for now, or see `@sym/isNone` for tests.  Basically, its not a SymPy object but it can get returned...

inst/private/check_and_convert.m
cbm755: I'm not sure yet... perhaps this is related to `object_filter` in `python_header.py`?

inst/private/check_and_convert.m
cbm755: See tests in `python_cmd`: we could get back e.g., a dict or an `int` (not  `Integer`).   `python_ipc_pytave` is *not* only used for `@sym`.

inst/private/python_ipc_driver.m
cbm755: `sympref ipc pytave`

but this is fine for our testing.  Maybe slightly nicer to change the default in `sympref` (so you can still switch to others for testing).

inst/private/python_ipc_pytave.m
cbm755: hmmm, not sure.  Possibly error handling in `python_cmd`.  Otherwise, its just something that either @latot or I had when debugging `python_cmd`.

cbm755:
Nice, this is looking like the right idea.

1.  It only works for a single return right?  Either we need to fix the creation of `@pyobj` for things in lists etc.  Or do some loop here (which would still not work for nested things like a list of a list).

2.  the `temp` variable thing is a bit weird for passing to `check_and_convert`.

3. Does this use `@pyobj` anywhere?

I'll think more about it later.

genuinelucifer:
On May 25, 2016 3:25 AM, "Colin Macdonald" <notifications@github.com> wrote:
>
> Nice, this is looking like the right idea.

Thanks :)

> It only works for a single return right? Either we need to fix the
creation of @pyobj for things in lists etc.

I thinks lists are working. I mean the following code works fine:

syms x y
z = [x, y]

I can access z(1) and z(2). Do you mean list of other types?

> Or do some loop here (which would still not work for nested things like a
list of a list).

Ohk . Will have to check that.

> the temp variable thing is a bit weird for passing to check_and_convert.

Hmm.

> Does this use @pyobj anywhere?

No. I mean we could have used @pyobj by returning the tuple '_d' but then
we can't do 'd(1)' as calling with params is not implemented yet (as in the
comment)

> I'll think more about it later.

Gr8! :)

cbm755:
I've tried this but it doesn't work for me.

How does `temp` get assigned to for `check_and_convert` to work?

````
error: pyexec: name 'temp' is not defined
error: called from
    check_and_convert at line 8 column 3
    python_ipc_pytave at line 75 column 5
    python_ipc_driver at line 59 column 13
    sym at line 361 column 5
    syms at line 158 column 9
    mytest at line 3 column 2
````

**Comment by cbm755** [comment]
> I thinks lists are working. 

I meant, multiple returned objects (that is, `_outs` has more than one thing).

Maybe as a start you can make it warn if `len(_outs) > 1`?

**Comment by cbm755** [comment]
Regarding the `_d` return stuff...  I think `pyeval`  can handle lists:

````
>> L = pyeval('[1, 2]')
L = 
{
  [1,1] =  1
  [1,2] =  2
}
>> L{1}
ans =  1
>> L{2}
ans =  2
````
So that could probably simplify things a bit.

**Comment by cbm755** [comment]
Suggestions for what to focus on:
1. how about a loop (in Octave) for all the entries in `_outs`, not just the first?
2. fix up the `temp` stuff.
3. something is still writing the `<f>PYTHON: successful variable import</f>` stuff (to stdout?)

**Comment by genuinelucifer** [comment]
> How does temp get assigned to for check_and_convert to work?

temp is first initialised as Matrix([_outs[0]])... Then only it is used...

> Regarding the _d return stuff... I think pyeval can handle lists:

oh thanks. Currently `_d` is tuple maybe that's why I faced error. I'll try list.

> 1. how about a loop (in Octave) for all the entries in _outs, not just the first?

Sure. That sounds good.

> 2. fix up the temp stuff.

I didn't get any error with temp as it is first assigned a value. Could you please clarify as to how do you wish it to be changed?

> 3. something is still writing the <f>PYTHON: successful variable import</f> stuff (to stdout?)

I think it's pytave. I looked a bit but couldn't pinpoint. I'll try again.

**Comment by genuinelucifer** [comment]
> I didn't get any error with temp as it is first assigned a value. Could you please clarify as to how do you wish it to be changed?

Sorry about this. I messed up the temp stuff just before submitting this PR and didn't test.

**Comment by cbm755** [comment]
re: Tuple problem: presumably a tuple should export to a cell array.  Can you file an issue for that at bitbucket.org/mtmiller/pytave?

The "Successful variable import" thing is from Symbolic: it's from `copy_vars_to_python.m` I think.

**Comment by genuinelucifer** [comment]
Except the output of successful import other 2 points have been complete. Please review.
I couldn't find the origin of that display. It is not in `copy_vars_to_python.m`.. I didn't check thoroughly though. I'll do it tomorrow.
.
I'll file the issue tomorrow. :)

File: inst/private/check_and_convert.m [hunk 7]
```diff
@@ -0,0 +1,38 @@
...
+  pyexec(str_check_is_list);
+  is_list = pyeval('is_list');
+  
+  if is_list
+    n = pyeval('len(_outs)');
+  else
+    n = 1;
+  end
+  
+  obj = [];
```
cbm755: Does this work for you?  shouldn't `obj` be a cell array?  It needs to be heterogeneous...

genuinelucifer: > > +  obj = [];
>
> Does this work for you? shouldn't obj be a cell array? It needs to be
heterogeneous...
>

Sorry. 'cell array'?
And yes it worked for me. But I tested only the conditions where either one
element was returned or a list with elements of same type were returned.

cbm755:
> It is not in `copy_vars_to_python.m`

yes it is:
````
$ grep * -r -e "successful variable import"
private/python_copy_vars_to.m:          '    octoutput_drv("PYTHON: successful variable import")' ...
private/python_ipc_popen2.m:  if (ischar(A) && strcmp(A, 'PYTHON: successful variable import'))
private/python_ipc_sysoneline.m:  if (ischar(A) && strcmp(A, 'PYTHON: successful variable import'))
private/python_ipc_system.m:  if (ischar(A) && strcmp(A, 'PYTHON: successful variable import'))
````

I think we should also look at replacing this block of code with PyTave:
````
  % load all the inputs into python
  loc = python_copy_vars_to('_ins', true, varargin{:});
  s = strjoin(loc, newl);
  pyexec(s)
````
(Maybe you can update the goals on the wiki page with that?)

File: inst/private/check_and_convert.m (hunk 7 above)
Comment (cbm755) [inline inst/private/check_and_convert.m]: cell array:
`obj = {}`.

If one output is a dict and the next is a sym, then they cannot go in an array together.

**Comment by cbm755** [comment]
I get some kind of infinite loop with horzcat when testing this.  I guess because of `obj = [obj foo]` bit, see above about cell-arrays.

**Comment by cbm755** [comment]
which branch of pytave do you test with?  hg tip?

**Comment by genuinelucifer** [comment]
I'm testing with your branch that you told about on the mailing list.

**Comment by cbm755** [comment]
funny, that's where I get infinite loops just doing `syms x`.  Oh well, get the cell array in there will probably help.

**Comment by cbm755** [comment]
Ah, wait maybe the `bookmark` is the key: I'm using `cbm_pyobj`.

**Comment by genuinelucifer** [comment]
I too got that infinite loop once. But then I changed a lot in the code and
now i don't get it. I too am using cbm_pyobj.
.
I guess u have to add 'octsymp/inst/private' to path in octave.

**Comment by cbm755** [comment]
thanks! that fixed it!

**Comment by cbm755** [comment]
Hey, this is great, a surprising amount of things just work!  So nice compared to when I was writing the IPC stuff a couple years ago.  I'm very excited about this.

File: inst/private/check_and_convert.m (hunk 7 above)
Comment (genuinelucifer) [inline inst/private/check_and_convert.m]: Oh great. I didn't know about this. Thanks. I'll add that.

**Comment by genuinelucifer** [comment]
> Hey, this is great, a surprising amount of things just work! So nice compared to when I was writing the IPC stuff a couple years ago. I'm very excited about this.

:D Great to hear that... I'm looking forward to having every piece working.. 

> thanks! that fixed it!

Well, previously too (although i never mentioned it) I got errors like python_header not defined. So, I had to add inst/private to path. But sometimes, it just worked without that. Could there be some sort of script loading error here?

**Comment by genuinelucifer** [comment]
> yes it is:
Oops. I don't know how I missed that. :P 

> I think we should also look at replacing this block of code with PyTave:
Sure

> (Maybe you can update the goals on the wiki page with that?)
Do you mean, marks goals as in progress or done? Or just edit the goal to add conversion logic to symbolic and not pytave?

**Comment by cbm755** [comment]
I usually need to do `clear all; clear classes` or restart after modifying stuff in `private`.  I've never tried to track it down too carefully.  Its worse with `@pyobj`: maybe because its classdef-style.

re: wiki: I just meant we should note somewhere that its both import and export of variables that needs updated to use PyTave.

**Comment by genuinelucifer** [comment]
Now many things work.
`rets = python_cmd('return [10, [20, 30], 40]');`
works too...
`
rets =
{
  [1,1] =  10
  [1,2] =
  {
    [1,1] =  20
    [1,2] =  30
  }
  [1,3] =  40
}
`
.
.
Only one thing. In pytave, there is code to convert dictionary into octave map. But when I do:
`ret = python_cmd('return {1:2, 3:4, 5:6}')`
All I get is a python object:
`
pyeval: could not convert return value to Octave-native object, making pyobj...
ret =
{
  [1,1] =

  [PyObject id 0x7f17371f1d70]
{1: 2, 3: 4, 5: 6}

}
`
Am I using the dictionary wrong?

File: inst/python_cmd.m [hunk 8]
```diff
@@ -121,7 +121,7 @@
 %% Author: Colin B. Macdonald
 %% Keywords: python
 
-function varargout = python_cmd(cmd, varargin)
+function varout = python_cmd(cmd, varargin)
```
genuinelucifer: I had to change this because if I do:
`python_cmd('return [10 10 10]')` Then I only get `ans=10`
same with `a = python_cmd('return [10 10 10]')` as `a=10`
.
Since now we are converting the returned objects into a cell array anyway, I thought it would be better to return the complete cell array and not just some element(s).
.
Is it okay. Or should I change it back to `varargout`?

cbm755:
not all dicts can be converted to structs: compare `pyeval('{"abc":2, "xyz":4}')`.  I think this is documented somewhere in PyTave: but if not it needs to be.

Re: varargout.  I tihnk the behaviour on master is as it should be: consider the following, and note the extra comma!
````
>> [a,b] = python_cmd('return [1,2]')
a =  1
b =  2
>> L = python_cmd('return [1,2],')
L = 
{
  [1,1] =  1
  [1,2] =  2
}
````

**Comment by cbm755** [comment]
Let me explain further.  Your example is incorrect:
````
rets = python_cmd('return [10, [20, 30], 40]');
````
That should give `rets = 10` (and a warning).  What you want to return is a single-element tuple, either of these commands:
````
rets = python_cmd('return [10, [20, 30], 40],');
rets = python_cmd('return ([10, [20, 30], 40],)');
````
I think this is documented in `help python_cmd`.

**Comment by cbm755** [comment]
I expect no changes to python_cmd, at least not while we still support the other ipc mechanisms.  

How's the test coverage?  For me, everything is broken because even simple things like `syms x` return cell arrays.  See above points, don't change python_cmd at all.

**Comment by genuinelucifer** [comment]
> I expect no changes to python_cmd, at least not while we still support the other ipc mechanisms.
>
>How's the test coverage? For me, everything is broken because even simple things like syms x return cell arrays. See above points, don't change python_cmd at all.

Sorry. I did it again. Did last minute changes before push, just to make my tests work.
I understood the usecase now. I've reverted it. Also, I sent the PR on bitbucket. please have a look.
.
I'm gonna update the wiki now.

**Comment by genuinelucifer** [comment]
The Comma thing to is not working for me! I get:
```
>> rets = python_cmd('return ([10, [20, 30], 40],)')
warning: called from
    python_cmd at line 188 column 5
rets =  10
>> rets = python_cmd('return [10, [20, 30], 40],')
warning: called from
    python_cmd at line 188 column 5
rets =  10
```

**Comment by genuinelucifer** [comment]
Also, `pyeval('_outs')` was giving error. I just casted `_outs` to string just to remove the error for now.

**Comment by genuinelucifer** [comment]
Also, on the wiki, when I went through the goals, I just found `2 a)` that I felt needed changes. I changed that. Also, should we change `3` also? (since we are actually using m-files only?)
Are there other changes to be done on wiki?

**Comment by cbm755** [comment]
>> The Comma thing to is not working for me!

delete this from check_and_convert:
````
  if n==1
    obj = obj{1};
  end
````

**Comment by cbm755** [comment]
re: wiki, I think its mostly ok to leave as is.  But maybe add a goal that "objects passed from Symbolic back to Python should use PyTave mechanisms".  It seems inappropriate to add midterm goals, so add to stretch goals?

BTW, did you ever post a blog entry with the plan?  Maybe I missed it.

**Comment by cbm755** [comment]
Things I notice:

1. unicode is not working on python2 (not too urgent and probably PyTave's problem; I filed https://bitbucket.org/mtmiller/pytave/issues/5)

2. booleans are not working properly.  Here is how master works:
````
>> r = python_cmd('return S.true')
r = (sym) True
>> r = python_cmd('return true')     % same
r = (sym) True
>> r = python_cmd('return True')
r =  1
````
but here:
````
>> r = python_cmd('return S.true');
pyeval: could not convert return value to Octave-native object, making pyobj...
>> isa(r, 'sym')
ans = 1
>> r
error: strrep: S must be a string or cell array of strings
error: called from
    disp>make_indented at line 101 column 5
    disp at line 85 column 5
    display at line 107 column 11
>> r.pickle
ans = S.true
>> r.ascii
ans = True
>> 
````

**Comment by cbm755** [comment]
I should elaborate: `True` is the built-in python boolean.  `S.true` is a SymPy thing.

**Comment by cbm755** [comment]
Hmm, the boolean thing is partially Symbolic's problem.  But its also https://bitbucket.org/mtmiller/pytave/issues/6

**Comment by cbm755** [comment]
3.  Somewhere in `python_header.py` is a "filter" function that converted Matrix of size 1 x 1 back into scalar.  Your `get_sym_from_python.m` needs to do this too.  E.g., `flipud(x)` gives `[x]` instead of `x`.

**Comment by genuinelucifer** [comment]
> re: Tuple problem: presumably a tuple should export to a cell array. Can you file an issue for that at bitbucket.org/mtmiller/pytave?

Instead of filing a bug for the tuple thing. I pushed a fix and did a PR to your repo on bitbucket.

**Comment by genuinelucifer** [comment]
After your changes too, `return True` doesn't work. Only `return S.true` and `return true` work.
I tried adding to pytave similar code as I did for tuples, but that didn't work too.

> I think we should also look at replacing this block of code with PyTave:
> % load all the inputs into python
>  loc = python_copy_vars_to('_ins', true, varargin{:});
>  s = strjoin(loc, newl);
>  pyexec(s)

Currently too the python script is run with `pyexec`.. Did you mean having a `pystore` in pytave to store variables by just giving a list of variables as input?

> Somewhere in python_header.py is a "filter" function that converted Matrix of size 1 x 1 back into scalar. Your get_sym_from_python.m needs to do this too. E.g., flipud(x) gives [x] instead of x.

Done.

> re: wiki, I think its mostly ok to leave as is. But maybe add a goal that "objects passed from 
> Symbolic back to Python should use PyTave mechanisms". It seems inappropriate to add midterm 
> goals, so add to stretch goals?

Will do.

> BTW, did you ever post a blog entry with the plan? Maybe I missed it.

Sorry, I actually never did that.


Most of the things seem to working now. (considering my PR on bitbucket)
What should be our immediate focus now? I mean should I now try to build pytave on windows? Or should we first focus on adding more functionalities to symbolic? If we are to work on symbolic, which features should be the next immediate ones to work on? (Sorry, when I can't find the just next thing to do, I tend to ask too many questions...)

File: inst/private/python_ipc_pytave.m [hunk 9]
```diff
@@ -61,25 +61,28 @@
...
-  ind = strfind(out, '<output_block>');
-  assert(length(ind) == 1)
-  A = extractblock(out(ind(1):end));
-  info.raw = out;
+  %out = pyeval('octoutput_drv(_outs, tostdout=False)')
+  %assert(ischar(out));
+  %ind = strfind(out, '<output_block>');
+  %assert(length(ind) == 1)
+  %A = extractblock(out(ind(1):end));
+  %info.raw = out;
```
cbm755: just delete all this.

inst/private/python_ipc_pytave.m
```diff
@@ -61,25 +61,28 @@
 
   % TODO: bit wasteful to repeat this every call
   headers = python_header();
-  pyexec(headers)
+  pyexec(headers);
```
cbm755: these semicolons do anything?  if not, please remove

cbm755:
> What should be our immediate focus now

In general, when in doubt, run the tests: there will be many things broken :)   

> 

> Did you mean having a pystore in pytave to store variables by just giving a list of variables as input?

Not sure what you mean.  But take a look at `python_copy_vars_to`: it has special code for converting i.e., structs to dicts, doubles to floats, integers to integers.  And of course `@sym` back to their `srepr` (stored in `x.pickle`).  All that code will need to be replaced with PyTave.  I'm not sure how hard that will be.  As in your `check_and_convert`, we might assume we have only a scalar, array or cell array, but no nested cell-arrays for now.

Anyway, let's figure out a way to do that as a different PR.

I'll make some comments to be addressed before we can merge this.

inst/private/python_copy_vars_to.m
```diff
@@ -30,7 +30,6 @@
     L = { 'try:' ...
           sprintf('    %s = []', in) ...
           L{:} ...
-          '    octoutput_drv("PYTHON: successful variable import")' ...
```
cbm755: We need a different solution here b/c this will break popen2 ipc.  Can you just pass `false` for `te`?

genuinelucifer:
@cbm755 Should I squash commits?

I am working on doing the `python_copy_vars_to` using PyTave and will submit a new PR for that.

cbm755:
I don't think we need a squash.

Here are some thoughts on priorities:

1. Can you resend https://bitbucket.org/macdonald/pytave/pull-requests/3 against mtmiller?

2. Is fixing the following similar to your tuple fix?  https://bitbucket.org/mtmiller/pytave/issues/6

3. Many tests fail because something returns a list of SymPy objects: think `return ([x, y],)`.  For example:
````
>> syms x y
>> f=x*y
f = (sym) x⋅y
>> findsymbols(f)
````
Maybe `check_and_convert` could call itself when it sees a list or tuple?  This could be another PR.

genuinelucifer:
> Can you resend https://bitbucket.org/macdonald/pytave/pull-requests/3 against mtmiller?

Done

> Is fixing the following similar to your tuple fix? https://bitbucket.org/mtmiller/pytave/issues/6

Nope. I tried. It didn't work.

> Many tests fail because something returns a list of SymPy objects

I'll look into it. I ran tests but soooo many of them failed that I couldn't start debugging any of them. :P  I'll do it today.
.

For `python_copy_vars_to.m`, we need to have one more function in pytave. Something similar to `pycall` but either more general or more specific.
I mean we have 2 options:

1. Extend (or overload or have a new function with diff name) pyexec to take many octave types as input list. And the string to run could have %0 %1 or maybe just many %% or anything like that. And then we can simply convert all octave types and store then into python objects and replace those %* with the python names in the same order.

2. We can create a new function specific to variable saving (maybe `pystore`) . So, it takes octave variables as input, converts them and then stores them into python. We can take the names of the corresponding variables as input too (or just store in `_ins[]` as it is done currently).

genuinelucifer:
> Maybe check_and_convert could call itself when it sees a list or tuple? This could be another PR.

Done (locally). Will submit in another PR. Many tests pass due to this now. :)
While some of the errors have now converted to Python Exceptions. So, instead of getting back the exact error we get '<method_name> failed'
How to deal with such test? Cause be will get errors when using other ipcs. Only pytave gives different error.

cbm755:
Ok, I will merge this.  For anyone following along, last change before I merge this on the pytave branch is a72d6b9dc3f3077b897866ec0d668d45833017b8: it might pass more tests, but this PR is closer to what we want.

cbm755:
Let's start filing bugs (and labelling them "pytave"--maybe only I have permission to do that...)

Please file one for the exception thing: but I think its low priority.

PR #460, opened 2016-05-30
Recursively call check_and_convert for lists/tuples

genuinelucifer:
Added a new parameter to `check_and_convert` to facilitate recursive calls.

### DIFF
--- a/inst/private/check_and_convert.m
+++ b/inst/private/check_and_convert.m
@@ -1,10 +1,10 @@
-function obj = check_and_convert()
+function obj = check_and_convert(var_name)
   newl = sprintf('\n');
   
-  pyexec(strjoin({'if isinstance(_outs, sp.Matrix) and _outs.shape == (1, 1):',
-                  '  _outs = _outs[0, 0];'}, newl));
+  pyexec(strjoin({strjoin({'if isinstance(', var_name, ', sp.Matrix) and ', var_name, '.shape == (1, 1):'}),
+                  strjoin({'  ', var_name, ' = ', var_name, '[0, 0];'})}, newl));
 
-  str_check_is_list = strjoin({'if isinstance(_outs, (list, tuple)):',
+  str_check_is_list = strjoin({strjoin({'if isinstance(', var_name, ', (list, tuple)):'}),
                               '  is_list = True',
                               'else:',
                               '  is_list = False'},newl);
@@ -12,7 +12,7 @@ function obj = check_and_convert()
   is_list = pyeval('is_list');
   
   if is_list
-    n = pyeval('len(_outs)');
+    n = pyeval(strjoin({'len(', var_name, ')'}));
   else
     n = 1;
   end
@@ -20,22 +20,27 @@ function obj = check_and_convert()
   obj = {};
   for i=1:n
     if is_list
-      cur_var = sprintf('_outs[%d]', i-1);
+      cur_var = sprintf('%s[%d]', var_name, i-1);
     else
-      cur_var = sprintf('_outs');
+      cur_var = sprintf(var_name);
     end
     
-    str_check_is_sym = strjoin({strjoin({'if ', cur_var, ' is None or isinstance(', cur_var, ', (sp.Basic, sp.MatrixBase)):'}),
-                                '  is_sym = True',
-                                'else:',
-                                '  is_sym = False'},newl);
-    pyexec(str_check_is_sym);
-    is_sym = pyeval('is_sym');
-    
-    if is_sym
-      obj(end+1) = {get_sym_from_python(cur_var)};
+    is_list_curvar = pyeval(strjoin({'isinstance(', cur_var, ', (list, tuple))'}));
+    if is_list_curvar
+      obj(end+1) = {check_and_convert(cur_var)};
     else
-      obj(end+1) = {pyeval(cur_var)};
+      str_check_is_sym = strjoin({strjoin({'if ', cur_var, ' is None or isinstance(', cur_var, ', (sp.Basic, sp.MatrixBase)):'}),
+                                  '  is_sym = True',
+                                  'else:',
+                                  '  is_sym = False'},newl);
+      pyexec(str_check_is_sym);
+      is_sym = pyeval('is_sym');
+
+      if is_sym
+        obj(end+1) = {get_sym_from_python(cur_var)};
+      else
+        obj(end+1) = {pyeval(cur_var)};
+      end
     end
   end
 end
--- a/inst/private/check_and_convert.m
+++ b/inst/private/check_and_convert.m
@@ -1,10 +1,10 @@
 function obj = check_and_convert(var_name)
   newl = sprintf('\n');
   
-  pyexec(strjoin({strjoin({'if isinstance(', var_name, ', sp.Matrix) and ', var_name, '.shape == (1, 1):'}),
-                  strjoin({'  ', var_name, ' = ', var_name, '[0, 0];'})}, newl));
+  pyexec(strjoin({['if isinstance(', var_name, ', sp.Matrix) and ', var_name, '.shape == (1, 1):'],
+                  ['  ', var_name, ' = ', var_name, '[0, 0];']}, newl));
 
-  str_check_is_list = strjoin({strjoin({'if isinstance(', var_name, ', (list, tuple)):'}),
+  str_check_is_list = strjoin({['if isinstance(', var_name, ', (list, tuple)):'],
                               '  is_list = True',
                               'else:',
                               '  is_list = False'},newl);
@@ -25,11 +25,11 @@ function obj = check_and_convert(var_name)
       cur_var = sprintf(var_name);
     end
     
-    is_list_curvar = pyeval(strjoin({'isinstance(', cur_var, ', (list, tuple))'}));
+    is_list_curvar = pyeval(['isinstance(', cur_var, ', (list, tuple))']);
     if is_list_curvar
-      obj(end+1) = {check_and_convert(cur_var)};
+      obj(i) = {check_and_convert(cur_var)};
     else
-      str_check_is_sym = strjoin({strjoin({'if ', cur_var, ' is None or isinstance(', cur_var, ', (sp.Basic, sp.MatrixBase)):'}),
+      str_check_is_sym = strjoin({['if ', cur_var, ' is None or isinstance(', cur_var, ', (sp.Basic, sp.MatrixBase)):'],
                                   '  is_sym = True',
                                   'else:',
                                   '  is_sym = False'},newl);
@@ -37,9 +37,9 @@ function obj = check_and_convert(var_name)
       is_sym = pyeval('is_sym');
 
       if is_sym
-        obj(end+1) = {get_sym_from_python(cur_var)};
+        obj(i) = {get_sym_from_python(cur_var)};
       else
-        obj(end+1) = {pyeval(cur_var)};
+        obj(i) = {pyeval(cur_var)};
       end
     end
   end
--- a/inst/private/get_sym_from_python.m
+++ b/inst/private/get_sym_from_python.m
@@ -1,15 +1,15 @@
 function retS = get_sym_from_python(var_name)
   newl = sprintf('\n');
-  ascii = pyeval(strjoin({'sp.pretty(', var_name,', use_unicode=False)'}));
+  ascii = pyeval(['sp.pretty(', var_name,', use_unicode=False)']);
   % workaround https://bitbucket.org/mtmiller/pytave/issues/5
   pyexec (strjoin ( { ['_d = sp.pretty(' var_name ', use_unicode=True)']
                       'if sys.version_info < (3, 0):'
                       '    _d = _d.encode("utf-8")' }, newl));
   unicode = pyeval('_d');
-  str = pyeval(strjoin({'sympy.srepr(', var_name,')'}));
-  flat = pyeval(strjoin({'str(', var_name,')'}));
+  str = pyeval(['sympy.srepr(', var_name,')']);
+  flat = pyeval(['str(', var_name,')']);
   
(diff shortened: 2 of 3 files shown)


genuinelucifer:
Well, it's 31st May here (In India) and time to go to our next goal as per the timeline.  
@cbm755 @mtmiller  Would you like to give some feedback as to how effectively we could cover the first goal of timeline:  
`30th May 	Work on adding the sym conversion to PyTave and cleaning up the conversion mechanism in Symbolic.`  
.  
Also, looking on the midterm goals, I feel we have acheived most part of 1 a) through d) and also 2 a) [I hope so].  
As 2b) and 2c) state, to increase test coverage for pytave. I saw that almost all basic functions have a test for them. Do we have to get more specific test cases? Some clarification would be good.  
Also the next phase of timeline states: `15th June        Improve tests and doctests. Work on building PyTave and testing on Windows. No more crufty XML!`  
I guess we have already gotten rid of the XML part (haven't we?)  
.  
So, should we now focus on tests? (for pytave and also symbolic). Or do we have to get some other conversion stuff cleared up before that?  
.  
(should I put this on mailing list)

cbm755:
I think some blog posts are needed (both @mtmiller and I have said at least once).  One goals/timeline and maybe another one summarizing the work so far.

And yes, I think the discussion about where to go next probably should happen on the maintainers maillist.

cbm755:
The importing of variables from Octave to Python is not "xml" but it is certainly is "crufty" :) so I guess we can thing of that as part of June 15th point on your timeline.

Does PyTave have tests?  I add doctests as I work on `@pyobject` but are there `%!` style tests anywhere?

inst/private/check_and_convert.m
```diff
@@ -1,41 +1,46 @@
...
-      obj(end+1) = {get_sym_from_python(cur_var)};
+    is_list_curvar = pyeval(strjoin({'isinstance(', cur_var, ', (list, tuple))'}));
+    if is_list_curvar
+      obj(end+1) = {check_and_convert(cur_var)};
     else
-      obj(end+1) = {pyeval(cur_var)};
+      str_check_is_sym = strjoin({strjoin({'if ', cur_var, ' is None or isinstance(', cur_var, ', (sp.Basic, sp.MatrixBase)):'}),
+                                  '  is_sym = True',
+                                  'else:',
+                                  '  is_sym = False'},newl);
```
cbm755: nested strjoin looks awkward.  Can the inner one just be `['if ' cur_var ' is None or ...']`?

inst/private/check_and_convert.m
```diff
@@ -1,41 +1,46 @@
...
+      pyexec(str_check_is_sym);
+      is_sym = pyeval('is_sym');
+
+      if is_sym
+        obj(end+1) = {get_sym_from_python(cur_var)};
+      else
+        obj(end+1) = {pyeval(cur_var)};
```
cbm755: this is not really specific to this PR, but this `end+1` is not very nice is it assumes and invariant `end + 1 == i` (which is probably true but still...)

Does `obj{i} = ` work?

inst/private/check_and_convert.m
```diff
@@ -1,41 +1,46 @@
...
+      end
     end
   end
 end
```
cbm755: Again, not specific to this PR, but see this little red thing on github?   Git doesn't like how you end files ;-)

genuinelucifer:
Not that style. But there is a file test.py which has some tests.
Do we need octave style tests?

inst/private/check_and_convert.m
cbm755: I see strjoin used in this way elsewhere too.  I think I prefer the `[]` notation.  (Your `strjoin({}, newl)` idiom is fine.)

cbm755:
This looks pretty good, see some comments above about style.

How many failing tests do we have with this?

cbm755:
> Do we need octave style tests?

I replied to this on bitbucket.  But short answer, yes.  Every change should add tests.  The actually implementation is kind of trivial compared to the tests :)

genuinelucifer:
I did the changes. But its weird that `obj{i}` messed everything up and I got many unusual errors. While `obj(i)` worked. Aren't cell arrays supposed to be accessed via `{}` and not `()`?
.
Also, if I just do `sympref ipc pytave` and then run the tests using `octsympy_tests` then only 7 tests fail! But if after the tests have been run I issue `sympref ipc` I get the answer as `default` and not `pytave`. So, I think some test changed it. So, instead I changed the `which_ipc` line again and tested with that to force pytave everytime. And now this is the summary:
```
Summary:

  PASS      1439
  FAIL       132
  XFAIL       22
  TIME      808s (683s CPU)
```
The amount of failed tests seems too much and I'll start looking into them one by one from tomorrow. Most of them should be easy fix. The tricky part might be to keep them working for all 3 IPCs at the same time.

cbm755:
`obj{i}` and `obj(i)` are not the same.  Lauren Shure says "curly for contents": worth reading her blog on this.

Yes `test sympref` changes ipc several times.  It sets it back to the default when its done.  So perhaps we should change the default ipc to pytave to on this branch... I just filed #461.

132 is not too bad!  I guess they're be some common patterns, e.g., exceptions.  Some advice: don't change tests without discussion (there certainly could be bad or inappropriate tests, but think of tests as very heavy physical objects: lots of energy to get them moving)

cbm755:
Please see a4558f27206cb034db1dd5446a5815a84214f:

* Your editor leaves some extraneous whitespace on otherwise empty lines.  If possible, please configure it to highlight such things so you don't accidentally commit it.

* Python indent is four spaces, Octave is two spaces.

* I switched it back to `obj{i}` as I find that more readable in this case.

* The lines were kind of long so I reformatted a bit.

cbm755:
Here's the blog entry I mentioned: http://blogs.mathworks.com/loren/2006/06/21/cell-arrays-and-their-contents

I first tried "Lauren curly content" but that didn't really find it ;-)

genuinelucifer:
> I switched it back to obj{i} as I find that more readable in this case.
>

If you do this then this doesn't work:
syms x
z = x^10 + 10 + 20*x;
That's why I used parentheses.

.
Thanks for other formatting and the blog. I'll go through it.

cbm755:
> That's why I used parentheses.

??? seems fine to me:

````
>> syms x
>> z = x^10 + 10 + 20*x
z = (sym)

   10
  x   + 20⋅x + 10

````

Is the pytave branch working for you?

**Comment by genuinelucifer** [comment]
> ??? seems fine to me:

Oops. I forgot to remove the `{}` from RHS.
.

> Is the pytave branch working for you?

I am using the branch on which I sent you PR on bitbucket. It's working fine.

**Comment by genuinelucifer** [comment]
> Some advice: don't change tests without discussion

Sure. Yes, there are many similar errors and we can get back to them later.

> So perhaps we should change the default ipc to pytave to on this branch... I just filed #461.

I too think we can get to building pytave on travis a bit later. but it seems somewhat of a straightforward task (since it's ubuntu). all the dependecies could be installed using `sudo apt-get` and we might have it working. Will try it later though.

**Comment by cbm755** [comment]
I'm far from the one to help debug hg problems!    But one thing I would try is to clone a fresh clone from mtmiller/pytave.  You can also send `.patch` (output of `diff -u`). 

These sorts of changes should go to `mtmiller/pytave`.  I suspect your fork on bitbucket should be of his repo.  Otherwise, you and I will take the *outer product* of our mercurial incompetence ;-)


### Re: tests:

1. does this test anything?  possible it tests the isequal method of `@pyobject`?

2. this is more like it.  Sounds like @mtmiller is considering lists staying as `@pyobject` so let's make some simpler tests for now:
````
%!test
%! q = pyeval('10.1')
%! assert (isnumeric(q))
%! % note: floating-point equality test: usually bad but here we expect the exact same float
%! assert (q, 10.1)

%!test
%! q = pyeval('"I <3 Octave"')
%! assert (ischar (q))
%! assert (strcmp (q, '1 <3 Octave'))

%!assert (islogical (pyeval ('True')))
````
(note spacing is important here: one space after %! for multiline tests).
I recommend at least one test like that for each thing we can convert to Octave.  Keyword here is "coverage".

3 -- As above.  But this could be useful, so I suggest:
````
%!test
%! % This might change if we stop converting lists
%! z = pyeval ...
````
(so we recognize later that this test has low "momentum")

4 -- Ok, looks reasonable.  Can you do same with a matrix [1 2 3; 4 5 6], to ensure we keep sizes properly.

### other thoughts

Re: pyeval('10')  what does this do?  Does it give double or some kind of int such `int64`?  Let's have a test that makes it clear---such as `assert (! isinteger (q))`.  BTW, do we have docs that make that clear?   If not, let's document that.

I think contributions to PyTave probably should follow Octave's weird `fcn (4)` and `array(4)` convention.

**Comment by cbm755** [comment]
`pyexec` needs tests too.

**Comment by genuinelucifer** [comment]
> These sorts of changes should go to mtmiller/pytave. I suspect your fork on bitbucket should be of his repo. Otherwise, you and I will take the outer product of our mercurial incompetence ;-)

Well, his repo still doesn't have `@pyobject` or `@pyobj`. So, it would be better if first you submit that PR to him. Then I can add upon it with tests.
Or I can submit the tests in some other file like `test/tests.m`!?

> does this test anything? possible it tests the isequal method of @pyobject?

Umm... Just verifying that what we get from pyeval is preserved.. I mean consistent however many times you call it.

> pyeval('10') what does this do? Does it give double or some kind of int such int64? Let's have a test that makes it clear---such as assert (! isinteger (q)). BTW, do we have docs that make that clear? If not, let's document that.

I don't think we have any document as such. I'll make that.

> I think contributions to PyTave probably should follow Octave's weird fcn (4) and array(4) convention.

I'll search for those and try to follow them.

> pyexec needs tests too.

Well, in `pycall` test, `pyexec` is called. Separately too, I'll add a couple of tests for it.

.
Thanks for the other pointers, I'll do the edits... :)

**Comment by cbm755** [comment]
> Well, his repo still doesn't have |@pyobject| or |@pyobj|. So, it would
> be better if first you submit that PR to him. Then I can add upon it
> with tests.
> Or I can submit the tests in some other file like |test/tests.m|!?

For the tests of `pyexec` and `pyeval`, they should go in the c++ code
(IIRC, see Octave source).

I don't think we should wait for @pyobject to be merged.

(The tests don't all have to be in one PR.)

I mainly want to get in the habit of building tests as we go along: its
frustrating to add them later.

Colin

**Comment by genuinelucifer** [comment]
> For the tests of `pyexec` and `pyeval`, they should go in the c++ code (IIRC, see Octave source).

I saw the octave source. The tests seem to be in `test/*.tst` files. They seem to follow octave syntax.
Should I follow the convention in octave tests which first define a function and then tests it. One of the examples:
```
## one input with two possible inputs
%!function f (x, y)
%!  assert (nargin, 1);
%!  assert (nargout, 0);
%!endfunction
%!test
%! f (1);
```

and did you mean the same by this:

> I think contributions to PyTave probably should follow Octave's weird fcn (4) and array(4) convention.

cbm755:
Oh sorry for confusing things: nothing so complicated as that: I just meant putting a space between function and arguments as in `my_fcn (4)` (instead of `my_fcn(4)`).

cbm755:
> The tests seem to be in test/*.tst files.

They can also go comments with the c++ code.  But maybe I prefer `.tst` file because you can run the tests without having the `.cc` files around.  Let's do that for now, can always more them later.

Any progress fighting with hg to do a PR?

genuinelucifer:
> Any progress fighting with hg to do a PR?

Well, not quite. I went to mercurial irc yesterday and got some of the doubts cleared.
But still now I face one problem. I clone my `pytave_main_fork` from bitbucket (it is fork of mtmiller/pytave)..
I can see the 2 commits to fix tuple and boolean in my hg log but i cannot see the changes in the cc files :p
I'll try again today..

genuinelucifer:
Using fork of mtmiller/pytave, many of the tests don't work.
The sym conversion is not working. Logical and tuple things are not working.
.
I tried to base my work on the PR that I submitted on bitbucket but
couldn't do. It got even more messed up. Now hg push is not working. I'm
really finding hg a lot more dificult than git. When I started with git, it
naturally became more and more easy. It's completely opposite with hg.
I'll try with a fresh start tomorrow.

cbm755:
We do not expect symbolic to work on master pytave yet!

Sorry to hear that re hg.

cbm755:
In case it help, the workflow we probably want is:
1. lots of fairly small PRs.  We should not try to have one PR that fixes *everything* in PyTave.
2. each self-contained.
3. each including appropriate tests.
4. generally each PR will not get merged as is, or not merged for many weeks (well maybe we try to be better than this but just speaking from experience submitting fixes to other projects).
5. a consequence is you'll need to find a workflow were you have many bookmarks on the go at any one time.  When one get's some review, make those changes.  Then switch bookmarks and work on something else.

That's not really specific to mercurial but I thought you might find a list of meta-goals like that helpful while you get your hg skills organized :)

Finally, maybe a expert (@mtmiller) would scoff but I think there is no harm in having a couple independent directories, cloning the same project.  Especially while learning!

mtmiller:
No I definitely agree having multiple directories helps when working on separate "topics" at the same time. Especially for beginners. And especially for a large project like Octave itself. I keep two directories for working on Octave's `default` and `stable` branches, plus more if I want to work on a bigger change for an extended period of time.

genuinelucifer:
I finally got my hg repo organised. Submitted 3 PRs [9](https://bitbucket.org/mtmiller/pytave/pull-requests/9), [10](https://bitbucket.org/mtmiller/pytave/pull-requests/10), [11](https://bitbucket.org/mtmiller/pytave/pull-requests/11).
We have the basic tests covered for now. Should I put some more tests? I am going to start trying to build pytave on windows. But, if you feel like we need more tests, just let me know and I'll add more tests.

cbm755:
Looks good for now.  Good to get into the habit of always adding tests with each change.

The other project is the conversion from Symbolic to Python: I filed #464.

PR #463, opened 2016-06-03
Ezplot fixes and make tests quieter

cbm755:


cbm755:
A good habit for a project is that all PRs are reviewed: often we don't have enough active developers... but this summer we do!

@genuinelucifer can you review this?

inst/@sym/ezplot.m
```diff
@@ -126,58 +135,66 @@
 end
 
 
+%!shared hf
+%! hf = figure ('visible', 'off');
```
mtmiller: This will probably be perfectly fine for some users, and cause catastrophic crashes for others (any Windows user, Nvidia users on Linux).

cbm755: Ouch.  That's probably not worth it just so *I* don't see popup windows when I run local tests.

I'll revise and leave a pointer to https://savannah.gnu.org/bugs/?44478

PR #468, opened 2016-06-08
Pytave fixes

cbm755:
@genuinelucifer can you review this please?

### DIFF
--- a/inst/private/python_ipc_pytave.m
+++ b/inst/private/python_ipc_pytave.m
@@ -70,7 +70,6 @@ function [A, info] = python_ipc_pytave(what, cmd, varargin)
 
   s = strjoin(cmd, newl);
   pyexec(s)
-  
-  info.raw = pyeval('str(_outs)');
+
   A = check_and_convert('_outs');
 end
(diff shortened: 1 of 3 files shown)


inst/private/python_ipc_pytave.m
```diff
@@ -70,7 +70,6 @@
 
   s = strjoin(cmd, newl);
   pyexec(s)
-  
-  info.raw = pyeval('str(_outs)');
```
genuinelucifer: If we are removing info. We should remove `info.prelines` too (you probably missed it)

cbm755: On 08/06/16 22:28, Abhinav Tripathi wrote:
> If we are removing info. We should remove |info.prelines| too (you
> probably missed it)

No I think its needed for error messages.  The error messages might be
wrong, but for now we need it.

genuinelucifer: Oh ok. Actually I still don't know how this is used. I probably haven't
looked much into it.

cbm755: > Actually I still don't know how this is used.

Its just a hack that @latot came up with: reports the line number an 
exception occurred within the `cmd` block of `python_cmd`.  But 
different IPC mechanisms added various amounts of stuff to `cmd`. 
`prelines` keeps track of how much stuff.  Its probably zero for PyTave ipc.

Rest of this look ok to merge?

genuinelucifer: Hmm. Sure :)

PR #470, opened 2016-06-09
Round trip tests, `@double` fcns

cbm755:
Add some `@double` functions which call SymPy.  This allows us to get `sinint`, `cosint` without using the broken versions from `specfun` pkg for example.

All add "round-trip" tests: these improve `function_handle` for many special functions, and allow plotting with `ezplot` (for example).

The `@double` fcns are not particularly fast: but better than nothing.  They could be replaced with native-implementations at a later date (left notes in their help text to that effect).


cbm755:
This depends on upstream https://github.com/sympy/sympy/pull/11219

cbm755:
this is merged to master instead.

PR #477, opened 2016-06-16
[maint] Add PyTave to CI matrix

cbm755:
This doesn't actually work yet, but first steps to get it downloaded...
Idea is that Symbolic will automatically use PyTave if available.

- [ ] the test should fail if pytave build fails

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,11 @@ language: python
 matrix:
   include:
     - python: "2.7"
-      env: SYMPY_VER=1.0 OCT_PPA=yes PYTAVE=yes DOCTEST=yes COLUMNS=80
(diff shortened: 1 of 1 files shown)


cbm755:
So far this stops with:
````
checking for python... /home/travis/virtualenv/python2.7.10/bin/python

checking for numpy includedir... /home/travis/virtualenv/python2.7.10/lib/python2.7/site-packages/numpy/core/include

checking whether linking to numpy library works... yes

configure: error: cannot find install-sh, install.sh, or shtool in "." "./.." "./../.."

make: *** No targets specified and no makefile found.  Stop.
````
(interestingly, it does not fail the test because of that: failing to build pytave should fail the test!)

I don't have a lot more time for this: its here if @genuinelucifer or others want to try to get it working.

PR #482, opened 2016-06-21
[maint] CI needs some GPG key when using updated Octave PPA

cbm755:
I've not looked into exactly what has changed and why but I assume
its related to Octave 4.0.2 going into the "octave/stable" PPA.
(Blindly copying commands about GPG off the internet: you'd think I'd
know better!)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ matrix:
 # need octave devel pkgs for doctest (has compiled code as of July 2015)
 install:
   - if [ "x$OCT_PPA" = "xyes" ]; then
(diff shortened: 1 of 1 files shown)


cbm755:
... or maybe we don't need this afterall (temporary problem with travis?)

Will leave for a few days in case it comes back, then close this...

genuinelucifer:
I just triggered build on `ci_pytave` branch and this key was not needed... Probably it was a temporary problem with travis.

cbm755:
Ok, I'll close this.

PR #484, opened 2016-06-21
sort assumptions list alphabetically

cbm755:


cbm755:
@genuinelucifer please review this minor fix.

(basic issue here is dict order can and does depend on python version).

inst/assumptions.m
```diff
@@ -112,7 +112,7 @@
       'if d == {}:'
       '    astr = ""'
       'elif all(d.values()):'  % all True so list them
-      '    astr = ", ".join([str(i) for i in d.keys()])'
+      '    astr = ", ".join(sorted([str(i) for i in d.keys()]))'
```
genuinelucifer: I too looked some places and it seems there's no standard way to do it in python. So, this change looks good.. :) :+1:

inst/@sym/assumeAlso.m
```diff
@@ -30,15 +30,7 @@
 %% @group
 %% syms x integer
 %% x1 = x;
-%% assumptions(x1)
-%%   @result{} ans =
-%%     @{
-%%       [1,1] = x: integer
-%%     @}
```
genuinelucifer: Why to remove this doctest? Removing the `+SKIP` is ok but why this test?
Am I missing something!?

cbm755: I guess its because I only added that test as a workaround because of the skipped test.

But I guess it does read ok with it, I'll put it back!

cbm755:
Thanks, updated and merging

PR #490, opened 2016-06-23
xtest 2 tests which currently cause problems in pytave IPC

genuinelucifer:


inst/python_cmd.m
```diff
@@ -426,7 +427,8 @@
 %! a = python_cmd('return _ins[0]*2', 3);
 %! assert (isequal (a, 6))
 
-%!error <octoutput does not know how to export type>
+%!xtest error <octoutput does not know how to export type>
+%! % This test will pass with pytave IPC when we have @pyobject incorporated
```
cbm755: Comment is not correct.
````
%! % This command does not fail with PyTave and `@pyobject`.
````

genuinelucifer:
Changed..

PR #491, opened 2016-06-23
Converts dicts with sym keys to structs...

genuinelucifer:
Required by `@sym/solve.m`

### DIFF
--- a/inst/private/check_and_convert.m
+++ b/inst/private/check_and_convert.m
@@ -29,10 +29,22 @@ function obj = check_and_convert(var_name)
         'if isinstance(temp, sp.Matrix) and temp.shape == (1, 1):',
        ['    ' cur_var ' = temp[0, 0]']}, newl));
 
-
+    is_dict_with_sym_keys_curvar = pyeval(['isinstance(', cur_var, ', dict) and len(' cur_var '.keys())>0 and isinstance(' cur_var, '.keys()[0], (sp.Basic, sp.MatrixBase))']);
     is_list_curvar = pyeval(['isinstance(', cur_var, ', (list, tuple))']);
+
     if is_list_curvar
       obj{i} = check_and_convert(cur_var);
+    elseif is_dict_with_sym_keys_curvar
+      %if cur_var is dictionary with symbols as keys then convert it to a struct
+      pyexec(strjoin({'_allKeysStr = []',
+                      '_allValues = []',
+                      ['for key, value in ' cur_var '.iteritems():'],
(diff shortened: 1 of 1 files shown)


cbm755:
I'll need to think about this a bit.  Is this roughly the same as `elif isinstance(x, dict):` part of `python_header.py`?

genuinelucifer:
Not exactly... But yes, somewhat similar...
.
It just takes care of the special sub-case where the keys are sym objects.. Else it leaves the conversion to PyTave

inst/private/check_and_convert.m
```diff
@@ -29,10 +29,22 @@
...
     if is_list_curvar
       obj{i} = check_and_convert(cur_var);
+    elseif is_dict_with_sym_keys_curvar
+      %if cur_var is dictionary with symbols as keys then convert it to a struct
+      pyexec(strjoin({'_allKeysStr = []',
+                      '_allValues = []',
+                      ['for key, value in ' cur_var '.iteritems():'],
+                      '    _allKeysStr.append(str(key))',
+                      '    _allValues.append(value)'}, newl));
+      _allKeysStr = pyeval('_allKeysStr');
```
genuinelucifer: I seeing all these weird looking variables on github (when the variable name is preceeded by '_' (undersscore)..
Is there a convention not to use variable names starting with underscore in octave?

cbm755: Not allowed in Matlab.  I think github does not know about Octave.  That's sometimes a problem for octsympy, if Matlab even parses one of our files but I don't think it will get to this one, so don't worry about it.

genuinelucifer: Oh. Thanks. :)

inst/private/check_and_convert.m
```diff
@@ -29,10 +29,22 @@
...
+      for j = 1:numel(_allKeysStr)
+        obj{i}.(_allKeysStr{j}) = get_sym_from_python(sprintf('_allValues[%d]', j-1));
```
cbm755: This solution is a bit fragile: but let's leave it for another fix, I filed #492.

cbm755: I think I prefer `setfield` but this is ok too (no need to change).

But I do think you want `obj{i} = struct()` before the for loop though.

cbm755:
Ok, I did the struct initialization thing and merged

genuinelucifer:
Thanks. I was actually changing a few things and running all the tests to try if something breaks.. So couldn't do this...

PR #495, opened 2016-06-23
Use SymPy's ODE IVP/BVP solver

cbm755:
Weeee, we can drop our custom code!

@maprieto are you still interested?  good things are happening upstream in https://github.com/sympy/sympy/pull/11264

TODO list:
- [x] remove temporary xtest from underdetermined ICs example
- [x] ~~~push some tests upstream too~~~  might be nice, but not essential to hold this up.


maprieto:
Great news! I've been offline for a while... but I'll try to check the new implementation for dsolve

cbm755:
This was merged upstream so I've set dsolve to use the upstream solution for sympy > 1.1.1.  But our original implementation will be used for a while yet, to be removed when we depend on sympy > 1.1.1.

cbm755:
@mtmiller I used similar logic to one of your tests instead of using an xtest.  Look ok?

PR #496, opened 2016-06-24
Add assumption number to see which test fails...

genuinelucifer:
trying to build on travis to see the result

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ install:
         sudo apt-get install -qq -y libboost-python-dev python-numpy;
         hg clone https://bitbucket.org/genuinelucifer/pytave_main;
         cd pytave_main;
+        hg checkout travis;
         pwd;
         autoreconf --install;
         ./configure;
--- a/inst/private/store_vars_in_python.m
+++ b/inst/private/store_vars_in_python.m
@@ -31,13 +31,7 @@ function store_vars_in_python (varname, L)
       pyexec([varname '.append(' tempname ')'])
     else
       pycall ('pystoretemp', x)
-      if (isscalar (x) && isinteger (x))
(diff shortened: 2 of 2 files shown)


cbm755:
My suspicion from yesterday's tests is that it fails the first time pycall takes an argument... just in case that helps.

genuinelucifer:
I added some pycall tests.. But they seem to pass :(
Well, I'll try some more tests...

genuinelucifer:
So, for sym pycall throws exception and then terminates. While for anything else except strings it is directly causing segfault!
Any suggestions? @mtmiller

cbm755:
We're on Octave 4.0.2, with a non-master branch of PyTave... I suspect @mtmiller will tell us to try to octave unstable PPA.

Perhaps you can try to track down the exception bug in pyexec et al:
https://bitbucket.org/mtmiller/pytave/issues/24/pyexec-inconsistent-error-handling

genuinelucifer:
ohk. I'll try that then. :)

cbm755:
Also this one:
https://bitbucket.org/mtmiller/pytave/issues/16/octave-to-pythoncc-gives-cryptic-error-msg

cbm755:
In fact isn't 16 likely what we see on travis?  Anyway, the point is there are many PyTave bugs that make this harder to debug: let's spend some time fixing PyTave.

genuinelucifer:
Sure. I guess most of the work on `pytave_ex` is done anyway. We can definitely now focus on PyTave bugs..

cbm755:
Yes, or at least this first draft of `pytave_ex` is done :)

genuinelucifer:
I did some changes in pytave_ex. Trying to bring those here caused this wreck and this has conflicts. I will close this PR, test travis on the pytave_ex branch directly and when travis starts working I'll squash the travis commits on pytave_ex...

PR #502, opened 2016-07-17
Pytave_ex travis test

genuinelucifer:
Trying to see why some pytave tests are causing segfault on pytave_ex


genuinelucifer:
So, Finally it has come down to the point that on trying to store a struct into python is causing segfault.  
.  
Trying same things on same bookmark/branch of pytave and symbolic is not giving any error for me on Ubuntu 16.04  
.  
I'll try to add some logs in my pytave repo and rebuild here tomorrow to see exactly what is causing the segfault...

cbm755:
Can you produce the struct thing with a minimal nonworking example (i.e., no symbolic) and submit a bug to Pytave?  Errors are ok but obviously we don't want it segfaulting!

cbm755:
This MNWE seems fine to me:
````
octave:8> pycall ('print', struct('a', 1, 'b', 'hello', 'c', [1,2,3]))
error: pycall: error in argument type conversion
````
Are you still trying to get pytave running on Octave 4.0.3?

genuinelucifer:
> Are you still trying to get pytave running on Octave 4.0.3?

Sorry, I forgot about octave version on travis. I think it's 4.0.2 It doesn't matter if it's greater than 4.0.0 right?

And yes I did reproduce a minimal example which I ran on this PR. But since I saw a 'green tick' on a commit I hastily removed and overwrote this branch... I now feel stupid... -_-

I will reproduce this and file the issue...

cbm755:
Whatever, 4.0.x.  I mean as opposed to dev version (which is what pytave supports).

I checked and `ppa:dac922/octave-unstable` unfortunately doesn't support trusty, so we can't use it here.  I used the launchpad form to email the author and ask about backporting it,

cbm755:
dac922 said:
````
I gave up building octave regularly. Even my current PPA build for
xenial is pretty outdated.
````
So if  pytave "just works" on 4.0.x then maybe this is worth doing, but as we've said a bunch of times, very low priority.

genuinelucifer:
Sure... We can start merging stuff after struct-to-dict is fixed upstream...

genuinelucifer:
Sure... We can start merging stuff after struct-to-dict is fixed upstream...

genuinelucifer:
Closing this as travis can't support pytave (yet)!

PR #508, opened 2016-08-13
[wip] Proper check and convert

genuinelucifer:
This is based upon `pytave_ex`. There were many required changes in pytave_ex and hence I based it upon that. The last 2 commits belong to this PR.  
  
The last 2nd commit is attempt to rewrite the entirety of `check_and_convert` and `get_sym_from_python`... Although it is working correctly now (all the tests pass), it seems overly SLOOOW on my system... I was focussed on making it work (as I had to do many minor edits for every test that failed) till now. Now, I will look into optimising it...  
  
So, this is just for suggestions now, NOT for merging...


genuinelucifer:
By mistake I opened this PR to `master` and not to `pytave` and hence so many commits here. It seems github currently doesn't support changing target branch for a PR. I will open a new PR when this will more mature (don't want to open too many PRs unnecessarily)..

Please comment on the commits (last 2) directly for any required changes...  
  
Also I forgot to mention that this relies on `isequal` change in pytave which is currently under review upstream..

inst/private/get_sym_from_python.m
```diff
@@ -1,25 +1,22 @@
...
-  pyexec (strjoin ( { ['_d = sp.pretty(' var_name ', use_unicode=True)']
-                      'if sys.version_info < (3, 0):'
-                      '    _d = _d.encode("utf-8")' }, newl));
-  unicode = pyeval('_d');
-  str = pyeval(['sympy.srepr(', var_name,')']);
-  flat = pyeval(['str(', var_name,')']);
+function retS = get_sym_from_python(var_pyobj, sp)
+  pyexec('def sp_pretty_proxy(s, u): return sp.pretty(s, use_unicode=u)');
+  ascii = pycall('sp_pretty_proxy', var_pyobj, false);
+  unicode = pycall('sp_pretty_proxy', var_pyobj, true);
```
cbm755: add todo comment here for upstream kwargs issue.

inst/private/get_sym_from_python.m
```diff
@@ -1,25 +1,22 @@
...
-                      '  _d = list(temp.shape)' ,
-                      'elif isinstance(temp, sp.MatrixExpr):' ,
-                      '  _d = [float(r) if (isinstance(r, sp.Basic) and r.is_Integer)',
-                      '        else float("nan") if isinstance(r, sp.Basic)',
-                      '        else r for r in temp.shape]',
-                      'else:' ,
-                      '  _d = [1, 1]'}, newl);
-  pyexec(str_eval);
-  _d = pyeval('_d');
+  str = sp.srepr(var_pyobj);
```
cbm755: maybe `srepr` is better than `str` here?

cbm755: what is the advantage of passing `sp` into this function versus `py.sympy.srepr(...)`?  I guess maybe its faster?  But looks a bit like premature optimization to me.

inst/private/get_sym_from_python.m
```diff
@@ -1,25 +1,22 @@
...
+  flat = py.str(var_pyobj);
+
+  if py.isinstance(var_pyobj, py.tuple({sp.Matrix, sp.ImmutableMatrix}))
+    _d = py.list(var_pyobj.shape);
+  elseif py.isinstance(var_pyobj, sp.MatrixExpr)
+    shape_func = pyeval(strjoin({'lambda x : [float(r)',
+                                 'if (isinstance(r, sp.Basic) and r.is_Integer)',
+                                 'else float("nan") if isinstance(r, sp.Basic)',
+                                 'else r for r in x.shape]'}, ' '));
```
cbm755: re: passing in `sp`: for example, caching this lambda might save more time (don't bother doing that yet, its just for example).

inst/private/check_and_convert.m
```diff
@@ -1,63 +1,59 @@
...
-      %if cur_var is dictionary with symbols as keys then convert it to a struct
-      pyexec(strjoin({'_allKeysStr = []',
-                      '_allValues = []',
-                      ['for key, value in ' cur_var '.iteritems():'],
-                      '    _allKeysStr.append(str(key))',
-                      '    _allValues.append(value)'}, newl));
-      _allKeysStr = pyeval('_allKeysStr');
+      obj{i} = check_and_convert(cur_pyobj);
+    elseif dict_to_struct
+      %if cur_var is dictionary with symbols/strings as keys then convert it to a struct
```
cbm755: comment here with upstream issue URL and comment out code `obj{i} = struct(cur_pyobj)`

inst/private/check_and_convert.m
```diff
@@ -1,63 +1,59 @@
...
     if is_list
-      cur_var = sprintf('%s[%d]', var_name, i-1);
+      cur_pyobj = var_pyobj{i};
     else
-      cur_var = var_name;
+      cur_pyobj = var_pyobj;
+    end
+
+    if(py.isinstance(cur_pyobj, sp.Matrix) && isequal(cur_pyobj.shape, py.tuple({1, 1})))
+      cur_pyobj = cur_pyobj.__getitem__(py.tuple({int8(0),int8(0)}));
```
cbm755: upstream issue URL in comment please

inst/private/check_and_convert.m
```diff
@@ -1,63 +1,59 @@
...
     end
 
-    pyexec(strjoin({ ...
-       ['temp = ' cur_var],
-        'if isinstance(temp, sp.Matrix) and temp.shape == (1, 1):',
-       ['    ' cur_var ' = temp[0, 0]']}, newl));
+    if py.isinstance(cur_pyobj, builtins.dict)
+      cur_keys = cur_pyobj.keys();
+      dict_to_struct = true;
```
cbm755: Can you leave a brief comment in the code for why we only do this for some dicts (those with Symbols for keys?)  I've forgotten

cbm755:
Looks nicer already!  See comments above.  

> I forgot to mention that this relies on isequal change in pytave which is currently under review upstream..

Can you add a comment in the code where this is used?

inst/private/check_and_convert.m
```diff
@@ -1,63 +1,59 @@
...
-     ['        ' var_name ' = list(' var_name ')'],
-      'else:',
-      '    is_list = False'}, newl));
-  is_list = pyeval('is_list');
+function obj = check_and_convert(var_pyobj)
+  persistent builtins
+  persistent sp
+  if isempty(builtins)
+    builtins = pyeval("__builtins__");
+    sp = pyeval("sympy");
```
cbm755: `sp = py.sympy`?

genuinelucifer:
The `isequal` change is finally merged into pytave and hence this currently builds fines with `@` bookmark of pytave...  

Even after caching the lambda and many more constants (like `py.tuple({1,1})`) too, the performance improvement is not much...  
  
The earlier `pyexec`->`pyeval` approach was on par with `popen2` IPC (I did test it atleast 10-20 times on different files and popen2 was only slightly faster)... But this new approach to `py.*` has far more `pycall`s and hence is a lot slower (5-50 times slower than popen2!!!)...  
  
I think the only way to make it faster is to reduce the `pycall`s (direct/indirect)... And thinking again, our previous approach was better as it used lesser calls to pytave... Are we sure, we need to use the `py.*` notion?

genuinelucifer:
I rewrote `store_vars_to_python` also and added to this PR only as I thought it would be better together with the changes in `check_and_convert`... Also, `outs` returned was better handled by the new format for `check_and_convert`.

inst/sympref.m
```diff
@@ -369,14 +369,14 @@
...
 %% popen2 because test suite should be portable.
-%!error <line 1> python_cmd('raise ValueError');
-%!error <line 1> python_cmd('raise ValueError', sym('x'));
-%!error <line 1> python_cmd('raise ValueError', sym([1 2 3; 4 5 6]));
-%!error <line 1> python_cmd('raise ValueError', {1; 1; 1});
-%!error <line 1> python_cmd('raise ValueError', struct('a', 1, 'b', 'word'));
-%!error <line 2> python_cmd( {'x = 1' 'raise ValueError'} );
-%!error <line 3> python_cmd( {'x = 1' 'pass' '1/0'} );
-%!error <line 3> python_cmd( {'a=1' 'b=1' 'raise ValueError' 'c=1' 'd=1'} );
+%!error <ValueError> python_cmd('raise ValueError');
```
cbm755: these tests are specifically designed to check that the line numbers are correct.  I won't merge this.

I suppose we could drop the line number feature...

cbm755:
We should probably fix the slowness rather than working around it.  But maybe ok for now.  Can you rebase this against my `pytave` branch?  You can force push.

While you're at it, please drop this commit: 8264115.

cbm755:
Re: slowness, I noticed Octave master is much slower than than 4.0.x for traditional clsases.

Possible its this: https://savannah.gnu.org/bugs/?48785

genuinelucifer:
> I noticed Octave master is much slower than than 4.0.x for traditional clsases.

Oh. great. Probably better to be fixed upstream then..

genuinelucifer:
closing this as will open a new PR against the right branch...

PR #509, opened 2016-08-14
cast sym to pyobject

cbm755:
This is for use with the Pytave project.


genuinelucifer:
Should this be used in `store_vars_in_python`?
.
I see that you have mentioned some errors using this approach. I've currently used `pyeval(char(x))` which you have also mentioned in the comments...

cbm755:
Eventually yes, but maybe not yet.  `pyeval(char(x))` will work better for now.

cbm755:
hmmm, something closed this, I see because it isn't against master...

PR #515, opened 2016-08-24
Add method to convert a sym to a pyobject

cbm755:
This is for use with the Pytave project.

### DIFF
--- a/inst/@sym/pyobject.m
+++ b/inst/@sym/pyobject.m
@@ -1,0 +1,54 @@
+%% Copyright (C) 2016 Colin B. Macdonald
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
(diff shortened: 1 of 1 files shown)


cbm755:
This was #509, reopened after pytave merge

inst/@sym/pyobject.m
```diff
@@ -0,0 +1,54 @@
...
+%%
+%% @end group
+%% @end example
+%% @seealso{@@sym/char, py, @@pyobject/pyobject}
+%% @end defmethod
+
+
+function y = pyobject (x)
+
+  y = py.sympy.S (char (x));
```
cbm755: this will need to be `sympy (x)` not `char (x)` b/c of recent changes.

mtmiller:
My only nits to pick are the names. The name `pyobject` isn't Matlab compatible, it's what we came up with to be the hidden implementation of PyObjects, which is fine, but it's kind of like an implementation detail at this point.

I also prefer `have_pytave` to be something else since the name pytave will be obsolete soon.

cbm755:
I changed to `have_py`.  I agree with the other point and have moved the discussion upstream: https://bitbucket.org/mtmiller/pytave/issues/61/have-subsasgn-pycall-others-cast-their

mtmiller:
Thanks for the reminder about the upstream issue, I had been wondering what the use case would be, now I see that you want to be able to assign a sym variable into a container, and not that the user is calling the conversion method explicitly. So the intent is that this effectively acts as an implicit casting operator.

PR #516, opened 2016-08-30
fix python_cmd - return without ,

latot:
Hi all, well i found this little problem, when we send some return to can set the varargout the A should be in the form:

```
A = 
{  
    [1,1] =
    {
        (sym) 1
.....
    }
}
```

but when we send a return without "," like 'return A' and A its a list octave interpret it as
```
A =
{
    (sym) 1
    ....
}
```

causing varargin only takes the first value of the list.

Cya.


latot:
i do somthing wrong let me fix D:

latot:
mm, i'll close this for now and open an issue

Issue #519, opened 2016-08-31
problem symvar/findsymbols don't detect symfun


latot:
Hi again, well i found this trying to edit the diff function, i think its more simple with a example:

```
>> syms t f(g) g(t)
>> symvar(f)
ans = (sym) g
>> symvar(symvar(f))
ans = (sym) g
```

as you can see this is weird, the first part its fine the symbol of f is g, but the symbol of g is t and we get g again, as a note, symvar return a list of symbols but it don't exactly match with the input, its correct say the vars of f is g, but in the text g is printed as a symbol and should be a symfun, theoretically, this should be like this:

```
>> syms t f(g) g(t)
>> symvar(f)
ans = (symfun) g
>> symvar(symvar(f))
ans = (sym) t
```

a little trick is do this (i don't like this):

```
>> syms t f(g) g(t)
>> eval(disp(symvar(f)))
ans = (symfun) g
>> eval(disp(symvar(eval(disp(symvar(f))))))
ans = (sym) t
```

And this don't works inside a functions D:
the idea in this its can apply correctly the diff to functions, for that reason this is necessary.

Thx. Cya.

cbm755:
Well, here's what SMT does:
````
>> syms t f(g) g(t)
>> t
 
t =
 
t
 
>> f
 
f(g) =
 
f(g)
 
>> g
 
g(t) =
 
g(t)
 
>> symvar(f)
 
ans =
 
g
 
>> 
````

`syms` goes left to right.  It first makes `f(g)` then it makes `g(t)` which overwrites the sym `g`.  It think messing around with previous symfun's would be too magical and probably annoy someone.

PR #521, opened 2016-09-01
rewrite diff

latot:
Hi!, it feels to much time, well here i have a rewrite to the diff function, the reasons are simple, first, when we have a function with more than one var the diff command send the partial derivation only of the first founded var, ex:

```
>> syms f(g, h)
>> diff(f)
ans(g, h) = (symfun)

  ∂          
  ──(f(g, h))
  ∂h         
```
i think this is really bad, theoretically the diff should use the chain rule and apply it, but actually we can't do that, at least while this issue is open: https://github.com/cbm755/octsympy/issues/519
So, i think its better have a warning/error instead a wrong calc, (and start searching the problem etc).

I try keep the main functions of diff, so the new syntax would be :

```
diff(FUNCTION, VAR1, GRADE OF VAR1, VAR2, GRADE OF VAR2 .....)
```

this code recognize symbols on functions if you don't send the first derive var, and if the first arg is a number is derived in that grade, one of the differences with the actual diff is this:

actual diff:
```
>> syms x
>> f=sin(x)
f = (sym) sin(x)
>> diff(f, 2, x)
ans = (sym) -sin(x)
```

new diff
```
>> syms x
>> f=sin(x)
f = (sym) sin(x)
>> diff(f, 2, x)
ans = (sym) -cos(x)
```

basically the new diff(f, 2, x) menas, derive f two times in x, then derive it in x again.
I think include the chain rule but i will leave that to sympy: https://github.com/sympy/sympy/issues/11573

Thx. Cya.


latot:
mm, reading this will break compatibility with SMT, so this don't is an option D:
other pr!

latot:
Thinking, the diff of SMT diverge of the diff of SymPy, maybe can be good idea move this diff to diff2 to better compatibility with SymPy

latot:
mm, this seems a octave bug:
```
>> test @sym/diff2.m
***** test
 % octave's vector difference still works
 assert(isempty(diff2(1)))
 assert((diff2([2 6]) == 4))
!!!!! test failed
'diff2' undefined near line 4 column 17
shared variables 

x = (sym) x
y = (sym) y
z = (sym) z
>> syms x
>> diff2(x)
ans = (sym) 1
```

latot:
ups, don't is a octave problem, my bad again, i notice this now, the actual diff don't its compatible with octave D:
that should be fixed!
lets work a little more in this.

cbm755:
Regarding your specific example: that does look wrong, I think symvar should choose `g` as the variable to differentiate w.r.t.; I've filed #550.

cbm755:
Now: I don't understand what you're trying to fix.  `diff(f, 2, x)` is old legacy SMT stuff as discussed elsewhere.  But it is supposed to calculate the 2nd partial wrt x: not a third partial.

Was there something else you're trying to address here?

PR #522, opened 2016-09-01
little clean

latot:
Hi again, well a little clean of some little things, use needlessly the lambda function have a bad effects in performance, so removing because binop_helper already support it.

Cya.

### DIFF
--- a/inst/@sym/and.m
+++ b/inst/@sym/and.m
@@ -44,7 +44,7 @@ function r = and(x, y)
     print_usage ();
   end
 
-  r = binop_helper(x, y, 'lambda a,b: And(a, b)');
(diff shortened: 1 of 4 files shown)


inst/private/python_header.py
```diff
@@ -43,6 +43,7 @@ def echo_exception_stdout(mystr):
     import struct
     import codecs
     import xml.etree.ElementTree as ET
+    import mpmath
```
cbm755: I'm not sure we want this change: I don't think there is a significant performance penalty to importing the same thing repeatedly.  And I think functions should not be using mpmath directly (where possible).

latot:
okidoki, i'll add that info to the wiki.

PR #523, opened 2016-09-01
rewrite diff2 

latot:
Well, this time its only rewrite of diff to python, why? time of execution:
with the actual diff
```
>> f=(x-1)^10;
>> tic;diff(f);toc;
Elapsed time is 0.0878189 seconds.
```
with python:
```
>> f=(x-1)^10;
>> tic;diff(f);toc;
Elapsed time is 0.0254591 seconds.
```

i have some questions, reading this: https://es.mathworks.com/help/symbolic/diff.html
and this tests:

```
%! assert (isequal (diff(f,sym(2),x), g))
...
%! assert (isequal (diff(f,1,x), g))
```

if i read right ```diff(f,sym(2),x)``` should diff f in 2 grades in x, and then derive it again in x? because in the example ```diff(f,sym(2),x)``` its the same as ```diff(f,sym(2))```...
i never used SMT so i don't know if that its right, but its doesn't feel it match with the doc.

Well Cya.

### DIFF
--- a/inst/@sym/diff.m
+++ b/inst/@sym/diff.m
@@ -66,40 +66,25 @@
 
 function z = diff(f, varargin)
 
-  % simpler version, but gives error on differentiating a constant
-  %cmd = 'return sp.diff(*_ins),';
-
-  %% some special cases for SMT compat.
-  % FIXME: with a sympy symvar, could move this to python?
-  if (nargin == 1)  % diff(f) -> symvar
-    x = symvar(f, 1);
-    if (isempty(x))
-      x = sym('x');  % e.g., diff(sym(6))
-    end
-    z = diff(f, x);
-    return
-  else
-    q = varargin{1};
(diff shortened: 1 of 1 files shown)


latot:
as a note, i don't change that tests yet, for that reason travis fails.

cbm755:
Re: `diff(f, 2, x)`: I dimly recall SMT used to do it this way, and at least in R2014b it is still mentioned in `help diff`.  I haven't checked if its supported in R2016---at least it doesn't seem to be mentioned in the link you sent.  Maybe we could remove that "feature": you could file a new issue for that.

cbm755:
I read release notes, I don't see it being removed from SMT... so maybe we leave it too for now.

cbm755:
I think the bit about `diff(f, 2, x)` is resolved.

Re: speed: I don't think we should do this sort of thing unless we write a re-useable "symvar" call in Python.  Its non trivial and there are also subtly like #550.

Shall I close this?

Issue #215, opened 2015-03-16
Functions it would be nice to have for compatibility


cbm755:
- [ ] bernoulli
- [ ] catalan
- [ ] charpoly
- [ ] chebyshevT
- [ ] chebyshevU
- [ ] coshint
- [ ] cosint
- [ ] dawson
- [ ] dilog
- [ ] dirac
- [ ] divisors
- [ ] ei
- [ ] ellipticCE
- [ ] ellipticCK
- [ ] ellipticCPi
- [ ] ellipticE
- [ ] ellipticF
- [ ] ellipticK
- [ ] ellipticPi
- [ ] euler
- [ ] eulergamma
- [ ] finverse
- [ ] fresnelc
- [ ] fresnels
- [ ] funtool
- [ ] gegenbauerC
- [ ] harmonic
- [ ] heaviside
- [ ] hermiteH
- [ ] hypergeom
- [ ] igamma
- [ ] jacobiP
- [ ] kroneckerDelta
- [ ] kummerU
- [ ] laguerreL
- [ ] legendreP
- [ ] logint
- [ ] minpoly
- [ ] pochhammer
- [ ] poles
- [ ] polylog
- [ ] quorem
- [ ] rectangularPulse
- [ ] signIm
- [ ] sinhint
- [ ] sinint
- [ ] ssinint
- [ ] triangularPulse
- [ ] whittakerM
- [ ] whittakerW
- [ ] zeta

These are from: http://hg.savannah.gnu.org/hgweb/octave/rev/fc6c87e254bf

AlexanderMisel:
I am using Octave 3.8.2 with OctSymPy v2.1.0 connecting SymPy 0.7.6 on LMDE. 
How can I `ezplot(heaviside(x)` as `x` being a symbol?
```
@(x) % Not supported in Octave:
% Heaviside
Heaviside(x)
```

cbm755:
@AlexanderMisel good catch.  Can you file a separate issue for this?

cbm755:
@AlexanderMisel see #226.

rocketsound:
I hope this is the right place: You might add `cross` to the list above

cbm755:
thanks!  patch to add `cross` definitely welcome :)

genuinelucifer:
Hi @cbm755!
I came to this page via Octave's GSoC page. I would like to work on this project for GSoC 2016.
[...]

PR #527, opened 2016-09-04
Cell2syms divisors

latot:
```
>>>>> processing @sym/cell2sym.m
***** test
 A = {1, 2, 3;4, 5, 6;7, 8, 9};
 B = [1, 2, 3;4, 5, 6;7, 8, 9];
 assert( isequal( cell2sym(A), sym(B)))
!!!!! test failed
'cell2sym' undefined near line 5 column 19
```

### DIFF
--- a/inst/@sym/cell2sym.m
+++ b/inst/@sym/cell2sym.m
@@ -1,0 +1,55 @@
+%% Copyright (C) 2016 Lagu
+%%
+%% This program is free software; you can redistribute it and/or
+%% modify it under the terms of the GNU General Public
+%% License as published by the Free Software Foundation;
+%% either version 3, or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied
+%% warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+%% PURPOSE.  See the GNU General Public License for more
+%% details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.  If not,
+%% see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym beta (@var{x})
+%% Convert cell array to symbolic array.
+%%
+%% Examples:
+%% @example
+%% @group
+%% syms x y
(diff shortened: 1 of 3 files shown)


cbm755:
If you put this in `inst/` instead of `inst/@sym/` does that solve the problem you mentioned elsewhere?  Things in `inst/@sym/` are only called if at least one input is type `sym` (which cell is not).

inst/@sym/cell2sym.m
```diff
@@ -0,0 +1,56 @@
...
+%% Convert cell array to symbolic array.
+%%
+%% Examples:
+%% @example
+%% @group
+%% syms x y
+%% cell2sym({x, y})
+%%   @result{} ans = (sym) [x  y]  (1×2 matrix)
+%% @end group
+%% @end example
```
cbm755: Can you add an example like `cell2sym({'x', 2; pi 'y'})'?  Or does that not work yet?  It would be an `xtest` if its not working yet.

inst/@sym/cell2sym.m
```diff
@@ -0,0 +1,56 @@
...
+
+function c = cell2sym(p)
+
+  if (nargin < 1 || nargin > 2)
+    print_usage ();
+  elseif (nargin == 2)
+    error('Flag not supported yet');
+  end
+
+  s = size(p);
```
cbm755: Add `assert (length (s) == 2)`?

inst/@sym/cell2sym.m
```diff
@@ -0,0 +1,56 @@
...
+  c = sym([]);
+
+  for i=1:s(1)
+    for j=1:s(2)
+      c(i, j)=p{i, j};
```
cbm755: spaces around these `=` please

inst/@sym/cell2sym.m
```diff
@@ -0,0 +1,56 @@
...
+    end
+  end
+
+end
+
+
+%!test
+%! A = {1, 2, 3;4, 5, 6;7, 8, 9};
+%! B = [1, 2, 3;4, 5, 6;7, 8, 9];
```
cbm755: not a big deal but spaces after these internal semicolons: `3; 4` please

latot:
Ready!

cbm755:
xref with #215.

PR #529, opened 2016-09-04
add simplify identities

latot:
Hi, well this time i propose implement in simplify the expand_func, its very useful specially if we use special functions like beta or gamma, in this case the user can choose how treat the function, in sympy we have 3 options to simplify, "simplify", "combsimp", and "expand_func", actually simplify can cover the combsimp functions, so we only add two.

its possible use full simplify like:
```
y = python_cmd('return simplify(expand_func(*_ins)),', x);
```
instead of choose the method, what do you think?

Cya.


inst/@sym/simplify.m
```diff
@@ -19,8 +19,13 @@
 %% -*- texinfo -*-
 %% @documentencoding UTF-8
 %% @defmethod @@sym simplify (@var{x})
+%% @defmethodx @@sym simplify (@var{x}, @var{y})
 %% Simplify an expression.
 %%
+%% @var{y} can be 'special' or 'complete'.
+%% 'special' will expand the equation using identities to simplify.
```
latot: i choose 'special' because are special functions, like beta, gamma, etc.

Issue #13, opened 2014-04-13
eye(3,'sym'), zeros, ones, etc are missing


cbm755:
One way is probably with static methods, which i think needs classdef.

So wait for octave to get classdef?

cbm755:
Min Octave version is 4 which has classdef so this might be possible now.

PR #532, opened 2016-09-05
(Z)ero (O)nes (E)yes

latot:
mm, originally i make this funcs because are used in the hilb function (the actual don't use it), but now i think this can help to have a better work flow, when we decide the number of matrix in function of some calc we most start using ```zeros(doubles(var))```, basically to simplify this.
Saying in other way, full symbolic code!

### DIFF
--- a/inst/@sym/eye.m
+++ b/inst/@sym/eye.m
@@ -38,7 +38,7 @@
 %% @seealso{@@sym/zeros, @@sym/ones}
 %% @end defmethod
 
-%% Source: http://docs.sympy.org/dev/modules/matrices/matrices.html
+%% Reference: http://docs.sympy.org/dev/modules/matrices/matrices.html
 
 
 function y = eye(varargin)
--- a/inst/@sym/eye.m
+++ b/inst/@sym/eye.m
@@ -20,7 +20,6 @@
 %% @documentencoding UTF-8
 %% @defmethod @@sym eye (@var{n})
 %% @defmethodx @@sym eye (@var{n}, @var{m})
-%% @defmethodx @@sym eye (@var{n}, @var{m}, @var{class})
 %% Return an identity matrix.
 %%
 %% Example:
--- a/inst/@sym/eye.m
+++ b/inst/@sym/eye.m
@@ -18,9 +18,9 @@
 
 %% -*- texinfo -*-
 %% @documentencoding UTF-8
-%% @defmethod @@sym eye (@var{x})
-%% @defmethod @@sym eye (@var{x}, @var{y})
-%% @defmethod @@sym eye (@var{x}, @var{y}, @var{class})
+%% @defmethod @@sym eye (@var{n})
+%% @defmethodx @@sym eye (@var{n}, @var{m})
+%% @defmethodx @@sym eye (@var{n}, @var{m}, @var{class})
 %% Return an identity matrix.
 %%
 %% Example:
@@ -28,11 +28,11 @@
 %% @group
 %% y = eye (sym(3))
 %%   @result{} y = (sym 3×3 matrix)
-%%  ⎡1  0  0⎤
-%%  ⎢       ⎥
-%%  ⎢0  1  0⎥
-%%  ⎢       ⎥
-%%  ⎣0  0  1⎦
+%%       ⎡1  0  0⎤
+%%       ⎢       ⎥
+%%       ⎢0  1  0⎥
+%%       ⎢       ⎥
+%%       ⎣0  0  1⎦
 %% @end group
 %% @end example
 %%
@@ -41,37 +41,50 @@
 
 %% Source: http://docs.sympy.org/dev/modules/matrices/matrices.html
 
+
 function y = eye(varargin)
 
-  if (nargin >= 2)
-    for i=1:size(varargin)(2)
-      if (ischar(varargin{i}))
-        y = sym(eye(cell2nosyms(varargin){:}));
-        return;
-      end
-    end
+  % partial workaround for issue #13: delete when/if fixed properly
+  if (strcmp (varargin{nargin}, 'sym'))
+    nargin = nargin - 1;
+    varargin = varargin(1:nargin);
   end
 
-  if (nargin > 1)
-    y = sym(eye(cell2nosyms(varargin){:}));
-    return;
+  if (isa (varargin{nargin}, 'char'))
+    y = eye (cell2nosyms (varargin){:});
+    return
   end
 
-  %% Be careful, varargin should be always sym.
-  y = python_cmd('return eye(*_ins),', varargin{:});
+  if nargin > 1 %%Sympy don't support eye(A, B)
+    y = sym(eye (cell2nosyms (varargin){:}));
+  else
+    y = python_cmd ('return eye(*_ins)', sym(varargin){:});
+  end
+
 end
+
 
 %!test
 %! y = eye(sym(2));
-%! x = [1 0;0 1];
+%! x = [1 0; 0 1];
 %! assert( isequal( y, sym(x)))
 
 %!test
 %! y = eye(sym(2), 1);
-%! x = [1;0];
+%! x = [1; 0];
 %! assert( isequal( y, sym(x)))
 
 %!test
 %! y = eye(sym(1), 2);
 %! x = [1 0];
 %! assert( isequal( y, sym(x)))
+
+%% Check types:
+%!assert( isa( eye(sym(2), 'double'), 'double'))
+%!assert( isa( eye(3, sym(3), 'single') , 'single'))
+%!assert( isa( eye(3, sym(3)), 'sym'))
+%!assert( isa( eye(3, sym(3), 'sym'), 'sym'))
+
+%!xtest
+%! % Issue #13
+%! assert( isa( eye(3, 3, 'sym'), 'sym'))
--- a/inst/@sym/eye.m
+++ b/inst/@sym/eye.m
@@ -1,0 +1,77 @@
+%% Copyright (C) 2016 Lagu
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym eye (@var{x})
+%% @defmethod @@sym eye (@var{x}, @var{y})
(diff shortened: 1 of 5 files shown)


cbm755:
Related idea is #13: it would be nice if `zeros(3, 3, 'sym')` worked.

cbm755:
is `zeros(3, sym(4), 'double')` supposed to DTRT (do the right thing)?  Its not tested, also doesn't look to me like it will work.

Is this the right logic?

1. If the last argument is a `char` type, we convert the other arguments do doubles and pass everything to the regular `zeros`.
2. otherwise, we make an appropriate symbolic matrix.

latot:
mm, the question first would be, when we transform for example a char matrix to sym, it respects the condition? or we will get a normal matrix?

cbm755:
Well, here's what I would expect to happen:
````
eye(sym(2), 'double')       % double result
eye(3, sym(3), 'single')    % single result
eye(3, sym(3))              % symbolic result
````

These two will be broken but that is issue #13:
````
eye(3, sym(3), 'sym')     % symbolic result
eye(3, 3, 'sym')          % symbolic result
````

**Comment by latot** [comment]
Okidokis, now works! i hope this can cover your expectations.

(At least pass your tests)

**Comment by cbm755** [comment]
See my PR on your fork.

1.  Can you make those changes in eye and ones as well?

2.  In doctests, indent as far as the `u` of `result`.

3. Don't add my name to the copyright of eye and ones (as I won't have changed those).

**Comment by latot** [comment]
Are you sure you want this? that will disable this call:
```
zeros(1, 2, 'sym')
```

**Comment by cbm755** [comment]
Oh, that's true.  But I think upstream Octave would not like to see packages "layering" on top of builtins in that way.  We'd have to ask anyway.

Let's do it this way in `@sym/` for now: there may be other ways to fix #13.

File: inst/@sym/eye.m [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+%% @defmethodx @@sym eye (@var{n}, @var{m})
+%% @defmethodx @@sym eye (@var{n}, @var{m}, @var{class})
+%% Return an identity matrix.
+%%
+%% Example:
+%% @example
+%% @group
+%% y = eye (sym(3))
+%%   @result{} y = (sym 3×3 matrix)
+%%  ⎡1  0  0⎤
```
cbm755: indent til `⎡` is under `u` of result (didn't I say this already?)

inst/@sym/eye.m
```diff
@@ -0,0 +1,86 @@
...
+%%  ⎢0  1  0⎥
+%%  ⎢       ⎥
+%%  ⎣0  0  1⎦
+%% @end group
+%% @end example
+%%
+%% @seealso{@@sym/zeros, @@sym/ones}
+%% @end defmethod
+
+%% Source: http://docs.sympy.org/dev/modules/matrices/matrices.html
```
cbm755: Why do you add these `Source: ` lines?  Is it recommended by Octave docs?

inst/@sym/eye.m
```diff
@@ -0,0 +1,86 @@
...
+
+function y = eye(varargin)
+
+
```
cbm755: there is an extra blank line here

inst/@sym/eye.m
```diff
@@ -0,0 +1,86 @@
...
+  if (isa (varargin{nargin}, 'char'))
+    y = eye (cell2nosyms (varargin){:});
+    return
+  end
+
+  y = python_cmd ('return eye(*_ins)', varargin{:});
+
+end
+
+%!test
```
cbm755: two blank lines before tests please

inst/@sym/eye.m
```diff
@@ -0,0 +1,86 @@
...
+%! y = eye(sym(1), 2);
+%! x = [1 0];
+%! assert( isequal( y, sym(x)))
+
+%!test
+%! %% Check types:
+%! assert( isa( eye(sym(2), 'double'), 'double'))
+%! assert( isa( eye(3, sym(3), 'single') , 'single'))
+%! assert( isa( eye(3, sym(3)), 'sym'))
+%! assert( isa( eye(3, sym(3), 'sym'), 'sym'))
```
cbm755: I prefer these be separate tests: so when one fails we can see which one (see e.g., my commit in the PR I sent you).

inst/@sym/ones.m
```diff
@@ -0,0 +1,86 @@
...
+%% @defmethodx @@sym ones (@var{n}, @var{m})
+%% @defmethodx @@sym ones (@var{n}, @var{m}, @var{class})
+%% Return a matrix or @var{n}-dimensional array whose elements are all 1.
+%%
+%% Example:
+%% @example
+%% @group
+%% y = ones (sym(3))
+%%   @result{} y = (sym 3×3 matrix)
+%%  ⎡1  1  1⎤
```
cbm755: as above

inst/@sym/ones.m
```diff
@@ -0,0 +1,86 @@
...
+%% @end example
+%%
+%% @seealso{@@sym/zeros}
+%% @end defmethod
+
+%% Source: http://docs.sympy.org/dev/modules/matrices/matrices.html
+
+function y = ones(varargin)
+  
+
```
cbm755: remove one blank line

inst/@sym/ones.m
```diff
@@ -0,0 +1,86 @@
...
+  if (isa (varargin{nargin}, 'char'))
+    y = ones (cell2nosyms (varargin){:});
+    return
+  end
+
+  y = python_cmd ('return ones(*_ins)', varargin{:});
+
+end
+
```
cbm755: add one more blank line

inst/@sym/ones.m
```diff
@@ -0,0 +1,86 @@
...
+%! y = ones(sym(1), 2);
+%! x = [1 1];
+%! assert( isequal( y, sym(x)))
+
+%!test
+%! %% Check types:
+%! assert( isa( ones(sym(2), 'double'), 'double'))
+%! assert( isa( ones(3, sym(3), 'single') , 'single'))
+%! assert( isa( ones(3, sym(3)), 'sym'))
+%! assert( isa( ones(3, sym(3), 'sym'), 'sym'))
```
cbm755: as above

inst/@sym/private/cell2nosyms.m
```diff
@@ -0,0 +1,31 @@
...
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+function z = cell2nosyms(x)
+
+  for i=1:size(x)(1)
```
cbm755: How about just for `i = 1:length(x)`?   We can 1D index into a multi-d array: works fine, and is standard Octave practice to do that when possible.

inst/@sym/zeros.m
```diff
@@ -0,0 +1,85 @@
...
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym zeros (@var{n})
+%% @defmethodx @@sym zeros (@var{n}, @var{m})
+%% @defmethodx @@sym zeros (@var{n}, @var{m}, @var{class})
+%% Return a matrix or @var{n}-dimensional array whose elements are all 0.
```
cbm755: It is not `@var{n}-dimensional`.

How about `Return a symbolic matrix whos elements are all 0.`   (there are calls that make non-symbolic but let's pretend those are not methods of `sym`).

Repeat this for the other two functions.

cbm755: Actually, let's delete the line with `%% @defmethodx ... @var{class}`: that's really only for compatibility, doesn't need to be part of our docs!

inst/@sym/eye.m
```diff
@@ -0,0 +1,86 @@
...
+    return
+  end
+
+  y = python_cmd ('return eye(*_ins)', varargin{:});
+
+end
+
+%!test
+%! y = eye(sym(2));
+%! x = [1 0;0 1];
```
cbm755: `0; 0` (extra space)

inst/@sym/eye.m
```diff
@@ -0,0 +1,86 @@
...
+%! assert( isequal( y, sym(x)))
+
+%!test
+%! y = eye(sym(2), 1);
+%! x = [1;0];
```
cbm755: `1; 0` (I think I have said this many times!).   Precision, consistency and style are important: people judge our code based on these things: like going to a job interview with shirt buttons mis-aligned...

latot:
Updated!

inst/@sym/eye.m
latot: No, its because when i read or try to works with this function i think its very useful know where we can find exactly the functions in Sympy, instead of start searching in all the docs, because are a lot of it and with differents informations, this link its the full doc of that function.

inst/@sym/private/cell2nosyms.m
latot: okidokis

inst/@sym/zeros.m
latot: mm, i really don't like the idea don't doc all functions of the code, if we assume that as "not supported" we can simplify even more the code, for now i'll put that changes in other commit.

latot:
Re-Updated.

inst/@sym/eye.m
cbm755: I see, ok maybe `%% Reference: ...` is better for future.

inst/@sym/private/cell2nosyms.m
```diff
@@ -0,0 +1,29 @@
...
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+function z = cell2nosyms(x)
+
+  for i=1:length(x)
+      if (isa(x{i}, 'sym') || isa(x{i}, 'symfun'))
```
cbm755: minor nitpick but isn't a `symfun` also a `sym` so the "or" will never happen?  Also this block is intended too far.

inst/@sym/eye.m
latot: mm, the true i don't know what say here, its true but its little hard to me assimilate this things.

inst/@sym/private/cell2nosyms.m
latot: ups you are right i forgot an object can have multiples classes, the indent was because i remove a section and i forgot fix it.

latot:
okidokis, ReReRe-Checked

inst/@sym/zeros.m
cbm755: Understood.  It is supported, but from the user's point of view, `zeros(2, sym(3), 'double')` is more like the builtin `zeros`.  They don't need to read our docs to learn how to do that.

I don't feel that strongly, put it back if you want.

Actually, how about we do #566?

inst/@sym/eye.m
cbm755: Sorry, I meant it constructively.  I very much appreciate all your edits, hope you don't feel otherwise.  Just trying to help make them even better (and lower my review time ;-)

cbm755:
LGTM!  I will merge, any future improvements can be new PRs.  Thanks @latot!

inst/@sym/eye.m
latot: Thx, i appreciate yours intention, i'll try to be better!

Issue #530, opened 2016-09-05
Improve dawsonm


latot:
Improve dawson in a matrix.

cbm755:
1.  Why do we need this?  (SMT doesn't have it, Sympy doesn't have it)
2.  Even if we do want it, I'm not convinced #533 is correct: for example, wouldn't `exp` need to be `expm`?

latot:
Ups, the second part its right, should be expm, for the first, i'm not sure, generally this type of things are interpretations, i only make it for the idea, from https://github.com/cbm755/octsympy/issues/215.
I don't like to much left ideas on the air.

cbm755:
I'm sorry if [my comment](https://github.com/cbm755/octsympy/issues/215#issuecomment-244666406) was ambiguous: I don't think we want `dawsonm`: I just tried to say what it should look like **if** we wanted it.

PR #533, opened 2016-09-05
add dawsonm

latot:
Fixes https://github.com/cbm755/octsympy/issues/530


cbm755:
still not right, depends on non-existent `erfim`.  I don't think this is worth working on.

latot:
Okis, closing.

Issue #535, opened 2016-09-05
adjoint doesn't do the same thing as SMT


cbm755:
Looks like `.adjoint()` in sympy does what we call `ctranpose`.

@latot do you know how do fix this?  Otherwise, we could remove `adjoint`.

We should also document/explain what `adjoint(sym('x'))` does.

latot:
i can fix it, but can i add an option like adjoint(A, 'conjugate') to use it naturally in SymPy?

cbm755:
See example here: http://www.mathworks.com/help/symbolic/adjoint.html

New PR (or current code in master) does not do that example.

latot:
checking things! https://github.com/cbm755/octsympy/pull/536

PR #536, opened 2016-09-06
[WIP] fix adjoint

latot:
maybe if you like split this to adjoint and adjointc, but actually we don't have exactly funcs with that behaviors, and ajoint its technically part of SMT, and ajoint with conjugate part of SymPy, so i think make sense have both in the same place.

Fixes https://github.com/cbm755/octsympy/issues/535

### DIFF
--- a/inst/@sym/adjoint.m
+++ b/inst/@sym/adjoint.m
@@ -24,9 +24,14 @@
 %% Example:
 %% @example
 %% @group
-%% A = [sym(1); 2];
+%% syms x
+%% A = [x x^3; 2*x i];
 %% y = adjoint(A)
-%%   @result{} y = (sym) [1  2]  (1×2 matrix)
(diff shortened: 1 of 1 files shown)


latot:
first test of travis works, the second fails because travis is loose.

latot:
Okidokis, Fixed, a little question what means WIP? (i'm imagining "work in progress")

cbm755:
yes, that's right.  Helpful to mark PRs that are not ready for merging.

cbm755:
This still doesn't look like the right thing to me, see my comment https://github.com/cbm755/octsympy/issues/535#issuecomment-244856762

latot:
ups, you are right, i was confused because exist an equal name of adjoint in sympy and SMT, but are differents, now is right and match with the doc of SMT.

PR #538, opened 2016-09-06
Minor fixes

cbm755:
@latot thanks for all your recent work: I'm busy with other things but trying to review when I can!

Here's some minor cleanups to recent commits: please let me know if this is ok to merge.

### DIFF
--- a/inst/@sym/adjoint.m
+++ b/inst/@sym/adjoint.m
@@ -25,10 +25,11 @@
 %% @example
 %% @group
 %% A = [sym(1); 2];
-%% y =  adjoint(A)
+%% y = adjoint(A)
 %%   @result{} y = (sym) [1  2]  (1×2 matrix)
 %% @end group
 %% @end example
+%% @seealso{@@sym/ctranspose}
 %% @end defmethod
 
 %% Source: http://docs.sympy.org/dev/modules/matrices/matrices.html
--- a/NEWS
+++ b/NEWS
@@ -4,11 +4,19 @@ octsympy 2.5.0-dev
   * New symbolic commands:
 
           acosd
+          adjoint
+          angle
           asind
           atand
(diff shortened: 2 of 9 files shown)


latot:
Hi, thx for your concern, i'm agree with this changes, the only thing would be, its possible include a example in charpoly with symfuns?
because when i make that file one of my problems was we can't use charpoly to calc in that way:
```
>> a=sym([1 2;3 4])
a = (sym 2×2 matrix)

  ⎡1  2⎤
  ⎢    ⎥
  ⎣3  4⎦

>> mu=sym('mu');
>> charpoly(a, mu);
>> b=charpoly(a, mu)
b = (sym)

   2          
  μ  - 5⋅μ - 2

>> b(1)
ans = (sym)

   2          
  μ  - 5⋅μ - 2
```
idea to use it:
```
>> b(mu)=charpoly(a, mu)
b(mu) = (symfun)

   2          
  μ  - 5⋅μ - 2

>> b(1)
ans = (sym) -6
```
i'm little worried if the user can't eval it, i know they can but i think its better show are compatible with symfuns.

Note, wiki updated.
Note 2, the others pr can wait, when you have time check it please, some pr fails in travis for two reasons, travis don't works like this: https://travis-ci.org/cbm755/octsympy/jobs/157658063 or waiting your opinion to decide the course of it.

Thx Cya.

inst/@sym/adjoint.m
```diff
@@ -25,10 +25,11 @@
 %% @example
 %% @group
 %% A = [sym(1); 2];
-%% y =  adjoint(A)
+%% y = adjoint(A)
```
latot: its possible avoid this line for https://github.com/cbm755/octsympy/pull/536?
or you can merge it and then i update the pr.

Issue #531, opened 2016-09-05
Support charpoly(x) - Vector Form


latot:
To do this we will need some function to transform sums to vectors like:
```
x+y+z -> [x y z]
```

Cya.

cbm755:
Something else does this kind of thing.  sym2poly and poly2sym maybe?

latot:
O_O
Thx. (seems i need read more docs!)

PR #540, opened 2016-09-06
Update Charpoly

latot:
Fixes https://github.com/cbm755/octsympy/issues/531

Left update the doc with examples, but first this https://github.com/cbm755/octsympy/pull/538 or we will have branch problems.

### DIFF
--- a/inst/@sym/charpoly.m
+++ b/inst/@sym/charpoly.m
@@ -66,7 +66,7 @@
 %% @end group
 %% @end example
 %%
-%% If you don't send any @var{x}, you will get a vector with the
+%% If you don't pass in any @var{x}, you will get a vector with the
 %% coefficients of the polynomi:
 %% @example
 %% @group
--- a/inst/@sym/charpoly.m
+++ b/inst/@sym/charpoly.m
@@ -54,11 +54,17 @@ function y = charpoly(a, b)
   if (nargin >= 3)
     print_usage ();
   end
+
+  cmd = {'return _ins[0].charpoly(_ins[1]).as_expr(),'};
+
   if (nargin == 1)
-    error('Charpoly as vector its not supported now');
+    syms b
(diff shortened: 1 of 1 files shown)


inst/@sym/charpoly.m
```diff
@@ -54,11 +54,17 @@
   if (nargin >= 3)
     print_usage ();
   end
+
+  cmd = {'return _ins[0].charpoly(_ins[1]).as_expr(),'};
+
   if (nargin == 1)
-    error('Charpoly as vector its not supported now');
+    syms b
+    y = python_cmd(cmd , sym(a), b);
```
cbm755: does this cause some problem if `b` is in the matrix already?  Also, given #543, maybe we should just do some python code here.  Then we can use `b = Dummy()` for the temp variable.

Or better yet, there must be a command other than `as_expr` which extracts the coefficients into a list...

cbm755:
Also, will need another example.

inst/@sym/charpoly.m
```diff
@@ -65,23 +65,42 @@
 %%       λ  - 2⋅λ⋅x + x  - x
 %% @end group
 %% @end example
+%%
+%% If you don't send any @var{x}, you will get a vector with the
+%% coefficients of the polynomi:
```
cbm755: Change "send" to "pass in".  Polynomial is spelled wrong.

latot: Okidokis, Done

inst/@sym/charpoly.m
```diff
@@ -65,23 +65,42 @@
 %%       λ  - 2⋅λ⋅x + x  - x
 %% @end group
 %% @end example
+%%
+%% If you don't pass in any @var{x}, you will get a vector with the
+%% coefficients of the polynomi:
```
cbm755: polynomial still misspelled, I'll fix it after merge

PR #541, opened 2016-09-06
find all assumptions for sympy versions (ver2) - fully-automatic

latot:
Hi again, here i'm dealing with this little problem, basically octsympy only have specific assumptions to can use independent of the sympy version, and checking the docs of sympy it doesn't says what assumptions are supported for version, only in the dev:http://docs.sympy.org/dev/modules/core.html

So this code use the dev assumptions and detect which are available in the actual version of Sympy, actually i can't found a way to completely auto-detect (that's why ver1) it but if we update this list when Sympy bring new versions should works fine.

Edited:

Now is fully automated!

Cya.

### DIFF
--- a/inst/assumptions.m
+++ b/inst/assumptions.m
@@ -66,7 +66,7 @@
 %% @group
 %% A = assumptions('possible');
 %% sprintf('%s ', A@{:@})
-%%   @result{} ans = real positive negative integer even odd rational finite
+%%   @result{} ans = ...
 %% @end group
 %% @end example
 %%
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,8 @@
 octsympy 2.5.0-dev
 ==================
 
+  * The full set of assumptions implemented by Sympy can now be used.
+
   * New symbolic commands:
 
           acosd         cosd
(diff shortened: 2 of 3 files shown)


inst/assumptions.m
```diff
@@ -66,7 +66,7 @@
 %% @group
 %% A = assumptions('possible');
 %% sprintf('%s ', A@{:@})
-%%   @result{} ans = real positive negative integer even odd rational finite
+%%   @result{} ans = ...
```
latot: this now change version by version.

cbm755:
This is pretty cool!  So `syms p prime` and `isprime(p)` would work right?

Slightly uncomfortable with using the private `_assume_defined` Python object but if that is the best way to get the list of assumptions, then ok.  I wonder why I wanted to whitelist them... let me sleep on this before I merge.

Can you add a line to NEWS?  Maybe something like: "The full set of assumptions implemented by Sympy can now be used".

latot:
Yay! it works:
```
>> syms a prime
>> isprime(a)
ans =  1
```
Adding info to NEWS.

Good Night.

cbm755:
Ok, pulled the trigger.

Its possible this will screw up for someone who wants:
````
syms q prime
````
to give two symbols named `q` and `prime`.  But meh :)

PR #543, opened 2016-09-06
we work with syms, don't transform to doubles!!

latot:
Cleaning!

### DIFF
--- a/inst/@sym/sym2poly.m
+++ b/inst/@sym/sym2poly.m
@@ -1,5 +1,6 @@
 %% Copyright (C) 2003 Willem J. Atsma
 %% Copyright (C) 2014-2016 Colin B. Macdonald
+%% Copyright (C) 2016 Lagu
 %%
 %% This program is free software; you can redistribute it and/or
 %% modify it under the terms of the GNU General Public
@@ -20,7 +21,7 @@
 %% @documentencoding UTF-8
 %% @deftypemethod  @@sym {@var{doublec} =} sym2poly (@var{p})
 %% @deftypemethodx @@sym {@var{c} =} sym2poly (@var{p}, @var{x})
-%% Return vector of coefficients of a symbolic polynomial.
+%% Return a double type vector of coefficients of a symbolic polynomial.
 %%
 %% In the two-input form, the second argument @var{x} specifies the free
 %% variable; in this case this function returns a row vector @var{c} of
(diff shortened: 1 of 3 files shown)


cbm755:
I tend to agree.  But IIRC, Matlab's SMT does this :(  We should check.

latot:
well considering that SMT behavior, i propose use this, isOctave its like isunix, because its little bussy write everywhere ```exist('OCTAVE_VERSION', 'builtin')``` and its more easy to understand.

cbm755:
hmmm, maybe...  I think usually octave tries to be matlab compatible.  Does `@sym/coeffs` do what we want?

The `exist('OCTAVE_VERSION'` thing is recommended upstream.  See also https://github.com/catch22/octave-doctest/blob/master/inst/private/is_octave.m.  Maybe I would consider this but it needs to be its own PR.

latot:
yes, its basically the same, but in isOctave i avoid call it every time, and i try use it in private folder but the problem was the tests, these can't access to the private functions.
While i'll split this.

cbm755:
This is a pretty minor point: we already do what upstream suggests: maybe don't bother.  As you say, it won't work in tests anyway, which is the main place we use this.

latot:
Fixed!

cbm755:
Multiple behaviours is dangerous: need to test both, support both, etc.  Also, running on Matlab is not really being tested anymore (not by me anyway).

So either:

1.  we stay with SMT's rather foolish behaviour (AND SHOUT LOUDLY in our docs---as we already do).
2.  deliberately break compatibility.

I think 1.

cbm755:
Wait, we could add a `sym2polysane`.  Because `coeffs` does not really do the same thing (because it skips zeros).

Then `sym2poly` would just be a wrapper doing `double(sym2polysane(varargin{:}))`...   Thoughts?

latot:
Yay!

latot:
okidokis, Fixed!

inst/@sym/sym2polysane.m
```diff
@@ -0,0 +1,119 @@
...
+%% Created: 18 April 2003
+%% Changed: 25 April 2003
+%%    Removed the use of differentiate to get to coefficients - round-off
+%%     errors cause problems. Now using newly created sumterms().
+%% Changed: 6 May 2003
+%%    Removed the attempt to use ldegree(), degree() and coeff() - results
+%%     with these are inconsistent.
+%% Changed: 16 April 2014
+%%    Used the comment header and tests in OctSymPy, but rewrote
+%%    the body (by Colin Macdonald).
```
latot: Can i remove the history?

cbm755: Yes, go for it.

inst/@sym/sym2poly.m
```diff
@@ -30,29 +31,22 @@
 %% @group
 %% syms x y
 %% sym2poly(2*x^2 + 3*x - pi, x)
-%%    @result{} (sym) [2  3  -π]  (1×3 matrix)
+%%    @result{} ans =
+%%   2.0000   3.0000  -3.1416
 %% sym2poly(x^2 + y*x, x)
```
cbm755: Do you think we should change the behaviour when x is specified?  I guess that's more consistent behaviour, but its a change from previous Octsympy releases...

Also line 27/28 says we return a row vector of symbolic expressions...

latot: okidokis, while i'll open anew issue for this things.

inst/@sym/sym2poly.m
```diff
@@ -115,20 +73,19 @@
...
-%!assert (isequal (sym2poly (x^2+y*x, x), [sym(1) y sym(0)]))
-%!assert (isequal (sym2poly (x^2+y*x, y), [x x^2]))
+%! assert (isequal (sym2poly (x^2+y*x, x), [sym(1) y sym(0)]))
+%! assert (isequal (sym2poly (x^2+y*x, y), [x x^2]))
 %% inverse relationship
-%!assert (isequal (sym2poly (poly2sym ([a b c], x), x), [a b c]))
-%!assert (isequal (poly2sym (sym2poly(a*x^2 + c, x), x), a*x^2 + c))
-%!assert (isequal (sym2poly (poly2sym ([1 2 3])), [1 2 3]))
+%! assert (isequal (sym2poly (poly2sym ([a b c], x), x), [a b c]))
+%! assert (isequal (poly2sym (sym2poly(a*x^2 + c, x), x), a*x^2 + c))
```
latot: only to know, what is the meaning change?

latot:
Okidokis, Updated!

inst/@sym/sym2poly.m
cbm755: > only to know, what is the meaning change?

These are two tests:
````
%!assert ...
%!assert ...
````

These are two assert statements within one test:
````
%!test
%! assert ...
%! assert ...
````

latot:
Code ready and formatted.

cbm755:
I'm adding the new `coeffs(..., 'all')` support in #708.  Perhaps we no longer need `sym2polysane`?

latot:
okis, so will need rewrite this and use reverse in coeffs when is ready, Thx!

cbm755:
I mean, maybe we don't need this at all?

The documentation of `sym2poly` could be improved to point at `coeffs('all')`.

latot:
mm, i would like add that to the doc, and rewrite sym2poly with coeffs function, what do you think? (with that i think we can simplify code)

cbm755:
Sure, that sounds good if it leads to simpler code.

Issue #548, opened 2016-09-14
Improve support for querying available assumptions


latot:
Hi, well, i think this is necessary for this:

```
Ok, pulled the trigger.

Its possible this will screw up for someone who wants:

syms q prime
to give two symbols named q and prime. But meh :)
```

At least notice to the user.

if you don't like this maybe a way to check assumptions.

Cya.

cbm755:
"help assumptions" explains how to list them...

Interesting idea to display at startup.  But its already so noisy that people think something has gone wrong...

latot:
mm, so i think the best option can be do a new function to check assumptions, or expand the assumptions function. if you like what do you prefer?

latot:
i think expand the actual assumptions function can be a bad idea, i was thinking in this syntax:
```
assumptionsCheck prime real zero
```
and return warnings to every not found assumptions and an error if that case, but mix this with the actual assumption..., we can't mix it.

Well what do you think with this propose?

cbm755:
I don't see how it will help.  We already have `assumptions('possible')`.

Fundamentally, the `syms` syntax is ambiguous.  E.g., `syms foo bar meh`: is `meh` supposed to be a variable or assumption?  Currently its taken as an assumption if it matches anything in the output of `assumptions('possible')`.

Consider an example: `syms x postive`: here I have misspelled `positive` so I silently get two variables `x` and `postive`.  I don't think there is anything we can do about that.

mtmiller:
Agree with @cbm755 (feels like this issue is jumping in the middle of another thread that I'm missing). I don't think listing an "assumptions banner" at the start is a good idea, or overthinking it too much at all. If the user can't do what they want with the simple syntax of `syms`, then they should use `sym`, `symfun`, and `assume` instead.

latot:
@mtmiller its right, the initial point of this was that, but now the focus is found some way to do this more easy.

mm, the conflict its when ppl can't know if an assumption is available in scripts, part of the idea is keep compt with SMT, and personally i works a lot with scripts, if i run the script in some Octave/Matlab i need know if it will works, to do that we can do a new function to know this, like this:
```
assumeCheck prime polar
syms var prime polar
```
So, if prime or polar don't is available in that version of Sympy we get an error saying the not available  assumptions.

cbm755:
There will surely be a one-liner that does what you want, something like:
````
assert (any (strcmp (assumptions ('possible'), 'prime')))
````

[...]

PR #551, opened 2016-09-14
add isassumption

latot:
You know.

Fixes https://github.com/cbm755/octsympy/issues/548

### DIFF
--- a/inst/@sym/assume.m
+++ b/inst/@sym/assume.m
@@ -99,7 +99,7 @@
 %% FIXME: idea of rewriting all sym vars is a bit of a hack, not
 %% well tested (for example, with global vars.)
 %%
-%% @seealso{@@sym/assumeAlso, assumptions, sym, syms}
+%% @seealso{@@sym/assumeAlso, assumptions, sym, syms, isassumption}
 %% @end deftypemethod
 
 %% Author: Colin B. Macdonald
--- a/inst/@sym/assumeAlso.m
+++ b/inst/@sym/assumeAlso.m
@@ -60,7 +60,7 @@
 %% and replace any @var{x} within expressions in the caller's
 %% workspace.  See @ref{assume}.
 %%
-%% @seealso{@@sym/assume, assumptions, sym, syms}
+%% @seealso{@@sym/assume, assumptions, sym, syms, isassumption}
 %% @end deftypemethod
 
 %% Author: Colin B. Macdonald
--- a/inst/assumptions.m
+++ b/inst/assumptions.m
@@ -70,7 +70,7 @@
 %% @end group
 %% @end example
 %%
-%% @seealso{sym, syms, @@sym/assume, @@sym/assumeAlso}
+%% @seealso{sym, syms, isassumption, @@sym/assume, @@sym/assumeAlso}
 %% @end deftypefun
 
 %% Author: Colin B. Macdonald
--- a/inst/isassumption.m
+++ b/inst/isassumption.m
@@ -23,7 +23,7 @@
 %%
 %% @example
 %% @group
-%% isassumption(@{"real", "prime", "polar"@})
+%% isassumption (@{"real", "prime", "polar"@})
 %%   @result{} ...
 %% @end group
 %% @end example
@@ -46,11 +46,11 @@
 
 function out = isassumption(in)
 
-  list = assumptions('possible');
+  list = assumptions ('possible');
   out = [];  
 
-  for i=1:length(in)
-    if (isempty(strmatch(in{i}, list, 'exact')))
+  for i=1:length (in)
+    if isempty (strmatch (in{i}, list, 'exact'))
       out = [out, i];
     end
   end
@@ -59,5 +59,5 @@ end
 
 
 %!test
-%! out = isassumption({"blabla", "mehmeh"});
-%! assert( isequal( out, [1, 2]))
+%! out = isassumption ({"blabla", "mehmeh"});
+%! assert (isequal (out, [1, 2]))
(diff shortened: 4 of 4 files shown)


cbm755:
See https://github.com/cbm755/octsympy/issues/548#issuecomment-247174684

cbm755:
I like what @mtmiller said would be better: just make it return true or false: more generally useful than raising an error.

mtmiller:
I also think it would be more useful to behave more like `isa` or `ishandle`, take a single argument that is an array, and return an array of `logical` of the same size. The caller can then use `any` or `all` if they want to interpret the results either way.

cbm755:
... and @latot's script would start with
````
assert (all (isassumption ({'real', 'finite'})), "unsupported assumptions, upgrade your SymPy?")
````

mtmiller:
Yes that looks better to me.

Incidentally what versions of Sympy are we talking about here?

cbm755:
IIRC, `0.7.5` used `bounded` instead of `finite`.  I dropped support for 0.7.5.
No idea after that.

@mtmiller: This is all happening because we (octsympy) used to have a whitelist of allowable assumptions.  @latot changed it to that whatever assumptions SymPy has are exposed.

mtmiller:
Thanks, I saw that change, just wasn't sure which versions of Sympy are actually affected, e.g. missing `prime` or `rational` assumption labels, etc.

mtmiller:
Grepping the sympy history, it looks like 0.7.6 has the same list of assumptions that 1.0 has. Yes, 0.7.6 added `algebraic`, `rational`, `transcendental`, `infinite`, and `nonzero` over 0.7.5.

Just curious what are the practical cases being addressed here. Earlier comments suggested that `real` or `prime` or `polar` may or may not be supported in a particular version of Sympy, but that actually seems to not be possible.

cbm755:
I'm not aware of a practical use case.

(There has been a change in octsympy in that `syms q prime` used to make two variables, now it makes a `q` which is assumed prime.  But adding `isassumption` doesn't help with that.)

latot:
Hi, mm, when i do assumeCheck i think basically in this things:
   * If the user use an assumption it can't be replaced, that mean in general the user can't do something to cover it if don't is supported, and if can do something the most probable is will need rewrite part of the script, so the best its get an error i think.
   * If don't exist an assumption the user shouldn't start searching in all are using which isn't supported, for that reason this function check one by one. 

So, with this in consideration, what are yours expectations for this functions as users and develop?

mtmiller:
I understand what you want to accomplish. You want a very easy `assert`-like syntax where you can do something like

    assert_assumptions real positive prime foo bar
    syms x y z real positive prime foo bar

with a hardcoded error message listing the assumptions that are not supported. If you both agree this is something useful for the user, don't let me stop you.

IMHO having a function check a cell array of assumptions is much more flexible and useful, to have either _instead_ or _in addition_ to what you want.

I would rather use something like this

    names = {"real", "positive", "prime", "foo", "bar"};
    idx = isassumption (names);
    if (any (names(idx))
      error ("myfunc: these assumptions are not supported: %s", strjoin (names(!idx), ", "));
    endif

This gives me the flexibility to error if I want to, with my own error message format, or to try to continue without certain assumptions if possible.

If you want to have both options available, then your check/assert function could be written like what I've shown here.

To me, the ambiguous `syms` command syntax is best suited for REPL interactive work anyway.

latot:
Okidokis, thx for your comment, when i have a little time i'll write it.

cbm755:
>  If you both agree this is something useful for the user, don't let me stop you.

FTR, I *do not* think it is useful.

latot:
Okidokis, what do you think?

latot:
Code ready and formatted.

cbm755:
There are variously places in the code where we check if a string is an assumption.  I think at least in `syms` and possible in `@sym/sym`: basically everywhere where `assumptions('possible')` is called.  Could any of that code be simplified to instead use the new `isassumption`?  If it cleans up that existing code, its probably worth merging (even if I don't have a use for this myself).

latot:
mm, you are right, i fix that concept in this pr: https://github.com/cbm755/octsympy/pull/594
i think i can update it or mix with the classdeff.

cbm755:
I think a separate PR that builds on top of this would be good.  #594 is too big!

latot:
yes, but sadly i try split it, but test fails to much, i fix a concept and something send an error, then i fix it and again other, for that reason is too big.... D:

cbm755:
Its ok, we can chip away at it, `@sym/sym` is a mess so probably good ideas in there.  But let's tackle the `isassumption` thing first, separately.  So leave this PR, and make another one that builds on it to use `isassumption` internally.

cbm755:
Maybe unclear: I mean the *current* `@sym/sym` is a mess, so probably good ideas in your PR to clean it up.

latot:
aa, right, one of the ideas of that pr is do that!

cbm755:
Possible to merge this before #594?  Just need to use the `isassumption` stuff as mentioned above in https://github.com/cbm755/octsympy/pull/551#issuecomment-251535399

latot:
Hi, yes, i'll need update this function, because sym function need a better checker, i'll notice when this is ready.

latot:
Hi, after read and think this, i conclude with the actual behaviors we can't use this function for the sym function, for one reason, we don't have a way to can index structs, sym support cells/chars and structs, assume, assumeAlso uses structs to get and set assumptions.
So, for now i can see 2 ways, first use two functions to check assumptions, one for user, and other for sym function to can keep the actual functionalities, the other way, remove struct support from sym function and rewrite the functions to use cells/char to set and get the assumptions.

Thx. Cya.

cbm755:
Note: now there is an "private/check_assumptions" in master.

cbm755:
not sure if there is anything here left to merge, closing.  Of course reopen if you still wanted to work on this.

Issue #448, opened 2016-05-17
solving inequations and inequalities for sets gives errors


cbm755:
````
syms x real
solve(x < 5, x)
solve(x ~= 5, x)
solve(x < 5)
solve(x ~= 5)
solve(x == 5, x)    # this one works
````
I think SymPy can solve these; we just need to wire it up properly...

Might need a recent SymPy version.

genuinelucifer:
I thought to work on this but I am unable to understand the exact mechanism of it...
If I do `solve(x == 5, x)` then the `arg` in python is `Eq(x, 5)` while if I use any other operator then it does not change. How is `x == 5` getting transformed to `Eq(x,5)` while others are not?

cbm755:
Testing in Sympy directly:

````
>>> from  sympy import *
>>> x = S('x')
>>> solve(Lt(x, 5), x)
And(-oo < x, x < 5)
````
But if I put them in lists:
````
>>> solve([Lt(x, 5)], [x])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-14-f68f532a789a> in <module>()
----> 1 solve([Lt(x, 5)], [x])

/home/cbm/work/octsympy/octsympy.git/inst/sympy/solvers/solvers.pyc in solve(f, *symbols, **flags)
    820             f[i] = fi.as_expr()
    821         elif isinstance(fi, (bool, BooleanAtom)) or fi.is_Relational:
--> 822             return reduce_inequalities(f, symbols=symbols)
    823 
    824         # rewrite hyperbolics in terms of exp

/home/cbm/work/octsympy/octsympy.git/inst/sympy/solvers/inequalities.pyc in reduce_inequalities(inequalities, symbols)
    584     if not iterable(symbols):
    585         symbols = [symbols]
--> 586     symbols = (set(symbols) or gens) & gens
    587     if any(i.is_real is False for i in symbols):
    588         raise TypeError(filldedent('''

TypeError: unhashable type: 'list'
````

cbm755:
So maybe its an upstream problem?  We could try to work around, but I generally prefer fixing sympy.  And this function already has a lot of python code (generally a bad thing as Symbolic should be as thin as possible).

PR #557, opened 2016-09-17
Inequalities go to work!

latot:
Fixes https://github.com/cbm755/octsympy/issues/448

### DIFF
--- a/inst/@sym/solve.m
+++ b/inst/@sym/solve.m
@@ -23,7 +23,7 @@
 %% @deftypemethodx @@sym {@var{sol} =} solve (@var{eq1, eq2})
 %% @deftypemethodx @@sym {@var{sol} =} solve (@var{eq1, @dots{}, eqn, v1, @dots{}, vm})
 %% @deftypemethodx @@sym {[@var{s1, @dots{}, sn}] =} solve (@var{eq1, @dots{}, eqm, v1, @dots{}, vn})
-%% Symbolic solutions of equations and systems.
+%% Symbolic solutions of equations, inequalities and systems.
 %%
 %% Examples
 %% @example
@@ -83,11 +83,30 @@
 %% @end group
 %% @end example
 %%
+%% You can solve inequalities with or without equations:
+%%
+%% Example:
+%% @example
+%% @group
+%% solve(x^2 == 4, x > 0)
+%%   @result{} ans = (sym) x = 2
+%% @end group
+%% @end example
+%%
+%% Example:
+%% @example
+%% @group
+%% solve(x^2 - 1 > 0, x < 10)
+%%   @result{} ans = (sym) (-∞ < x ∧ x < -1) ∨ (1 < x ∧ x < 10)
+%% @end group
+%% @end example
+%%
 %% @seealso{@@sym/eq, @@sym/dsolve}
 %% @end deftypemethod
 
 %% Author: Colin B. Macdonald, Andrés Prieto
 %% Keywords: symbolic
+
 
 function varargout = solve(varargin)
 
@@ -101,11 +120,10 @@ function varargout = solve(varargin)
             '    else:'
             '        symbols.append(arg)'
             '#'
-            'if len(symbols) > 0:'
-            '    d = sp.solve(eqs, symbols, dict=True)'
-            'else:'
-            '    d = sp.solve(eqs, dict=True)'
+            'd = sp.solve(eqs, *symbols, dict=True)'
             '#'
+            'if not isinstance(d, (list, tuple)):'
+            '    return d,'
             'if len(d) >= 1 and len(d[0].keys()) == 1:'  % one variable...
             '    if len(d) == 1:'  % one variable, single solution
             '        return d[0].popitem()[1],'
@@ -128,15 +146,15 @@ function varargout = solve(varargin)
             '    else:'
             '        symbols.append(arg)'
             '#'
-            'if len(symbols) > 0:'
-            '    (vars, solns) = sp.solve(eqs, symbols, set=True)'
-            'else:'
-            '    (vars, solns) = sp.solve(eqs, set=True)'
+            'd = sp.solve(eqs, *symbols, set=True)'
             '#'
-            'd = []'
+            'if not isinstance(d, (list, tuple)):'
+            '    return d,'
+            '(vars, solns) = d'
+            'q = []'
             'for (i, var) in enumerate(vars):'
-            '    d.append(sp.Matrix([t[i] for t in solns]))'
-            'return d,' };
+            '    q.append(sp.Matrix([t[i] for t in solns]))'
+            'return q,' };
 
     out = python_cmd (cmd, varargin{:});
 
@@ -229,6 +247,11 @@ end
 %! [X, Y] = solve(x*x == 4, x == 2*y, x, y);
 %! assert (isequal (X, [2; -2]))
 %! assert (isequal (Y, [1; -1]))
+
+%!test
+%! syms x
+%! a = solve(2*x >= 10, 10*x <= 50);
+%! assert (isequal( a, x==sym(5)))
 
 %!error
 %! syms a b


inst/@sym/solve.m
```diff
@@ -129,22 +149,23 @@
...
+            'return q,' };
 
     out = python_cmd (cmd, varargin{:});
 
-    varargout = out;
+    assert( isequal( nargout, length(out)), ['This system have ' num2str(length(out)) ' solution, favor set correct numbers of output.']);
 
-    if (length(out) ~= nargout)
-      warning('solve: number of outputs did not match solution vars');
-    end
```
latot: This can be removed, if the length of nargout != length of out octave will handle the error, so before that we can send a personalized error:
```
>> [X Y Z]=solve(x^2 == 4, x + y == 1, x, y);
warning: solve: number of outputs did not match solution vars
warning: called from
    solve at line 146 column 7
error: element number 3 undefined in return list
```

latot:
Hi, when you can, plis check this pr and https://github.com/cbm755/octsympy/pull/559, i think this both should have priority (compared with all others pr i did) (taylor i think its important too but we will wait to @genuinelucifer)

Thx.

cbm755:
Would it be hard to split this into several commits?

1.  There is a refactor of the code: this can be a good thing, but ideally its a separate commit.
2.  There is a fix for #448 (but I haven't found it yet b/c of the refactor).


Re: the refactor: I don't like the use of `eval`.
  * can we just pass true, false in for both `set=` and `dict=`?
  * if not, try `strrep` on `cmd`, the cellarray-of-strings

latot:
okidokis, i'll split this.

inst/@sym/solve.m
```diff
@@ -93,20 +111,23 @@
...
+  base = { 'eqs = list(); symbols = list()'
+           's = _ins[0]'
+           '_ins = _ins[1:]'
+           'for arg in _ins:'
+           '    if arg.is_Relational:'
+           '        eqs.append(arg)'
+           '    else:'
+           '        symbols.append(arg)'
+           'if len(symbols) > 0:'
+           '    d = eval("sp.solve(eqs, *symbols, " + s + ")")'
```
latot: This fixes the issue, send symbols with unpack operator:
```
*symbols
```

inst/@sym/solve.m
```diff
@@ -116,35 +137,22 @@
...
 
-    out = python_cmd (cmd, varargin{:});
+    out = python_cmd ([strrep(base, '{extra}', ', set=True'); cmd], varargin{:});
 
-    varargout = out;
+    assert( isequal( nargout, length(out)), ['This system have ' num2str(length(out)) ' solution, favor set correct numbers of outputs.']);
 
-    if (length(out) ~= nargout)
-      warning('solve: number of outputs did not match solution vars');
-    end
```
genuinelucifer: I think this should not be removed. @cbm755?

latot:
Okidokis!, i don't like the idea of true/solve in dict/set, i think the idea of a base cell its be enough flexible to avoid refactor that part in the future.

Your requests are ready!

inst/@sym/solve.m
```diff
@@ -116,35 +137,22 @@
...
-            '    d.append(sp.Matrix([t[i] for t in solns]))'
-            'return d,' };
+            '    q.append(sp.Matrix([t[i] for t in solns]))'
+            'return q,' };
 
-    out = python_cmd (cmd, varargin{:});
+    out = python_cmd ([strrep(base, '{extra}', ', set=True'); cmd], varargin{:});
 
-    varargout = out;
+    assert( isequal( nargout, length(out)), ['This system have ' num2str(length(out)) ' solution, favor set correct numbers of outputs.']);
```
genuinelucifer: Very minor. 'assert ('   <space>...

inst/@sym/solve.m
```diff
@@ -230,6 +238,11 @@
 %! assert (isequal (X, [2; -2]))
 %! assert (isequal (Y, [1; -1]))
 
+%!test
+%! syms x
+%! a = solve(2*x >= 10, 10*x <= 50);
+%! assert( isequal( a, x==sym(5)))
```
genuinelucifer: Space after assert..

inst/@sym/solve.m
latot: Thx, fixed.

inst/@sym/solve.m
latot: Thx, fixed.

genuinelucifer:
Sorry for nit-picking, but still there are many other places (i dint comment on) where spaces r jumbled up.
Use space after a function name but not after the opening parenthesis. Also use spaces around operators...

inst/@sym/solve.m
cbm755: I also don't understand this change: @latot explains above but I don't understand if its a bug fix, strengthening warning to error, or just adding some more information.

Also `s/favor/please/`

cbm755:
Following on from @genuinelucifer's comment: Python code *should not* follow these standards, but rather "PEP 8".

My old code, especially my tests, often does not follow Octave style.  But we should collectly do better going forward.

inst/@sym/solve.m
```diff
@@ -93,20 +111,23 @@
...
+  base = { 'eqs = list(); symbols = list()'
+           'for arg in _ins:'
+           '    if arg.is_Relational:'
+           '        eqs.append(arg)'
+           '    else:'
+           '        symbols.append(arg)'
+           'if len(symbols) > 0:'
+           '    d = sp.solve(eqs, *symbols{extra})'
+           'else:'
+           '    d = sp.solve(eqs{extra})'
```
cbm755: Can you drop the `len(symbols) > 0` check now that you use `*symbols`?

inst/@sym/solve.m
latot: mm, i think its a little of the last two things, basically the first thing its octave send an error, but you know, that don't have to much info, so with the assert we can say to user the problem.
in other words, when the requested values don't math with out we have the warning, but then octave send the error.., at least to me don't have to many sense, we can just send the error with the info.

latot:
Hi, ready!, for some reason when i try remove the ```len(symbols) > 0``` i get errors, but now works, so removed.
About the format, sadly we can't use it in cancat of cells, because its recognized as separated elements:
```
[f(a) f(b)] ~= [f (a) f (b)]
```

cbm755:
Re format: yes, this is documented in octave

cbm755:
I'm not sure about the refactor...

If not too much trouble, you could move the test and new docs to the fix ineq commit?  Then file a new PR for the refactor?

latot:
Okis, i don't know what its the problem with the refactor..., its basically the same except for the assert.

latot:
Code Updated!

latot:
For example, we can't pass this test:
```
%!test
%! syms x
%! [a b] = solve(x^2+x == 0)
```

inst/@sym/solve.m
```diff
@@ -101,11 +120,10 @@
...
             '        symbols.append(arg)'
             '#'
-            'if len(symbols) > 0:'
-            '    d = sp.solve(eqs, symbols, dict=True)'
-            'else:'
-            '    d = sp.solve(eqs, dict=True)'
+            'd = sp.solve(eqs, *symbols, dict=True)'
             '#'
+            'if not isinstance(d, (list, tuple)):'
+            '    return d,'
```
cbm755: what does this change do?

You asked why not comfortable with the previous refactor; this is one reason, new things happening that I don't understand.  Does this fix something?  If so, there is no test.

latot: Hi, when we works with ineq Sympy some times don't return lists with solutions and only send an expression with some conditions, so that case we need send that or we will get an error in the next steps of the script, sadly this is related with the non sense issue of here https://github.com/cbm755/octsympy/issues/581, this was on of the reasons of the assert in the  refactor.

latot: This command will fail without it:
```
solve(x*x==1, x>0)
```
and the test exist, is the new, check line 251

latot: To know about all this problems please check this issues of Symy:
11653 to 11662.

latot: The specific reason of that line is this issue: https://github.com/sympy/sympy/issues/11661

Issue #290, opened 2016-01-10
`syms FF(w)` fails


cbm755:

latot:
from python:
```
z = sympy.S("FF(w)")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python2.7/site-packages/sympy/core/sympify.py", line 322, in sympify
    expr = parse_expr(a, local_dict=locals, transformations=transformations, evaluate=evaluate)
  File "/usr/lib64/python2.7/site-packages/sympy/parsing/sympy_parser.py", line 892, in parse_expr
    return eval_expr(code, local_dict, global_dict)
  File "/usr/lib64/python2.7/site-packages/sympy/parsing/sympy_parser.py", line 805, in eval_expr
    code, global_dict, local_dict)  # take local objects in preference
  File "<string>", line 1, in <module>
  File "/usr/lib64/python2.7/site-packages/sympy/polys/domains/pythonfinitefield.py", line 17, in __init__
    return super(PythonFiniteField, self).__init__(mod, PythonIntegerRing(), symmetric)
  File "/usr/lib64/python2.7/site-packages/sympy/polys/domains/finitefield.py", line 29, in __init__
    if mod <= 0:
  File "/usr/lib64/python2.7/site-packages/sympy/core/relational.py", line 196, in __nonzero__
    raise TypeError("cannot determine truth value of Relational")
TypeError: cannot determine truth value of Relational
```

cbm755:
Thanks.  Might be related to https://github.com/sympy/sympy/issues/9057.

latot:
i think this isn't a bug, actually sympy claims if you try to make a sym of a function, its the same with matrix:
```
>>> S("Matrix(w)")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python2.7/site-packages/sympy/core/sympify.py", line 322, in sympify
    expr = parse_expr(a, local_dict=locals, transformations=transformations, evaluate=evaluate)
  File "/usr/lib64/python2.7/site-packages/sympy/parsing/sympy_parser.py", line 894, in parse_expr
    return eval_expr(code, local_dict, global_dict)
  File "/usr/lib64/python2.7/site-packages/sympy/parsing/sympy_parser.py", line 807, in eval_expr
    code, global_dict, local_dict)  # take local objects in preference
  File "<string>", line 1, in <module>
[...]

PR #559, opened 2016-09-18
try block to check sym creation

latot:
Fixes https://github.com/cbm755/octsympy/issues/290

Hi again, well thinking the last pr complicate things with the char command, so instead of disable the env functions use a try block to send the error, this is more... experimental?, or try to keep the same flexibility of the actual code, but this keep a similar problem making symbols:

```
>> sym('Eq(t)')
ans = (sym) t = 0
```

(for a normal user this is weird)
at least this is a step to handle this.
maybe send a warning when we use python code from the sym function?

Choose your preferred way to fix this :D 

Thx. Cya.

### DIFF
--- a/inst/@sym/sym.m
+++ b/inst/@sym/sym.m
@@ -238,9 +238,9 @@ function s = sym(x, varargin)
 
   elseif (islogical (x)  &&  isscalar(x)  &&  nargin==1)
     if (x)
-      cmd = 'z = sp.S.true';
+      cmd = {'z = sp.S.true'};
     else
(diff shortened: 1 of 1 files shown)


latot:
Deprecated pr, cover in https://github.com/cbm755/octsympy/pull/594

cbm755:
Quite a lot of this has changed.  We now support and test for this:
```
octave:10> syms Eq(t)
octave:11> Eq
Eq(t) = (symfun) Eq(t)
octave:12> sympy(formula(Eq))
ans = Function('Eq')(Symbol('t'))
```

Issue #31, opened 2014-04-30
taylor expansions need work


cbm755:
see taylor.m.  Hard-coded for 8 terms for example.

cbm755:
Sure @trueutkarsh that would be great.

Maybe this is helpful?   http://stackoverflow.com/questions/22857162/multivariate-taylor-approximation-in-sympy

trueutkarsh:
@cbm755 i have written the python code(function) which does it  ,tested and compared it with few outputs from the matlab .Its working fine.
But I am facing problems in how to incorporate it in octave. Since the function is not present in sympy how I can I use it ?
 Thanks

cbm755:
Sounds great so far.

If you look in `inst/@sym/taylor`, near the end there is a `cmd = { ' ... ' };` bit.  This is python code, you can probably do something similar.  Fork this project, start a new branch, then send a pull request--- even if its not perfect---we'll help!

trueutkarsh:
cool that should do the job..
as soon as I'll incorporate it in octave will send a pull request.

trueutkarsh:
@cbm755  Now that expr=f(x,y) my code needs a list [x,y] as parameter .
"(f, x, a, n) = _ins' " interprets f as expr  ,I want it to interpret x as list of variables and 'a' as list of evaluation points . 
Right now it throws
 "Python exception: TypeError: 'Zero' object does not support indexing
    occurred at line 8 of the Python code block
"
line 8 is
'expr=f'
'var=x'
'i=a'
'expn=expr.subs([(var[x],i[x]) for x in range(len(var))]) '  %this is line 8

cbm755:
Bit hard for me to comment without seeing your whole code...  I suggest you start the pull request, even if it doesn't work yet (then I can see the actual code and even check it out).

trueutkarsh:
https://github.com/cbm755/octsympy/pull/364

cbm755:
Not at the moment, no.  I should probably get on IRC.

trueutkarsh:
@cbm755  I am facing a new problem. the follwing code snippet does'nt seems to work 

  cmd ={'(f, x, a, n) = _ins'
        'z=f'
        'if not isinstance(x,list):'
        '    z=(z.series(x,a,n).removeO())'     
        'var=list(x)'
        'for m in range(len(var)):'
        '    z=z.series(var[m],a[m],3).removeO()'
        'return nsimplify(z)'      
  };
[...]

PR #560, opened 2016-09-19
We like Taylor Series: add multi-dim expansions

latot:
Hi again, now we can have the taylor series in multivariable (with all vars you want).

Fixes https://github.com/cbm755/octsympy/issues/31

Cya.

### DIFF
--- a/inst/@sym/taylor.m
+++ b/inst/@sym/taylor.m
@@ -120,10 +120,6 @@ function s = taylor(f, varargin)
     end
   end
 
-  if isa(n, 'sym')
-    n = double(n);
-  end
-
   if (isfloat(n))
     n = int32(n);
   end
--- a/inst/@sym/taylor.m
+++ b/inst/@sym/taylor.m
@@ -293,3 +293,7 @@ end
 %! g = sin(x*y*z);		
 %! h = taylor (g, [x,y,z], 'order', 4);		
 %! assert (isAlways(h == x*y*z)) ;
+
+%!error <length>
+%! syms x y
+%! taylor(0, [x, y], [1, 2, 3]);
--- a/inst/@sym/taylor.m
+++ b/inst/@sym/taylor.m
@@ -89,6 +89,9 @@ function s = taylor(f, varargin)
     % taylor(f, [x,y], [a,b], 'param', ...)
     x = varargin{1};
     a = varargin{2};
+    if length(a) ~= length(x) && length(a) == 1
+          a = a*ones(1, length(x));    
+    end
     i = 3;
   elseif (~ischar(varargin{1}) && ischar(varargin{2}))
     % taylor(f, x, 'param', ...)
--- a/inst/@sym/taylor.m
+++ b/inst/@sym/taylor.m
@@ -117,28 +117,40 @@ function s = taylor(f, varargin)
     end
   end
 
+  if isa(n, 'sym')
+    n = double(n);
+  end
+
   if (isfloat(n))
     n = int32(n);
   end
+
+  assert( isequal( length(x), length(a)), 'The length of the expansion point must be the same as the input variables.')
 
   if (numel(x) == 1)
     cmd = { '(f, x, a, n) = _ins'
         's = f.series(x, a, n).removeO()'
         'return s,' };
-  elseif (numel(x) == 2)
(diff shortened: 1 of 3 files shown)


inst/@sym/taylor.m
```diff
@@ -266,9 +278,3 @@
 %! g = x^2 + 2*x + 3;
 %! h = taylor (g, x, 4, 'order', 1);
 %! assert (isequal (h, 27))
-
-%!error<Only 2d expansions supported.>
```
cbm755: can make it into a test instead of removing it?

cbm755:
Can you add a test or two to verify it works?

latot:
okidokis, more tests!, the results match with wolframalpha (but sympy its very slow compared with it)

latot:
only to be sure i compare the examples of SMT with this and are equals (the only difference is here are simplified).

latot:
i feel i'm giving to you a lot of work with all this pr...

inst/@sym/taylor.m
```diff
@@ -89,6 +89,9 @@
     % taylor(f, [x,y], [a,b], 'param', ...)
     x = varargin{1};
     a = varargin{2};
+    if length(a) ~= length(x) && length(a) == 1
```
cbm755: this happens if `length(x) == `1 and `length(a) ~= 1`...?

Should it be: `if (length(a) == 1) && (length(x) > 1)`?

inst/@sym/taylor.m
```diff
@@ -117,28 +120,40 @@
...
 
+  if isa(n, 'sym')
+    n = double(n);
+  end
+
   if (isfloat(n))
     n = int32(n);
   end
 
+  assert( isequal( length(x), length(a)), 'The length of the expansion point must be the same as the input variables.')
```
cbm755: this should have a BIST like `%!error <length> taylor(...)`

inst/private/python_header.py
```diff
@@ -44,6 +44,7 @@ def echo_exception_stdout(mystr):
     import codecs
     import xml.etree.ElementTree as ET
     from distutils.version import LooseVersion
+    import itertools
```
cbm755: I think this will have to go in `python_ipc_pytave.m` as well.

inst/@sym/taylor.m
latot: mm half half, this should be an error:
```taylor(0, [x, y], [1, 2, 3])```
sadly don't are equals.

**Comment by latot** [comment]
This code is more slower compared with the actual 2D, because the actual is restricted to this:
```
     2                                   2          
    ∂                                  ∂            
  ─────(f(x, y))    ==     ─────(f(x, y))
  ∂y ∂x                            ∂x ∂y  
```
but this only is true when we have continuous diffs, and in some discontinuous diff, this code don't are restricted with this.
Actually are ways to have better performance to this, i put a ```FIXME``` for the moment, i think for at least have the base for multivar should be fine.

Code Updated!

File: inst/@sym/taylor.m [hunk 5]
```diff
@@ -117,28 +120,40 @@
     end
   end
 
+  if isa(n, 'sym')
+    n = double(n);
+  end
+
```
cbm755: Why is this important?

cbm755:
@genuinelucifer does this look ok?  In particular, can you sign off on the pytave change?

inst/@sym/taylor.m
latot: n can't have symbols (x, y, z.....), maybe use findsymbols? or remove it and lets python handle an error?

latot: mm, lets left this with the python error, this can work in the future:
```
syms x n
taylor(x, 'order', n)
```

latot:
Okidokis!

genuinelucifer:
+1 from me. The changes look good (after the sym to double removal)...

latot:
Yay! Thx.

bernamdc:
I'm not sure how github works. But I did not find a working multivariate Taylor expansion so I developed one. You can find it in my answer in:

[https://stackoverflow.com/questions/22857162/multivariate-taylor-approximation-in-sympy/63850672#63850672]

Hope it helps for something!

Issue #549, opened 2016-09-14
Implement `@sym/class` to return the Python class


cbm755:
In PyTave and Java `class(x)` gives the Python or Java class.  That might be nice for us too.

E.g.,
````
a = interval(1, 5);
class(a)
  ans = sympy.Interval 
````
For now, it would be a string.  Later, a pyobject...

No pressing reason to do this now, but maybe its fun to think about.

PR #561, opened 2016-09-20
[wip] Implement @sym/class to return python class name of @sym vars

genuinelucifer:
fixes issue #549

Currently it does not follow octave's builtin `class` symantics. I have kept a `full` flag just for discussion.

### DIFF
--- a/inst/@sym/class.m
+++ b/inst/@sym/class.m
@@ -1,0 +1,80 @@
+%% Copyright (C) 2016 Abhinav Tripathi and Colin B. Macdonald
+%%
+%% This file is part of Octave-Symbolic-SymPy
+%%
+%% Octave-Symbolic-SymPy is free software; you can redistribute
+%% it and/or modify it under the terms of the GNU General Public
+%% License as published by the Free Software Foundation;
+%% either version 3 of the License, or (at your option) any
+%% later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @deftypemethod  @var{c} = class (@var{x})
+%% @deftypemethodx @var{c} = class (@var{x}, @var{full})
+%% Return class name of the variable x.
+%%
+%% @var{full} decides if the fully qualified class name will be returned
+%% or not. It it true by default. 
+%%
+%% Example:
+%% @example
+%% @group
+%% syms x
+%% class(x)
+%%   @result{} ans = sympy.core.symbol.Symbol
+%% @end group
+%% @end example
+%%
+%% @example
+%% @group
+%% syms x
+%% class(x, false)
+%%   @result{} ans = Symbol
+%% @end group
+%% @end example
+%%
+%% @end deftypemethod
+
+function cname = class(x, full)
+
+  if (nargin > 2 || (nargin == 2 && !islogical(full)))
+    print_usage ();
+  end
+  
+  if (nargin == 1)
+    full = true;
+  end
+
+  % TODO: if the ipc is pytave then return @pyobject instead
+  cmd = { '(x,f) = _ins'
+          'return (x.__module__ + "." if f else "") + x.__class__.__name__'
+        };
+
+  cname = python_cmd (cmd, x, full);
+end
+
+
+%!error <Invalid> class (sym(1), true, 3)
+%!error <Invalid> class (sym(1), 2)
+
+%!test
+%! syms x y
+%! assert (class(x), 'sympy.core.symbol.Symbol')
+%! assert (class(x), class(x, true))
+%! assert (class(x), class(y))
+
+%!assert (class (sym (1)), 'sympy.core.numbers.One')
+%!assert (class (sym (1), false), 'One')
+
+%!assert (class (sym (1000)), 'sympy.core.numbers.Integer')
+%!assert (class (cos (sym (1))), 'sympy.functions.elementary.trigonometric.cos')


genuinelucifer:
if breaks most of the doctests currently. I'll update the doctests once this is finalised.

inst/@sym/class.m
```diff
@@ -0,0 +1,80 @@
...
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @deftypemethod  @var{c} = class (@var{x})
+%% @deftypemethodx @var{c} = class (@var{x}, @var{full})
+%% Return class name of the variable x.
+%%
+%% @var{full} decides if the fully qualified class name will be returned
+%% or not. It it true by default.
```
latot: very very little things, i think its "It its" or "Its"

latot:
Hi, this is very nice!
it seems works fine with syms, i think left symfuns, you will need improve this to can set class (values and all that things):
```
>> syms g(t)
error: Invalid call to class.  Correct usage is:

 -- Method on C: = class (X)
 -- Method on C: = class (X, FULL)
error: called from
    print_usage at line 90 column 5
    class at line 51 column 5
    symfun at line 184 column 5
    syms at line 176 column 10
```
from symfun.m line 184:
```
  f = class(f, 'symfun', expr);
```
sadly we can't use the builtin function in symfun, octave don't likes that, don't recognizes the call as the type (@symfun/symfun):
```
>> syms g(t)
error: class: 'symfun' is invalid as a class name in this context
error: called from
    symfun at line 184 column 5
    syms at line 176 column 10
```
its very probable this can be a octave bug...

Thx. Cya.

inst/@sym/class.m
```diff
@@ -0,0 +1,80 @@
...
+
+  cname = python_cmd (cmd, x, full);
+end
+
+
+%!error <Invalid> class (sym(1), true, 3)
+%!error <Invalid> class (sym(1), 2)
+
+%!test
+%! syms x y
```
latot: Hi its possible write instead of this line:
```
syms f(x, y)
```
?, to test sym and symfun.

inst/@sym/class.m
genuinelucifer: Oops.. Thanks for pointing out, I'll fix the typo.

genuinelucifer:
>  i think left symfuns, you will need improve this to can set class (values and all that things):
> i think its very normal use 0 and 1 as logical values, only to enable that.

Hi @latot . Sure I'll incorporate these changes.

inst/@sym/class.m
genuinelucifer: ohk. Or maybe we can have additional tests for symfuncs... This way, just to be sure, both can be tested separately. Will add a few more tests..

genuinelucifer:
I just wanted to note that I tried a few different things but as I am not able to find a decent way to make it work for both syms and symfuns...
`symfunc` calls `class` at one point in the constructor and if I make a `@sym/class.m` then that is called. Which causes the creation of symfunc to fail. Also, in python I am getting a `dict` when I pass the symfunc! I probably have lost track of how symfuncs are being created... Moreover if I return any string from class say using:
```
if (isa (x, 'symfunc'))
   cname = 'symfunc';
end
```
then if I try `syms g(t)` then `g` is actually the string `symfunc`!!
Also, as @latot mentioned, calling `builtin('class',...` also seems to be not working!
.
This probably would be a lot easier to fix on `pytave` IPC. If I don't get a way around this in windows, I will update the file to only work with `pytave` IPC and directly return the pyobject instead (which I hope should be fairly straightforward)...

cbm755:
1.  Does implementing `@symfun/class` help?  (I guess not)
2.  The ctor call to `class` may not be with classdef...  Maybe we can do this class override after we switch to classdef?
3.  We should take the `f = builtin('class', mystruct, 'symfun', myparentobject)` thing to upstream Octave.  Can you explain it or do you want me to file it?

latot:
I already report it, if you can complement the explanation it would be great:

https://savannah.gnu.org/bugs/?49169

Thx. Cya.

latot:
The its can overload class only to return a value, don't to set (when we will set we call builtin).

mtmiller:
> https://savannah.gnu.org/bugs/?49169

Commented upstream, hopefully we can get some information on what Matlab does in this situation.

> Does implementing @symfun/class help?

No, the error is in overload resolution and the extra arguments passed when doing inheritance with old-style class-based classes.

> The ctor call to class may not be with classdef

Right, classdef would work around this.

latot:
mm, personally i think this ```@sym/class``` helps a lot, i normally works with a mix of things, syms, symfuns, intervals and matrix, and when i display the info its too hard know what are matrix and what are intervals(when i have a lot of data)....

genuinelucifer:
> Right, classdef would work around this.

I will then try to do the classdef implementation now. If that works then we could implement this on top of that.

latot:
@genuinelucifer and @cbm755 a little question, i don't have very clear what are differences between the actual sym and implement classdef (i try search but i can get a clear answer), ex. why class should works from classdef and not now? or what the differences between the concepts vs functionalities?

The mains differences i have clear are, you can set methods from classdef (but we can do this too from syms in other files), in classdef you can set properties to the objects (in some way sym do this with a struct), the extras of classdef are the events and enumeration but this last two don't are available in octave (and i was start thinking how use it :D).

This can approached from other way, edit the display function and add the actual class code there.

Thx. Cya.

cbm755:
I'm not clear on differences either.  I think of them as two different approaches to OO, or at least two different implementations.  And so there are some differences between them, different bugs too ;-)

Classdef-style classes allow more OO features like static methods.

cbm755:
@latot where you working on the classdef?  @genuinelucifer if you start too, let's get a "WIP" PR up ASAP for collaboration and not duplicating effort.

latot:
I'm not working on it now, for can help i need first understand the concept.

Thx. Cya.

cbm755:
Thanks @latot: your review if @genuinelucifer does it would be very much appreciated (and good chance to learn).

genuinelucifer:
> if you start too, let's get a "WIP" PR up ASAP for collaboration and not duplicating effort.

Definitely. I am compiling octave on my linux machine. I can probably send a WIP PR by tomorrow.
I also have to go through classdefs once.

PR #570, opened 2016-09-22
add remove row/column

latot:
Fixes https://github.com/cbm755/octsympy/issues/569

I don't know how we skip this D:

### DIFF
--- a/inst/@sym/subsasgn.m
+++ b/inst/@sym/subsasgn.m
@@ -547,7 +547,7 @@ end
 %! a(1) = [];
 %! assert( isequal( a, b));
 
-%!error
+%!error <null assignment>
 %! a = sym([1, 2; 3, 4]);
 %! a(1, 2) = [];
 
--- a/inst/@sym/subsasgn.m
+++ b/inst/@sym/subsasgn.m
@@ -95,7 +95,42 @@ function out = subsasgn (val, idx, rhs)
                   'invalid indices: should be integers or boolean');
           end
 	end
-        out = mat_replace(val, idx.subs, sym(rhs));
+        
+        if isempty(rhs)
+
+          switch length(idx.subs)
+            case 1
+              if strcmp(idx.subs{1}, ':')
+                out = sym([]);
+              else
+                if rows(val) == 1
+                  out = python_cmd('_ins[0].col_del(_ins[1] - 1); return _ins[0],', val, sym(idx.subs{1}));
+                elseif columns(val) == 1          
+                  out = python_cmd('_ins[0].row_del(_ins[1] - 1); return _ins[0],', val, sym(idx.subs{1}));         
+                else
+                  out = sym([]);
+                  for i=1:val.size(2)
+                    out = [out subsref(val, substruct ('()', {':', i})).'];
+                  end
+                  out = subsasgn (out, substruct ('()', {idx.subs{1}}), []);
+                end
+              end
+            case 2
+              if strcmp(idx.subs{1}, ':')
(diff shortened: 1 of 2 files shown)


latot:
Still left 3 cases to this works, let me think how fix it.

latot:
Okidokis, Full Fixed!

inst/@sym/subsasgn.m
```diff
@@ -95,7 +95,42 @@
...
+                  out = subsasgn (out, substruct ('()', {idx.subs{1}}), []);
+                end
+              end
+            case 2
+              if strcmp(idx.subs{1}, ':')
+                out = python_cmd('_ins[0].col_del(_ins[1] - 1); return _ins[0],', val, sym(idx.subs{2}));
+              elseif strcmp(idx.subs{2}, ':')
+                out = python_cmd('_ins[0].row_del(_ins[1] - 1); return _ins[0],', val, sym(idx.subs{1}));
+              else
+                error('A null assignment can only have one non-colon index.'); %% Standard octave error
```
cbm755: does this error have BIST?

inst/@sym/subsasgn.m
```diff
@@ -464,4 +499,56 @@
...
+%! assert( isequal( a, sym([])))
+
+%!test
+%! % Dissamble Matrix
+%! a = sym([1, 2; 3, 4; 5 6]);
+%! b = sym([3, 5, 2, 4, 6]);
+%! a(1) = [];
+%! assert( isequal( a, b));
+
+%!error
```
cbm755: include some `<keyword>` for the error?

inst/@sym/subsasgn.m
latot: sorry my ignorance, what is BIST?

cbm755:
Tests look great and we definitely need a fix.  Fix feels long though...

Can it be made inside `mat_replace`?  Or maybe even `mat_rclist_asgn` and `mat_mask_asgn` need to deal with the empty `b` case?  I could be wrong: its so long since I wrote all that spaghetti code!

But I think the intention was the private helpers would do the actual work.

inst/@sym/subsasgn.m
cbm755: built in self test.  I just mean `%! error <null assignment> ...`

cbm755: oops `%!error <null assignment> ...`

inst/@sym/subsasgn.m
latot: okis, added

latot:
Okis, let me see how this parts works to mix all this things.

latot:
Okidokis, after think this, i move the code to ```mat_replace.m```, the ```mat_mask_asgn``` and ```mat_rclist_asgn```  works with assign of values case by case but ```[]``` its the special case of remove instead of set, so now the code start checking if b is empty, if not continues normally.

Code Updated!

latot:
@cbm755 a little question about Sympy, well you know i'm not an expert of python, but i want request to Sympy the solver for inequalities, actually its very incomplete, can't works with more than one var, it would be useful if i request the feature and in the issue i explain the logic of an inequality solver?

cbm755:
I'm not an expert on Sympy solvers.  There is/was a lot of work going on to improve solvers; I think called something like "solveset"/setsolve.  Search SymPy issues to learn more I guess?

latot:
Oh, sadly there are no issues about inequalities, only about eqs, (i search "ineq" in sympy) and i check the section you recommend D:
i think i'll write it, Thx.

Issue #576, opened 2016-09-23
Add `equationsToMatrix` command


cbm755:
Someone asked on the maillist.  Looks like its:

https://www.mathworks.com/help/symbolic/equationstomatrix.html

latot:
:O nice, i think i have an idea how do this!

latot:
mm, to do this we need more info, for the moment my doubt its this:
```
syms x y
equationsToMatrix(x, y, y)  %%What are eqs and what are vars?
equationsToMatrix(x, y)  %%Same, i can suppose x is eqs and y vars.
equationsToMatrix(x)  %%What this should return?
```

latot:
okis, that doubts was explained in the doc in matlab, so continues.

cbm755:
This octave-help post http://lists.gnu.org/archive/html/help-octave/2016-09/msg00092.html provides includes an example.

cbm755:
`sympy.solveset.linear_eq_to_matrix` is mentioned [here](https://groups.google.com/forum/#!topic/sympy/Wqs1OhTBexg), haven't tried it.

latot:
:O lets check it, (i try search but i can't found D:)

latot:
mmmmmmmmmmmmm, this seems very useful, don't is documented nowhere, but we can check the source directly: https://github.com/sympy/sympy/blob/master/sympy/solvers/solveset.py#L989
The only problem with this is it don't detect non linear systems (this is for SMT compat), but if we check it manually like in the pr it would be more expensive call ```linear_eq_to_matrix```, because we can make the solution in the recursion of the check.

So the question is, keep that feature of SMT?

Only to be clear, this should send an error:
```
equationsToMatrix(x*x, x)
```

Thx.

cbm755:
Oh, I'm sorry, I hadn't seen your PR when I sent that link.

I agree nonlinear equations should be an error.  It scares me if the sympy code does *not* check that...

PR #577, opened 2016-09-24
equationsToMatrix

latot:
Fixes https://github.com/cbm755/octsympy/issues/576

Hi, well again a new function, this still i need add the doc and tests, add info in the pytave section, etc, so don't check it for the moment plis.

Edited, I found the problem, fixing.

Thx. Cya.

### DIFF
--- a/inst/@sym/equationsToMatrix.m
+++ b/inst/@sym/equationsToMatrix.m
@@ -120,51 +120,49 @@ function [A b] = equationsToMatrix(varargin)
 end
 
 
-%%Matlab Tests
-
 %!test
 %! syms x y z
-%! [A, B] = equationsToMatrix ([x + y - 2*z == 0, x + y + z == 1, 2*y - z + 5 == 0], [x, y, z]);
-%! a = sym ([1 1 -2; 1 1 1; 0 2 -1]);
-%! b = sym ([0; 1; -5]);
+%! [A, B] = equationsToMatrix ([x + y - z == 1, 3*x - 2*y + z == 3, 4*x - 2*y + z + 9 == 0], [x, y, z]);
+%! a = sym ([1 1 -1; 3 -2 1; 4 -2 1]);
+%! b = sym ([1; 3; -9]);
 %! assert (isequal (A, a))
 %! assert (isequal (B, b))
 
 %!test
 %! syms x y z
-%! A = equationsToMatrix ([x + y - 2*z == 0, x + y + z == 1, 2*y - z + 5 == 0], [x, y, z]);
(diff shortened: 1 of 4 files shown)


inst/@sym/equationsToMatrix.m
```diff
@@ -0,0 +1,110 @@
...
+function [A b] = equationsToMatrix(varargin)
+
+  s = symvar([varargin{:}]);
+
+  cmd = {'eqns = list(_ins); vars = list()'
+         'if not _ins[-1].is_Matrix:'
+         '    for i in reversed(xrange(len(_ins))):'
+         '        if _ins[i].is_Symbol:'
+         '            vars.append(_ins[i])'
+         '            del _ins[-1]'
```
latot: This line don't works.

inst/@sym/equationsToMatrix.m
```diff
@@ -0,0 +1,110 @@
...
+         '        else:'
+         '            break'
+         'else:'
+         '    for i in _ins[-1]:'
+         '        if i.is_Symbol:'
+         '            vars.append(i)'
+         '    del _ins[-1]'
```
latot: Same.

latot:
Okidokis, Now works!

latot:
Code ready and formatted!

inst/@sym/equationsToMatrix.m
```diff
@@ -0,0 +1,173 @@
...
+
+
+  if ~s
+    error('Cannot convert to matrix form because the system does not seem to be linear.');
+  end
+
+end
+
+
+%%Matlab Tests
```
cbm755: We need to come up with our own tests.  Or we could take from a text book (which we cite).

latot: we can't use directly the matlab tests?

cbm755: Maybe its a grey area, but I don't think its a good idea.

And just a reminder: reading their docs is ok, but we should *never* look inside their m-files.

latot: xD i'm agree, that its a very bad idea.

but don't is impossible some codes be similars... (its a hard point)
(i don't have matlab)

latot:
ready!
why i need search in books? octave do the work!
```
a=[];
for i=1:3
  for j=1:4
    a(i, j) = (-1)^fix(rand*10)*fix(rand*10);
  end
end
```

cbm755:
Ok, I reviewed this and sent a PR to you.  Further changes could probably be done post-merge.

Note to self: rebase this before merging for clean history (because of examples noted above).

cbm755:
Merged as #684.

PR #578, opened 2016-09-24
rename xrange to range

latot:
xrange was removed in Python 3, and in the actual Python 2.7 supports xrange format in range.


cbm755:
Ok, I can merge this...

But shouldn't some tests have been failing because of this?

latot:
mm, should be, but seems depends to much of the versions of Python, my Python 3 send errors (3.4.3).
So seems some Python 3 have xrange and others not, but in python 3 and 2 range support xrange, su i think its safer change it.

cbm755:
Well, it certainly failed tests for me locally!  Something wrong with the CI setup, filed #586.

PR #580, opened 2016-09-25
Refactor of Solve, and fix a test

latot:
Hi, you know, the continuation of the other pr, idea?, compact solve and pass this test:

```
%!test
%! syms x
%! [a b] = solve(x^2+x == 0)
```

i think this should be:

```
%!error
%! syms x
%! [a b] = solve(x^2+x == 0)
```

and should be handled for solve instead of octave

Thx.Cya.

### DIFF
--- a/inst/@sym/solve.m
+++ b/inst/@sym/solve.m
@@ -31,9 +31,9 @@
 %% @example
 %% @group
 %% syms x
-%% solve(x == 2*x + 6, x)
+%% solve (x == 2*x + 6, x)
 %%   @result{} ans = (sym) -6
-%% solve(x^2 + 6 == 5*x, x)
+%% solve (x^2 + 6 == 5*x, x)
 %%   @result{} ans = (sym 2×1 matrix)
 %%       ⎡2⎤
 %%       ⎢ ⎥
@@ -73,7 +73,7 @@
 %% Alternatively:
 %% @example
 %% @group
-%% [X, Y] = solve(x^2 == 4, x + y == 1, x, y)
+%% [X, Y] = solve (x^2 == 4, x + y == 1, x, y)
 %%   @result{} X = (sym 2×1 matrix)
 %%       ⎡-2⎤
 %%       ⎢  ⎥
@@ -89,12 +89,11 @@
 %% inequalities and equations.  For example:
 %% @example
 %% @group
-%% solve(x^2 == 4, x > 0)
+%% solve (x^2 == 4, x > 0)
 %%   @result{} ans = (sym) x = 2
 %% @end group
-%%
 %% @group
-%% solve(x^2 - 1 > 0, x < 10)
+%% solve (x^2 - 1 > 0, x < 10)
 %%   @result{} ans = (sym) (-∞ < x ∧ x < -1) ∨ (1 < x ∧ x < 10)
 %% @end group
 %% @end example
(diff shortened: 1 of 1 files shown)


cbm755:
Oops, sorry, I forgot about this and made a couple extra commits.  Now there are conflicts.

latot:
don't worry, i'll fix it later, thx!

latot:
Code ready and formatted!

inst/@sym/solve.m
```diff
@@ -31,9 +31,9 @@
 %% @example
 %% @group
 %% syms x
-%% solve(x == 2*x + 6, x)
+%% solve (x == 2 * x + 6, x)
```
cbm755: no spaces around `*` and `^`

latot: okis, to know, what operators need spaces and what not?
https://www.gnu.org/software/octave/doc/v4.0.3/Operator-Index.html

latot:
ready!

inst/@sym/solve.m
cbm755: I don't know, just follow the conventions in the surrounding code.

latot: okis, the code its updated.
i'm asking because you mentioned you want follow some code style, and your old code don't follow it..
Thx.

cbm755:
@latot.  My opinion:

1.  After your recent changes, code is much simpler.  So IMHO, this refactor makes it more complicated.  Let's *not* factor out the common bit of code ok?

2.  If I understand correctly, the other change is to replace a warning with an error.  I still don't see why that is better.  If you think its helpful to report numbers, change the current warning?

3.  Reformatting stuff LGTM!

latot:
mm, okis, the warning issue is because the error is handled via octave instead of solve, and the error don't is to much clear, basically actually we have warning -> error(low info), i think its better have the error and more info.

i still think its better this refactor only for one reason, i don't like the idea of start writing the same ideas in two differents spaces, like the actual python_header and pytave, or in this case have the same code in all lines except in the last 5...

Well, i need say the first impression of the code its more complicated, but works with it i think its more simple, and less probable forget some change in one place or other.

Thx. Cya.

cbm755:
You're certainly right about the code duplication but the strrep is confusing too...

I don't understand the error/warning thing.  Can you paste in an example showing what the currently happens so I can see what is wrong?

latot:
okidokis, simpler way:
```
octave:30> syms x y
octave:31> [A B C]=solve(x==y, x+1==y)
warning: solve: number of outputs did not match solution vars
warning: called from
    solve at line 146 column 7
error: element number 1 undefined in return list
```
When this cap happen? when we work with a math workflow, we normally don't control what expressions exactly enter to the system, or thinking  in the same instead of an assert, send the warning when ```sols > length(out)```, and in the inverse case ```sols < length(out)``` cover it nan results.
What do you think?

thinking in other solution, i can't found one from python, because its use strrep or eval, but strrep is confusing and you want avoid eval, so thinking i will use a helper function to do this more easy, i think will be better, then you will can say if you like or not, if not i'll try search other sol.

Thx. Cya.

latot:
or we can check what does SMT in this case.

cbm755:
pretty sure SMT also gives an error when number of outputs doesn't match what user asked for.

PR #583, opened 2016-09-30
little fixes for Sets - SolveSet and Display

latot:
Hi, this is a fix to can work properly with this, i don't report this problems but anyway here are the fixes:

first for Display, `fprintf` don't support the backslash char and send errors:

```
octave:3> python_cmd('return S.UniversalSet - S.Reals')
warning: called from
    python_cmd at line 188 column 5
warning: unrecognized escape sequence '\ ' -- converting to ' '
warning: called from
    display at line 135 column 3
ans = (sym) UniversalSet()  ℝ
```

with the fix:

```
octave:20> python_cmd('return S.UniversalSet - S.Reals')
warning: called from
ans = (sym) UniversalSet() \ ℝ
```

how we only call display when we show info this its very very very unlikely effects the performance.

The second problem, when we call some set sadly srepr don't use the same expression to show it, so for can access it from out python env we need add the sets modules.

```
octave:4> a=python_cmd('return S.Naturals')
a = (sym) ℕ
octave:5> python_cmd('return S.UniversalSet - _ins[0]', a)
  File "<stdin>", line 4, in <module>
NameError: name 'Naturals' is not defined
error: Python exception: NameError: name 'Naturals' is not defined
    occurred while copying vars to Python
error: called from
    python_cmd at line 178 column 5
```

checking the format:

```
octave:5> char(a)
ans = Naturals()
```

Why solveset? solveset returns sets depending of the domain, so when we will eval some result with it in octave we will get the same error.

### DIFF
--- a/inst/@sym/display.m
+++ b/inst/@sym/display.m
@@ -136,11 +136,13 @@ function display(x)
   s = [s1 s2];
   n = ustr_length (s);
   %fputs (1, s);  % only in octave, not matlab
-  disp (s)
+
   if (display_snippet)
     % again, fputs safer, but not in matlab
-    fprintf (snippet_of_sympy (x, 7, term_width - n, unicode_dec))
+    s = [s snippet_of_sympy(x, 7, term_width - n, unicode_dec)];
   end
+
+  disp (s)
 
   if (toobig)
     if (loose), fprintf ('\n'); end
--- a/inst/@sym/display.m
+++ b/inst/@sym/display.m
@@ -135,10 +135,8 @@ function display(x)
   end
   s = [s1 s2];
   n = ustr_length (s);
-  %fputs (1, s);  % only in octave, not matlab
 
   if (display_snippet)
-    % again, fputs safer, but not in matlab
     s = [s snippet_of_sympy(x, 7, term_width - n, unicode_dec)];
   end
 
--- a/inst/@sym/display.m
+++ b/inst/@sym/display.m
@@ -136,12 +136,11 @@ function display(x)
   s = [s1 s2];
   n = ustr_length (s);
   %fputs (1, s);  % only in octave, not matlab
-  fprintf (s)
+  disp (s)
   if (display_snippet)
     % again, fputs safer, but not in matlab
     fprintf (snippet_of_sympy (x, 7, term_width - n, unicode_dec))
   end
-  fprintf ('\n');
 
   if (toobig)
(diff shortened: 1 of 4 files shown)


inst/@sym/display.m
```diff
@@ -136,12 +136,11 @@ function display(x)
   s = [s1 s2];
   n = ustr_length (s);
   %fputs (1, s);  % only in octave, not matlab
-  fprintf (s)
+  disp (s)
```
cbm755: this will break the display of "snippets" because `disp` always puts a newline.  Instead, do the display after the `if (display_snippet)`...

inst/@sym/display.m
```diff
@@ -136,12 +136,11 @@ function display(x)
...
   if (display_snippet)
     % again, fputs safer, but not in matlab
     fprintf (snippet_of_sympy (x, 7, term_width - n, unicode_dec))
```
cbm755: here `s = [s snippet_of_sympy (...)];`

inst/@sym/display.m
```diff
@@ -136,12 +136,11 @@ function display(x)
...
   end
-  fprintf ('\n');
```
cbm755: finally, `disp(s)`

inst/private/python_ipc_native.m
```diff
@@ -65,6 +65,8 @@
                     'import sympy as sp'
                     'from sympy import __version__ as spver'
                     'from sympy import *'
+                    'from sympy.sets.fancysets import *'
+                    'from sympy.sets.sets import *'
```
cbm755: maybe do this below the hypergeometric ones and leave a comment like `# for sets`.

latot:
Fixed!

inst/@sym/display.m
```diff
@@ -136,12 +136,13 @@ function display(x)
   s = [s1 s2];
   n = ustr_length (s);
   %fputs (1, s);  % only in octave, not matlab
```
cbm755: better remove this comment now, as we don't use printf

inst/@sym/display.m
```diff
@@ -136,12 +136,13 @@ function display(x)
...
-  fprintf (s)
+
   if (display_snippet)
     % again, fputs safer, but not in matlab
```
cbm755: same, remove comment

latot:
Hi, ready!

Issue #575, opened 2016-09-23
More functions to Sets (What names we will use to it?)


latot:
Hi, well, again with sets, this time i want write the proper functions to have it more functional, but i don't know exactly what names use in all cases, so here i'll list functions of it, and we can discuss if some function are necessary or not (its very probable i'll write this and then we can choose the names for it):

Reference: http://docs.sympy.org/dev/modules/sets.html

- boundary
- complement
- contains
- inf
- is_disjoint
- is_proper_subset
- is_proper_superset
- is_subset
- is_superset
- measure
- powerset
- sup
- imageset
- is_left_unbounded
- is_right_unbounded
- left_open
- right_open
- is_Complement
- is_EmptySet  /This can be in 'isempty' function
- is_Intersection
- is_UniversalSet
- reduce   /This can be added to simplify
- EmptySet  /This can be added to sym
- UniversalSet    /This can be added to sym
- Naturals    /This can be added to sym
- Naturals0    /This can be added to sym
- Integers    /This can be added to sym
- Range
- reversed (only in range)
- ComplexRegion
- a_interval
- b_interval
- polar
- psets
- sets
- normalize_theta_set

Left update this docs of operators:
'-' in sets that means the difference of both.
'+' union of two intervals
'*' its the ProductSet of two intervals

maybe for '-' and '*' make two extra functions only to explicit this?

Before send a lots of pr with this, i think i'll wait to finish the current prs.

Thx. Cya.

PR #585, opened 2016-09-30
[WIP] You like sets? I like sets! Full sets Module!

latot:
Hi @cbm755 this time i bring the full sets module (i skip some redundant functions), because we add intervals function with some things but is very incomplete, when i need works with it, it feels... without tools to handle it.

So i think with this we will can works properly with sets.

Obvs, the function names don't are absolutes, if you like we can discuss names for it, in the most of functions i use the sympy name, in others i change it to conserve the octave/matlab prefix.

And this time i take time to check every format!

This pr is required for improve solveset correctly.

Here are two special functions it not belongs directly from sympy sets:
  * domain: return a domain from sympy 'Complexes', 'Reals', etc (the only function added to user env)
  * point: express a point or a set of it (don't is from sets but its a feature of sympy).

Obvs you can like split this heavy pr, but i think before it select ready things and that type of things.
I think here no are unnecessary functions, things with which we can work.

The original issue: https://github.com/cbm755/octsympy/issues/575
have the original names of sympy functions and some data of this work.

i don't know if all this functions are available in the travis sympy, i'll check it.

Thx. Cya.


cbm755:
Wow!  This is pretty cool.

Pytave might make at least some of this possible without us writing code (by exposing things like `.openleft` automatically).

cbm755:
looks like it has run now (?)

latot:
Yay!, all updated and working!

latot:
Hi, is possible start merging this?

Thx.

inst/@sym/ainterval.m
```diff
@@ -0,0 +1,65 @@
...
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym ainterval (@var{x})
+%% Return the union of intervals of x when, @var{x} is in rectangular form
+%% or the union of intervals of r when self is in polar form.
```
cbm755: First line need to fit on one line (its used for `lookfor`).

inst/@sym/boundary.m
```diff
@@ -0,0 +1,48 @@
...
+    print_usage ();
+  end
+  y = elementwise_op ('lambda x: x.boundary', sym (x));
+end
+
+
+%!test
+%! a = interval (sym (0), 1);
+%! b = interval (sym (0), 1, true, false);
+%! assert( isequal( boundary (a), boundary (b)))
```
cbm755: I know I've been very sloppy in the past, but for new code, let's try to use core Octave style (except for `sym()`).

inst/@sym/complexregion.m
```diff
@@ -0,0 +1,63 @@
...
+    print_usage ();
+  elseif nargin == 1
+    y = false;
+  end
+  y = elementwise_op ('lambda x, y: ComplexRegion(x, polar=y)', sym (x), logical (y));
+end
+
+
+%% This function is tested in @sym/ainterval, @sym/binterval
+%% @sym/ispolar, @sym/psets, @sym/sets and @sym/normalizethetaset
```
cbm755: add an `%! assert true` or something here too so automated checks don't flag this as untested.

inst/@sym/complexregion.m
```diff
@@ -0,0 +1,63 @@
...
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym complexregion (@var{x}, @var{y})
+%% Represents the Set of all Complex Numbers.
+%% It can represent a region of Complex Plane in
```
cbm755: blank line between the "lookfor" and the rest of the body.  But this looks wrong?  Should it represent all complex numbers if no arguments are passed or what does the `lookfor` summary line mean?

inst/@sym/complexregion.m
```diff
@@ -0,0 +1,63 @@
...
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym complexregion (@var{x}, @var{y})
```
cbm755: `y` should be called something else like `@var{polar}`.  And please list both ways of calling it:
````
@defmethod  @@sym complexregion (@var{x})
@defmethodx @@sym complexregion (@var{x}, @var{polar})
````

File: inst/@sym/contains.m [hunk 6]
```diff
@@ -0,0 +1,50 @@
...
+%% @end example
+%%
+%% @end defmethod
+
+
+function y = contains(x, y)
+  if (nargin ~= 2)
+    print_usage ();
+  end
+  y = elementwise_op ('lambda x, y: x in y', sym (x), sym (y));
```
cbm755: Is `x in y` the correct way sympy to test for something being in a set?  What if y is a list?  How is this different from `ismember`?

inst/@sym/imageset.m
```diff
@@ -0,0 +1,82 @@
...
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym imageset (@var{f}, @var{d})
+%% @defmethodx @@sym imageset (@var{f}, @var{x}, @var{d})
+%% Image of a set under a mathematical function.
+%% The imageset is created in the function @var{f} in the variable
+%% @var{x} on the domain @var{d}, if @var{x} don't is passed will
```
cbm755: `if @var{x} is not passed, it will be detected using symvar, @pxref{@@sym/symvar}.`

Or something like that.

inst/@sym/infimum.m
```diff
@@ -0,0 +1,51 @@
...
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym infimum (@var{x})
+%% The infimum of @var{x}.
```
cbm755: `The infimum of the set @var{x}.`

inst/@sym/iscomplement.m
```diff
@@ -0,0 +1,48 @@
...
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym iscomplement (@var{x})
+%% Retrurn True if @var{x} is complement.
```
cbm755: spelling mistake and this lookfor line doesn't tell me any more than the name `iscomplement` already does.  At least put the word "set" in there somewhere.

inst/@sym/iscomplement.m
```diff
@@ -0,0 +1,48 @@
...
+%% Example:
+%% @example
+%% @group
+%% a = domain ('Reals');
+%% b = domain ('Complexes');
+%% iscomplement (b - a)
+%%   @result{} ans = (sym) True
+%% @end group
+%% @end example
+%%
```
cbm755: Some/most of these methods need `@seealso`

inst/@sym/isdisjoint.m
```diff
@@ -0,0 +1,53 @@
...
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym isdisjoint (@var{x}, @var{y})
+%% Returns True if @var{x} and @var{y} are disjoint.
```
cbm755: make sure the word "set" appears in most of these "lookfor" summary lines,

inst/@sym/isleftunbounded.m
```diff
@@ -0,0 +1,58 @@
...
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod @@sym isleftunbounded (@var{x})
+%% Return True if the left endpoint is negative infinity.
```
cbm755: "return True" is probably not needed (I forgot exactly, you could check the core Octave guidelines on this sort of thing).  Use the word "Interval" somewhere.

cbm755:
My biggest concern about all this is whether we should just wait until (and if?) `sym` objects support `x.is_leftopen` calling directly without us writing any code.

That would keeps things as close to sympy as possible, keeping things slim.

OTOH, this is here now and looks like it will be good.  I am undecided what to do.  Which is probably why I've let this sit so long...

Issue #520, opened 2016-09-01
Support mix symfun


latot:
Hi again, well i think this can be good here, sympy actually support it, mm, the truth i was thinking it was already supported, until i read in the syms.m `'mixing symfuns and assumptions not supported'`

What do you think?

A little example:

```
>> q=python_cmd('return S("f(g(t))")')
q = (sym) f(g(t))
>> diff(q, t)
ans = (sym)

  d        ⎛ d        ⎞│       
  ──(g(t))⋅⎜───(f(ξ₁))⎟│       
  dt       ⎝dξ₁       ⎠│ξ₁=g(t)
```

cbm755:
Only way I know to do this is sort of what is described at the end of `help symfun`:
````
octave:39> f = symfun(sym('f(g(t))'), t)
f(t) = (symfun) f(g(t))
octave:40> diff(f)
ans(t) = (symfun)

  d        ⎛ d        ⎞│       
  ──(g(t))⋅⎜───(f(ξ₁))⎟│       
  dt       ⎝dξ₁       ⎠│ξ₁=g(t)
````
(Which works the same as your example.)

**Comment by cbm755** [comment]
This is a bit less "magical":
````
octave:46> syms f(t) g(t)
octave:47> F(t) = f(g(t))
F(t) = (symfun) f(g(t))
octave:48> diff(F,t)
ans(t) = (symfun)

  d        ⎛ d        ⎞│       
  ──(g(t))⋅⎜───(f(ξ₁))⎟│       
  dt       ⎝dξ₁       ⎠│ξ₁=g(t)
````
Maybe we should add this example to `help symfun`?

**Comment by latot** [comment]
mm, i think should be added, and if its possible update syms function:
```
>> syms f(g(t))
error: invalid symfun expression
error: called from
    assert at line 94 column 11
    syms at line 169 column 7
```

**Comment by latot** [comment]
this was already supported but now unavailable, see discussion in https://github.com/cbm755/octsympy/issues/537.

**Comment by latot** [comment]
D: i can't found what exactly code disable this feature.....
there is a mix of things

**Comment by latot** [comment]
mm, updating a little the objetive of this issue, the idea its can do this:
```
syms f(t) g(t) r(f, g)
r(f, g)=f+g
```
actually we get this error:
```
>> r(f, g)=f+g
error: Python exception: TypeError: can't convert expression to float
    occurred at line 6 of the Python code block
error: called from
    python_cmd at line 174 column 5
[...]

PR #588, opened 2016-09-30
[Waiting 577] Mix Symfun 1/2

latot:
Fixes https://github.com/cbm755/octsympy/issues/520

Hi, well finally i can do this works, why 1/2? because we can mix symfuns normally but we can't link it directly (new issue).

First to can finish this i'll need the `flatten` function from this pr: https://github.com/cbm755/octsympy/pull/577

This pr basically have three things:
- Enable Mix symfuns
- Enable Mix symfuns with differents args `f(x, y)+g(t)` (its supported for sympy and works fine)
- Fix Display for the mix.

One problem actually, related but it will requieres more work because seems its a structural problem, but for now this should works.

the problem is when you try mix some things, ex:

```
octave:11> syms h(g) g(h)
octave:12> char(h)
ans = Function('h')(Symbol('g'))
octave:13> char(g)
ans = Function('g')(Symbol('h'))
```

(note in this example i disable the function of syms to re-create the var if exist in the user env)
so this problem its very ugly but is apart, i'll open a new issue for it.

Thx. Cya.

### DIFF
--- a/inst/@sym/subsasgn.m
+++ b/inst/@sym/subsasgn.m
@@ -70,8 +70,12 @@ function out = subsasgn (val, idx, rhs)
         all_syms = all_syms && isa(idx.subs{i}, 'sym');
       end
       if (all_syms)
-        cmd = { 'L, = _ins'
-                'return all([x is not None and x.is_Symbol for x in L])' };
+        % Waint for flatten function from https://github.com/cbm755/octsympy/pull/577
+        cmd = { 'for x in _ins[0]:'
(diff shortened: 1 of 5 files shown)


latot:
mm, i found other problem in octave, if you use this from a function ```isfield(x, 'flat')``` where x is an expression of sym it returns 0.....

cbm755:
> i disable the function of syms to re-create the var if exist in the user env

If you don't want that feature, call `symfun` directly instead of using `syms`...

latot:
okis, but that was only a reference of concepts/behavior, although exist a way to do it, i think should be consistent.

Issue #545, opened 2016-09-07
Make the sym class a new-style classdef class


cbm755:
Possibly this allows fixing #13.  

Good chance to learn more about Octave OO programming.  Maybe @latot is interested?

latot:
Yay! i'm interested, and to start:
https://es.mathworks.com/help/matlab/matlab_oop/getting-familiar-with-classes.html

We must be careful, i don't know if all functions of classdef are available in octave.

cbm755:
You might look at `@pyobject` which is classdef and works properly:

https://bitbucket.org/mtmiller/pytave/src/d0a7f66393fcaeaad789d83c41462ed6a4c353ed/@pyobject/?at=default

mtmiller:
There are some limitations in the Octave implementation of `classdef` to watch out for. The simplest thing to do would be to change `@sym/sym.m` to define the class and constructor only, and all methods defined in separate .m files should still work as-is.

PR #590, opened 2016-10-01
[wip] Change @sym to a classdef.

genuinelucifer:
Fixed issue #545

This is at very early stage. MOST of the tests fail with this change. This is here just for discussion.

### DIFF
--- a/inst/@sym/size.m
+++ b/inst/@sym/size.m
@@ -65,7 +65,7 @@ function [n, m] = size(x, dim)
 
   % Note: symbolic sized matrices should return double, not sym/string.
 
-  n = x.size;
+  n = x.symsize;
 
   % FIXME: for now, we artificially force symbolic sized objects
   % (where one or more dimension is recorded as NaN) to be 1x1.
--- a/inst/@sym/sym.m
+++ b/inst/@sym/sym.m
@@ -158,212 +158,224 @@
 %% Author: Colin B. Macdonald
 %% Keywords: symbolic, symbols, CAS
 
-function s = sym(x, varargin)
+classdef sym < handle
+  properties
+    pickle
+    symsize
+    flat
+    ascii
+    unicode
+    extra
+  endproperties
 
-  if (nargin == 0)
-    s = sym(0);
-    return
-  end
+  methods
+    function s = sym(x, varargin)
 
-  %% The actual class constructor
-  % Tempting to make a 'private constructor' but we need to access
-  % this from the python ipc stuff: outside the class.  We identify
-  % this non-user-facing usage by empty x and 6 inputs total.  Note
-  % that "sym([])" is valid but "sym([], ...)" is otherwise not.
-  if (isempty(x) && (nargin == 6))
-    s.pickle = varargin{1};
-    s.size = varargin{2};
-    s.flat = varargin{3};
-    s.ascii = varargin{4};
-    s.unicode = varargin{5};
-    s.extra = [];
-    s = class(s, 'sym');
-    return
-  end
-
-  %% User interface for defining sym
-  % sym(1), sym('x'), etc.
-
-  %if (strcmp (class (x), 'symfun')  &&  nargin==1)
-  %  % FIXME: pass a symfun to sym() ctor; convert to pure sym
-  %  % (SMT does not do this in 2014a).  bad idea?
-  %  s = x.sym;
-  %  return
-
-  if (isa (x, 'sym')  &&  nargin==1)
-    % matches sym and subclasses
-    s = x;
-    return
-
-  elseif (iscell (x)  &&  nargin==1)
-    s = cell_array_to_sym (x);
-    return
-
-  elseif (isnumeric(x)  &&  ~isscalar (x)  &&  nargin==1)
-    s = numeric_array_to_sym (x);
-    return
-
-  elseif (islogical (x)  &&  ~isscalar (x)  &&  nargin==1)
-    s = numeric_array_to_sym (x);
-    return
-
-  elseif (isa (x, 'double')  &&  ~isreal (x)  &&  nargin==1)
-    s = sym(real(x)) + sym('I')*sym(imag(x));
-    return
-
-  elseif (isinteger(x)  &&  nargin==1)
-    s = sym(num2str(x, '%ld'));
-    return
-
-  elseif (isa (x, 'double')  &&  nargin==1)
-    [s, flag] = magic_double_str(x);
-    if (~flag)
-      % Allow 1/3 and other "small" fractions.
-      % Personally, I like a warning here so I can catch bugs.
-      % Matlab SMT does this (w/o warning).
-      % FIXME: could have sympy do this?  Or just make symbolic floats?
-      warning('OctSymPy:sym:rationalapprox', ...
-              'Using rat() heuristics for double-precision input (is this what you wanted?)');
-      [N1, D1] = rat(x);
-      [N2, D2] = rat(x/pi);
-      if (10*abs(D2) < abs(D1))
-        % use frac*pi if demoninator significantly shorter
-        s = sprintf('Rational(%s, %s)*pi', num2str(N2), num2str(D2));
-      else
-        s = sprintf('Rational(%s, %s)', num2str(N1), num2str(D1));
+      if (nargin == 0)
+        s = sym(0);
+        return
       end
-    end
-    s = sym(s);
-    return
 
-  elseif (islogical (x)  &&  isscalar(x)  &&  nargin==1)
-    if (x)
-      cmd = 'z = sp.S.true';
-    else
-      cmd = 'z = sp.S.false';
-    end
-
-  elseif (nargin == 2 && ischar(varargin{1}) && strcmp(varargin{1},'clear'))
-    % special case for 'clear', because of side-effects
-    if (isa(x, 'sym'))
-      x = x.flat;    % we just want the string
-    end
-    s = sym(x);
-    % ---------------------------------------------
-    % Muck around in the caller's namespace, replacing syms
-    % that match 'xstr' (a string) with the 'newx' sym.
-    xstr = x;
-    newx = s;
-    context = 'caller';
-    % ---------------------------------------------
-    S = evalin(context, 'whos');
-    evalin(context, '[];');  % clear 'ans'
-    for i = 1:numel(S)
-      obj = evalin(context, S(i).name);
-      [newobj, flag] = symreplace(obj, xstr, newx);
-      if flag, assignin(context, S(i).name, newobj); end
-    end
-    % ---------------------------------------------
-    return
-
-  elseif (isa (x, 'sym')  &&  (nargin >= 2))
-    % support sym(x, assumption) for existing sym x
-    s = sym(x.flat, varargin{:});
-    return
-
-
-  elseif (isa (x, 'char'))
-    asm = [];
-    if (nargin == 2 && isequal(size(varargin{1}), [1 2]))
-      s = make_sym_matrix(x, varargin{1});
-      return
-    elseif (nargin >= 2)
-      % assume the remaining inputs are assumptions
-      asm = varargin;
-    end
-
-    doDecimalCheck = true;
-
-    % preprocess
-    if (strcmpi(x, 'inf')) || (strcmpi(x, '+inf'))
-      x = 'oo';
-    elseif (strcmpi(x, '-inf'))
-      x = '-oo';
-    elseif (strcmpi(x, 'i'))
-      x = 'I';
-    elseif (strcmpi(x, '-i'))
-      x = '-I';
-    elseif (strcmpi(x, 'nan'))
-      x = 'nan';
-    elseif (strcmp(x, 'lambda'))
-      x = 'lamda';
-    elseif (strcmp(x, 'Lambda'))
-      x = 'Lamda';
-    end
-
-    % Decide whether to pass to S() or Symbol()
-    if (any(strcmp(x, {'pi', 'I', 'oo', 'zoo', 'nan'})))
-      useSymbolNotS = false;
-    elseif (regexp(x, '^-?\d*\.?\d*(e-?\d+)?$'))
-      % Numbers: integers and floats
-      useSymbolNotS = false;
-    elseif (regexp(x, '^\w+$'))
-      % Words.  Note must follow numbers case.
-      % Use Symbol instead of S, e.g., for Issue #23:
-      % strcmp(x, {'beta' 'gamma' 'zeta' 'Chi' 'E' 'E1' 'Ei' 'S' 'N' 'Q'})
-      % But we also expect sym('Eq') to work, so match all single words
-      useSymbolNotS = true;
-    elseif (~isempty (strfind (x, '(') ))
-      % SymPy "srepr" or other raw python code
-      useSymbolNotS = false;
-      doDecimalCheck = false;
-    else
-      % Other non-symbols such as sym('1/3')
-      useSymbolNotS = false;
-    end
-
-    if (~useSymbolNotS)
-      % Use S(), as we're not forcing Symbol()
-      assert (isempty (asm))   % sym('pi', 'integer')
-      if (doDecimalCheck && ~isempty(strfind(x, '.')))
-        warning('possibly unintended decimal point in constructor string');
+      %% The actual class constructor
+      % Tempting to make a 'private constructor' but we need to access
+      % this from the python ipc stuff: outside the class.  We identify
+      % this non-user-facing usage by empty x and 6 inputs total.  Note
+      % that "sym([])" is valid but "sym([], ...)" is otherwise not.
+      if (isempty(x) && (nargin == 6))
+        s.pickle = varargin{1};
+        s.symsize = varargin{2};
+        s.flat = varargin{3};
+        s.ascii = varargin{4};
+        s.unicode = varargin{5};
+        s.extra = [];
+        %s = self;
+        return
       end
-      % x is raw sympy, could have various quotes in it
-      cmd = sprintf('z = sympy.S("%s")', strrep(x, '"', '\"'));
 
-    else % useSymbolNotS
-      if (isempty(asm))
-        cmd = sprintf('z = sympy.Symbol("%s")', x);
+      %% User interface for defining sym
+      % sym(1), sym('x'), etc.
 
-      elseif (isscalar(asm) && isscalar(asm{1}) && isstruct(asm{1}))
-        % we have an assumptions dict
-        cmd = sprintf('return sympy.Symbol("%s", **_ins[0]),', x);
-        s = python_cmd (cmd, asm{1});
+      %if (strcmp (class (x), 'symfun')  &&  nargin==1)
+      %  % FIXME: pass a symfun to sym() ctor; convert to pure sym
+      %  % (SMT does not do this in 2014a).  bad idea?
+      %  s = x.sym;
+      %  return
+
+      if (isa (x, 'sym')  &&  nargin==1)
+        % matches sym and subclasses
+        s = x;
         return
 
-      elseif (iscell(asm))
-        valid_asm = assumptions('possible');
-        for n=1:length(asm)
-          assert(ischar(asm{n}), 'sym: assumption must be a string')
-          assert(ismember(asm{n}, valid_asm), ...
-                 'sym: that assumption is not supported')
+      elseif (iscell (x)  &&  nargin==1)
+        s = cell_array_to_sym (x);
+        return
+
+      elseif (isnumeric(x)  &&  ~isscalar (x)  &&  nargin==1)
+        s = numeric_array_to_sym (x);
+        return
+
+      elseif (islogical (x)  &&  ~isscalar (x)  &&  nargin==1)
+        s = numeric_array_to_sym (x);
+        return
+
+      elseif (isa (x, 'double')  &&  ~isreal (x)  &&  nargin==1)
+        s = sym(real(x)) + sym('I')*sym(imag(x));
+        return
+
+      elseif (isinteger(x)  &&  nargin==1)
+        s = sym(num2str(x, '%ld'));
+        return
+
+      elseif (isa (x, 'double')  &&  nargin==1)
+        [s, flag] = magic_double_str(x);
+        if (~flag)
+          % Allow 1/3 and other "small" fractions.
+          % Personally, I like a warning here so I can catch bugs.
+          % Matlab SMT does this (w/o warning).
+          % FIXME: could have sympy do this?  Or just make symbolic floats?
+          warning('OctSymPy:sym:rationalapprox', ...
+                  'Using rat() heuristics for double-precision input (is this what you wanted?)');
+          [N1, D1] = rat(x);
+          [N2, D2] = rat(x/pi);
+          if (10*abs(D2) < abs(D1))
+            % use frac*pi if demoninator significantly shorter
+            s = sprintf('Rational(%s, %s)*pi', num2str(N2), num2str(D2));
+          else
+            s = sprintf('Rational(%s, %s)', num2str(N1), num2str(D1));
+          end
         end
-        cmd = ['z = sympy.Symbol("' x '"' ...
-               sprintf(', %s=True', asm{:}) ')'];
+        s = sym(s);
+        return
+
+      elseif (islogical (x)  &&  isscalar(x)  &&  nargin==1)
+        if (x)
+          cmd = 'z = sp.S.true';
+        else
+          cmd = 'z = sp.S.false';
+        end
+
+      elseif (nargin == 2 && ischar(varargin{1}) && strcmp(varargin{1},'clear'))
+        % special case for 'clear', because of side-effects
+        if (isa(x, 'sym'))
+          x = x.flat;    % we just want the string
+        end
+        s = sym(x);
+        % ---------------------------------------------
+        % Muck around in the caller's namespace, replacing syms
+        % that match 'xstr' (a string) with the 'newx' sym.
+        xstr = x;
+        newx = s;
+        context = 'caller';
+        % ---------------------------------------------
+        S = evalin(context, 'whos');
+        evalin(context, '[];');  % clear 'ans'
+        for i = 1:numel(S)
+          obj = evalin(context, S(i).name);
+          [newobj, flag] = symreplace(obj, xstr, newx);
+          if flag, assignin(context, S(i).name, newobj); end
+        end
+        % ---------------------------------------------
+        return
+
+      elseif (isa (x, 'sym')  &&  (nargin >= 2))
+        % support sym(x, assumption) for existing sym x
+        s = sym(x.flat, varargin{:});
+        return
+
+
+      elseif (isa (x, 'char'))
+        asm = [];
+        if (nargin == 2 && isequal(symsize(varargin{1}), [1 2]))
+          s = make_sym_matrix(x, varargin{1});
+          return
+        elseif (nargin >= 2)
+          % assume the remaining inputs are assumptions
+          asm = varargin;
+        end
+
+        doDecimalCheck = true;
+
+        % preprocess
+        if (strcmpi(x, 'inf')) || (strcmpi(x, '+inf'))
+          x = 'oo';
+        elseif (strcmpi(x, '-inf'))
+          x = '-oo';
+        elseif (strcmpi(x, 'i'))
+          x = 'I';
+        elseif (strcmpi(x, '-i'))
+          x = '-I';
+        elseif (strcmpi(x, 'nan'))
+          x = 'nan';
+        elseif (strcmp(x, 'lambda'))
+          x = 'lamda';
+        elseif (strcmp(x, 'Lambda'))
+          x = 'Lamda';
+        end
+
+        % Decide whether to pass to S() or Symbol()
+        if (any(strcmp(x, {'pi', 'I', 'oo', 'zoo', 'nan'})))
+          useSymbolNotS = false;
+        elseif (regexp(x, '^-?\d*\.?\d*(e-?\d+)?$'))
+          % Numbers: integers and floats
+          useSymbolNotS = false;
+        elseif (regexp(x, '^\w+$'))
+          % Words.  Note must follow numbers case.
+          % Use Symbol instead of S, e.g., for Issue #23:
+          % strcmp(x, {'beta' 'gamma' 'zeta' 'Chi' 'E' 'E1' 'Ei' 'S' 'N' 'Q'})
+          % But we also expect sym('Eq') to work, so match all single words
+          useSymbolNotS = true;
+        elseif (~isempty (strfind (x, '(') ))
+          % SymPy "srepr" or other raw python code
+          useSymbolNotS = false;
+          doDecimalCheck = false;
+        else
+          % Other non-symbols such as sym('1/3')
+          useSymbolNotS = false;
+        end
+
+        if (~useSymbolNotS)
+          % Use S(), as we're not forcing Symbol()
+          assert (isempty (asm))   % sym('pi', 'integer')
+          if (doDecimalCheck && ~isempty(strfind(x, '.')))
+            warning('possibly unintended decimal point in constructor string');
+          end
+          % x is raw sympy, could have various quotes in it
+          cmd = sprintf('z = sympy.S("%s")', strrep(x, '"', '\"'));
+
+        else % useSymbolNotS
+          if (isempty(asm))
+            cmd = sprintf('z = sympy.Symbol("%s")', x);
+
+          elseif (isscalar(asm) && isscalar(asm{1}) && isstruct(asm{1}))
+            % we have an assumptions dict
+            cmd = sprintf('return sympy.Symbol("%s", **_ins[0]),', x);
+            s = python_cmd (cmd, asm{1});
+            return
+
+          elseif (iscell(asm))
+            valid_asm = assumptions('possible');
+            for n=1:length(asm)
+              assert(ischar(asm{n}), 'sym: assumption must be a string')
+              assert(ismember(asm{n}, valid_asm), ...
+                     'sym: that assumption is not supported')
+            end
+            cmd = ['z = sympy.Symbol("' x '"' ...
+                   sprintf(', %s=True', asm{:}) ')'];
+          else
+            error('sym: invalid extra input, perhaps invalid assumptions?');
+          end
+        end % useSymbolNotS
+
       else
-        error('sym: invalid extra input, perhaps invalid assumptions?');
+        x
+        class(x)
+        nargin
+        error('conversion to symbolic with those arguments not (yet) supported');
       end
-    end % useSymbolNotS
 
-  else
-    x
-    class(x)
-    nargin
-    error('conversion to symbolic with those arguments not (yet) supported');
+      obj = python_cmd ({cmd 'return z,'});
+    end
   end
-
-  s = python_cmd ({cmd 'return z,'});
-
 end
 
 


inst/@sym/size.m
```diff
@@ -65,7 +65,7 @@
 
   % Note: symbolic sized matrices should return double, not sym/string.
 
-  n = x.size;
+  n = x.symsize;
```
genuinelucifer: This was required. Because if I used `size` as property in the `sym` class then too (weirdly) octave called the `size` function and did not use the property..

genuinelucifer:
I got some error about `no class symfun` so I thought of converting that to classdef too. Now many of the BIST tests are passing. :)
Hopefully we can have completely working classdefs within a few days...

inst/@sym/horzcat.m
```diff
@@ -62,7 +62,7 @@
           'return sp.Matrix.hstack(*_proc),'
           };
 
-  varargin = sym(varargin);
+  varargin = cell_array_to_sym(varargin);
```
genuinelucifer: This change (and other similar ones) are required because after converting `sym` to `classdef`, octave is not allowing to call the constructor with a cell array!!!
That is, `sym({1})` fails!!! I tried with some test classdef classes but the same error didn't appear. Something is different/weird with respect to `sym` only!

genuinelucifer: Yes I got that error message too. I just added the private dir to path... I had faced similar issues in past but could never figure out the cause. So i thought it was not related to classdef.
And, after classdef, did u try 'sym({1})'?? For me octave gave the error' cannot convert cell to object'

genuinelucifer: Also I am using 4.2-rc2, built from source downloaded from the gnu alpha website.

latot: mm, seems i misinterpret a little why you change sym to cell_array_to_sym, yes it don't works, but we can call normally the classdef with cells but the private function for some reason can't be called:
```
octave:1> sym({'f'})
error: 'cell_array_to_sym' undefined near line 210 column 13
error: called from
    sym at line 210 column 11
```
i try adding the function as private in the classdef but seems don't is supported for octave...

latot: only as a note i'm using the hg version (hg its like git for octave)

latot: a disadvantage of classdef is don't is reloaded if we change it.... exist some command for that? or  we need re-open octave in every change?

genuinelucifer: You are getting error not because of these changes. You can get rid of those errors by using "addpath('<path_to_octsympy>/inst/@sym/private')"
.
And yes i am using hg version. I don't think there's a git repo for octave, is there?

genuinelucifer: Instead of reopening octave you can do "clear classes". It will do clear all and then reload classes. 'rehash' would probably work too if you don't want to clear the workspace.

latot: okis, thx for the info, so about addpath, we will need add addpath to every method to do this works?

latot: now i get the right error:
```
error: cannot convert `cell' into `object'
```

latot: mm, seems addpath its a bad idea, it expose the private functions in the user env, in we use it we will need use addpath in the start in all methods and in the end use rmpath.

mtmiller: > a disadvantage of classdef is don't is reloaded if we change it.... exist some command for that?

Use `clear -c` or `clear classes`

mtmiller: Just realized @genuinelucifer already answered, sorry for the noise! 😄

mtmiller: The problem with `classdef` and `private` functions is a known bug that has not been fixed yet, see https://savannah.gnu.org/bugs/?45444

genuinelucifer: Hi @mtmiller, even after adding the private functions in path, we can't call the constructor of sym with cells! Although I tried this with some barebone classdefs but the error didn't appear. Could there be anything specific to sym constructor  that the octave parser isn't even trying to call the constructor? It just gives the error "cannot convert cell to object"!

mtmiller: I don't know of anything off the top of my head, please continue to debug to find out why this is happening.

mtmiller: IOW, I think the current situation of

    my_syms = sym ({'x', 'y', 'z'})

is neither good form nor does it seem to be legal under `classdef`-defined classes. A call to `sym (anything)` should always return an object of type `sym`.

Using `classdef`, it might be better to define a static method such as

    my_syms = sym.symarray ({'x', 'y', 'z'})

inst/@sym/sym.m
```diff
@@ -158,212 +158,224 @@
 %% Author: Colin B. Macdonald
 %% Keywords: symbolic, symbols, CAS
 
-function s = sym(x, varargin)
-
-  if (nargin == 0)
-    s = sym(0);
-    return
+classdef sym < handle
+  properties
```
mtmiller: Maybe make some or all of these properties `Access=Protected` or `Private`? Some are used outside of the class but many are only used directly inside class methods.

latot: Hi, i try it but didn't works D: i get the message can't found the function...

genuinelucifer: Well, currently only `symsize` is used outside the class. I will check to make sure and then keep others as Protected (just to allow symfun to use them if needed, assuming that derived class can use protected members of the base class in the classdef style too)....

inst/@sym/horzcat.m
genuinelucifer: Sure, it sounds better to be that way. But again, we have some odd conversions in place which might make it difficult (or rather would required many conditionals)....
Thanks for the debug.. I actually tried to debug it with breakpoints but the breakpoints could not be put on a classdef class.... I just read on the wiki that breakpoints via gui are not yet supported for classdef and have to be put via the cli....

genuinelucifer:
Hi, it seems that there is no way to define private functions in separate files (even within the same folder). Also, I ended up copying some of the functions from private folder directly into the `sym` class. Do we really want our single classdef to be sooo large? Is there a better work-around? Or should we wait till upstream patches the private functions bug in classdefs? If so, I would be happy to contribute a patch to fix that if anyone could guide me to the file(s) that govern the classdef behaviour...
.
Weirdly, `git push` is unable to resolve `github.com` even though I can successfully `ping github.com`.. I will try to push the changes later...

mtmiller:
Did you try defining the function prototype inside the `classdef` block and defining the method in a function file inside the `@sym` directory (but not inside a private subdirectory)? This works for me with a test class.

Inside `@foo/foo.m`:

    classdef foo < handle
    ...
      methods (Access = private)
        r = some_method (s);
      end

And inside `@foo/some_method.m`:

    function r = some_method (s)
    ...

This is the "official" way of defining private methods in external files on [Matlab's site](https://www.mathworks.com/help/matlab/matlab_oop/methods-in-separate-files.html), there is no mention of private subdirectories there.

Not saying it shouldn't be supported, but this seems to work in Octave (just move all methods up a level from private, and tag them as `Access = private` in the class definition).

genuinelucifer:
> Did you try defining the function prototype inside the classdef block and defining the method in a function file inside the @sym directory (but not inside a private subdirectory)?

I think yes. I actually copied all the files in private directory to the @sym directory. And then yes I included the prototype in the class. I played around with many different permutations of files in same directory or in private and different combinations of access of the methods. 
Although I am not sure about protected/private access... I'll check back tomorrow and reply back..

genuinelucifer:
Thanks @mtmiller . I finally found my mistake. I didn't know that to call functions within the same class too, I had to use `classname` 'dot' and then the function name! This behaviour is different from most of other OO languages... Altough it's fine for now, shouldn't it be allowed to call functions within the same class? Moreover to call a function by itself I have to use 'classname dot function_name'....
.
Almost all the @sym tests pass now. I will push the changes in a few minutes. Currently failing tests include symfun tests (all) because I can't seem to figure out how to call 'subasgn' of symfun, because octave automatically calls 'subasgn' of sym (the base class)... Also the tests involving the `==` operator are failing, though I haven't looked much into them.
.
The test summary for now is:
```
PASS      1803
FAIL       112
XFAIL       34
```
.
Also, one more concern now. We can't test the private functions!! Is there a way to test private functions in BIST tests?

mtmiller:
You are referring to static methods only, right? Yes, that is the way the language works, whether inside or outside of the class, a static method is called with `classname.methodname()`. Normal instance methods of a class can be called either as `method(obj, x, y, ...)` or `obj.method(x, y, ...)`

There is no way to test private functions. A BIST test is just like something you would run at the command line, and the problem is that the private function is not visible outside of its scope. So we have to keep BIST tests on user-facing functions only.

latot:
mm, some way is make a new public function of the classdeff with the function of test private functionalities and send an assert if don't works.

inst/@symfun/symfun.m
```diff
@@ -127,63 +127,81 @@
...
+        expr = sym(expr);
+      end
+
+      assert (isa (vars, 'cell'))
+      for i=1:length(vars)
+        assert (isa (vars{i}, 'sym'))
+      end
+
+      idx.type = ".";
+      idx.subs = {"vars"};
```
latot: ups remember here simples quotes .

latot:
This is lucking very nice! only a very very very little thing, its possible rename ```sym.symarray``` to ```sym.symcell```?? (array normally do reference to matrix, while cell, you know, cells)

genuinelucifer:
> You are referring to static methods only, right?

Yes. 

> we have to keep BIST tests on user-facing functions only.

Sure.

> some way is make a new public function of the classdeff with the function of test private functionalities and send an assert if don't works

Sure it's a way. But adding a 'public test function' to a class... I donno. Thoughts @mtmiller @cbm755 ?

> its possible rename sym.symarray to sym.symcell?? (array normally do reference to matrix, while cell, you know, cells)

Sure, I have no problem with the rename. I just picked out the name from one of Mike's comments. Also there is the function `cell_"array"_to_sym` which we are wrapping, so `symarray` seemed equally valid!
But, yes I too don't get the gist of 'array' here...

latot:
only as a note,actually is available to the user the function octsympy_tests, so don't think its too crazy add a test function.

cbm755:
Woo!  Looking good!  Why is it a subclass of handle?

Also re: BIST.  In Pytave, I did a `@pyobject/dummy.m` to hold the BIST that otherwise get missed due to bugs in Octave (typically the BIST from the ctor).  Whatever you do is probably not worse than that ;-)

re: `octsympy_tests.m`: that should go away soon: its a relic from Octave 3.6/3.8 and should be replaced by `runtests` or `__run_test_suite__`.

genuinelucifer:
> Woo! Looking good! Why is it a subclass of handle?

No particular reason. Probably passing by reference can help some memory and probably improve performance.. (assuming also treats handle subclasses like matlab)
Should I make it a value class?

mtmiller:
If I'm reading the code right, sym objects are immutable, right? So making it a value class would incur unnecessary copies of the internal properties for no added benefit. I think handle makes sense, unless I missed a spot where a sym can be modified in place.

inst/@sym/horzcat.m
cbm755: I filed #603 for this `cell = sym(cell)` thing.

latot:
Hi, only to be sure, Matlab have this behavior?, i refer to call ```subsasgn``` of the parent class instead of self (in a class of class context).

cbm755:
yes, I think Matlab has the same `subsasgn` behaviour.

latot:
maybe ask in their forums to know a what do here?

@genuinelucifer why symfun need their own ```subsasgn``` if the original don't need it?

a way to use is add ```subsref``` and ```subsasgn``` as static methods and call it with ```symfun.subsref```.

Thx. Cya.

genuinelucifer:
> why symfun need their own subsasgn if the original don't need it?

We because in the constructor of symfun if we try to assign value to a property then the subsasgn of 'sym' is being called which simply breaks on a assert statement. I didn't want to change behavior of sym class so though to implement it for symfun too.
.
P.S. - -  about the delay,  my laptop has stopped working (again)  due to hardware issues, i wil resume work on this as soon as possible.

latot:
Hi: https://github.com/genuinelucifer/octsympy/pull/5
i send to you this pr to fix subsasgn, but still symfun don't works, always get 0, but the assignment its right, so something else is failing, anyway, step by step.

i had move properties of symfun to sym because octave can't set properties of a subclass...
```
error: subsasgn: unknown property: vars
error: called from
    subsasgn at line 60 column 7
    symfun at line 200 column 9
    syms at line 176 column 10
```
```subsasgn``` and ```subsref``` need be static to can be called correctly from symfun.

Thx. Cya.

**Comment by genuinelucifer** [comment]
> i had move properties of symfun to sym because octave can't set properties of a subclass...

Are you sure? Is it documented anywhere?
As I can't test it as of now, I will wait for someone else (or until I find it documented) to confirm this before merging your changes.
Or could you do a little test with some barebone classdefs to confirm  this?

**Comment by latot** [comment]
Hi, sadly when i try reproduce it with a simplified version i can't do it works (i don't know to much about classdef), so for now to confirm it, someone else can change this line in symfun:
```
f.subsasgn (f, idx, vars);
```
to
```
f = builtin('subsasgn' f, idx, vars);
```
and you should have the error i post above.

Thx. Cya.

**Comment by latot** [comment]
the next commit its a simplified way to do this, remove the ```subsasgn``` of ```symfun``` and handle it from sym, you can choose what you like, or check both ways if can help.

**Comment by latot** [comment]
Hi all, well with the last commit symfun is fixed!
now, should we overload ```isa``` to fix symfun? or just move all ```isa(a, 'symfun')``` to ```isa(a, `sym')```?
```
octave:9> syms f(x)
octave:10> isa(f, 'symfun')
ans = 0
octave:11> diff(f)
ans = (sym)

  d       
  ──(f(x))
  dx     
```

Thx. Cya.

**Comment by genuinelucifer** [comment]
> with the last commit symfun is fixed!

Great :) 

> should we overload isa to fix symfun? or just move all isa(a, 'symfun') to isa(a, `sym')?

I would say we should have an overload. That seems more correct to me.

**Comment by latot** [comment]
Hi, its necessary expose ```symsize```?

and @cbm755 can we drop the first and last asserts of ```subsasgn```?
```
      assert( isa(rhs, 'sym'))
      assert( ~isa(idx.subs, 'sym'))
      assert( ~isa(val, 'sym'))
      val.(idx.subs) = rhs;
      out = val;
```
How now we are working with classdef i think we can use the properties more free, and how the prop are privates if the user try change someone octave will send the error, and if try set a non existent property octave will send an error too.

Thx. Cya.

**Comment by genuinelucifer** [comment]
> its necessary expose symsize?

Yes. It's actually used in 'size'  overload of sym to get the size! :P
We can definitely calculate it at every instant via python but that would just be a waste of resources.

**Comment by latot** [comment]
Hi, the private properties can be acceded from other functions, test it, seems works if is private:
```
octave:57> a=sym([1 2;3 4])
a = (sym 2×2 matrix)

  ⎡1  2⎤
  ⎢    ⎥
  ⎣3  4⎦

octave:58> size(a)
ans =

   2   2
```

Thx. Cya.

**Comment by latot** [comment]
mm, i found other problem, seems octave don't is calling methods of subclass in files....
maybe this is why subsref can't works...
So, for now i think if we continue this, we can move some functions from symfun and handle it from sym..... or move the the functions directly to the symfun subclass.
well before send a bug report to octave can someone say me if this code have some error plis?
```
pipe@Hi ~/Documentos/Cosas/octave/classes $ cat @A/A.m
classdef A
    methods
        function this = A()
            myMethod();
        end
    end
end
pipe@Hi ~/Documentos/Cosas/octave/classes $ cat @A/myMethod.m
function myMethod()
  fprintf('A:myMethod\n');
end
```
and the octave output:
```
octave:11> A
error: 'myMethod' undefined near line 4 column 13
error: called from
    A at line 4 column 13
```

Thx. Cya.

**Comment by mtmiller** [comment]
Yes, the way you defined `myMethod` it doesn't look like a method, it doesn't take an object argument, and you didn't call it on `this`. Maybe you meant either `this.myMethod()` or `myMethod(this)`?

**Comment by latot** [comment]
Hi, @mtmiller, thx for your comment you are right, i'm little confuse now, because checking the behavior of symfun, now works, but for some reason when we use it, its calling all methods from sym class...
now, checking the class:
```
octave:8> syms x
octave:9> a=symfun(x, {x})
a = (sym) x
octave:10> class(a)
ans = sym
```
but with the example of the previous comment i can confirm it should be the subclass (should return 'symfun')...
Well, we need check weird things.

Thx. Cya. and again thx @mtmiller!

**Comment by latot** [comment]
mm, this is very very weird, checking the class before and after of this line:
```
f@sym(expr);
```
i get this:
```
ans = symfun
ans = sym
ans = (sym) x
```
So, before init the subclass it is the symfun subclass, but when we init it its transformed to a sym class.................................................
I try this with a little example and it should works, but here is failing.....
at least here is the reason why symfun want calling their own methods.

Thx. Cya.

**Comment by mtmiller** [comment]
This is what prompted [#49421](https://savannah.gnu.org/bugs/?49421) right? When the `sym` constructor overwrites the `s` object, it's no longer a `symfun`, e.g.

      if (nargin == 0)
        s = sym(0);
        return
      end

I asked there for someone to test this kind of constructor chaining in Matlab to see if it's legal. If so, then maybe we need to fix something in Octave, and if not, then is there any recommended way in Matlab to have a base class constructor call itself with different arguments?

**Comment by genuinelucifer** [comment]
I tired many ways to debug this but no result. @mtmiller  could you please tell which version of matlab is  supposed to support this? I can get a copy of R2013b... Or I can ask one of my friends to try in R2016a if required...

**Comment by mtmiller** [comment]
I think any version since R2008 should be a candidate for testing this, but the newer the better for confirmation. Please take a look at the full example on [#49421](https://savannah.gnu.org/bugs/?49421), and maybe stick some debug `disp`s in there to see what the class of the `self` object is before and after calls to constructors at various places.

**Comment by mtmiller** [comment]
So it seems that Octave is Matlab compatible, the problem here is that it's bad practice to have a classdef-based constructor call itself if you are planning on subclassing that object. When a derived class calls a base class constructor, the object that is passed in is still an instance of the derived class. If the base class constructor calls itself and overwrites that object reference, then you've lost the derived instance and broken your class inheritance model.

Do any Matlab experts know a better way to do this? Or is the recommended model to just never have a constructor call itself? Does a constructor just have to have all possible conditional branches to handle different sets of arguments passed to it and fill in the properties of the object conditionally? There is no constructor or method overloading in Matlab, so maybe they figured there is no point to allowing a constructor to call itself.

**Comment by genuinelucifer** [comment]
@mtmiller I forgot to mention on the savannah bug that MATLAB gives an error if I call `derived()` and the base class constructor calls itself to override the object! Hence it is NOT possible in matlab for a base class to override a derived class object at all! So, in this minor way octave is not exactly compatible. But otherwise yes it seems that this is deprecated to call the constructor within itself.

**Comment by mtmiller** [comment]
Ok, thanks for clarifying. So in Octave it's allowed but produces bad results, in Matlab it's just an error. This is what I wanted you to help me discover.

So the sym class ctor will have to be reworked for the classdef implementation.

**Comment by genuinelucifer** [comment]
> So the sym class ctor will have to be reworked for the classdef implementation.

Sure. I'll do so and push the changes. This will also eliminate the need of `copy` which seems good.
.
Also, just for completeness, I get the following error in matlab:  
```
When constructing an instance of class 'derived', the constructor must preserve the class of the returned object
```

**Comment by latot** [comment]
Hi, well i want say i'm little worried about part of this, as i'm seeing this will need a lot of work and rewrite, now, actually i have this pr https://github.com/cbm755/octsympy/pull/594, its practically a rewrite of sym with a lot of fixes to octsympy, i don't like the idea start other rewrite when exist other, its possible in some way before continue merge that pr here or something similar?
You know work in this type of things we need time and all that things, and i think do this can help to can start working with a better base.

Thx. Cya.

**Comment by cbm755** [comment]
@latot, perhaps classdef change is going to need changes to upstream Octave (?) so it might be some time until this can be merged to master.

**Comment by genuinelucifer** [comment]
@cbm755 Actually as per te current conventions in both octave and matlab, we should rewrite thsym constructor so that it doesn't call itself! This might lead tomore redundant code or more helper functions but it seems the only solution as of now.  
Also, @latot 's work in #594 will really be helpful as he has already transferred much of the code to helper functions and hence the rewrite might be easier.

**Comment by cbm755** [comment]
Ok, so sounds like changes in #594 should go in ahead of these classdef changes.

**Comment by genuinelucifer** [comment]
I have based this on #594 and it seems this has some conflicts with master. I will rebase once that pr is merged. The last 2 commits are for this PR. Currently all the tests in `@logical` and `@double` pass and we have only 1 failing test in `findsymbols` in the outer directory because it needs `@symfun` to be classdef too. I will update other files one by one.
.
Also I am thinking of renaming `sym.symarray` to something like `sym.create_sym` because it is now used at a loooot of places and does a lot many things. Any suggestions?

**Comment by cbm755** [comment]
I rebase and merged quite a lot of #594: maybe enough that this could be rebased on top of it?

**Comment by cbm755** [comment]
> on top of it?

that is, on top of master

**Comment by genuinelucifer** [comment]
@cbm755 I have rebased this on top of master. As before, using `runtests .` in `inst` folder gives only 1 failing test since symfun is not a classdef yet. I will fix the tests in `@sym` dir and then convert symfun to classdef tomorrow.

**Comment by genuinelucifer** [comment]
I know it's too late to ask, but by making `sym` as classdef, we are basically removing MATLAB compatibility! Is this intended? Because if we convert to classdef then all the `sym([])`, `sym({1})` and other similar calls will have to use `sym.symarray` instead.  
While in MATLAB, all the former syntaxes are full correct. Moreover, I saw that in MATLAB `sym([1 2])` and `sym({1 2})` are same while `sym({1 2})` and `{sym(1) sym(2)}` are different as the former is converted to a sym array with 2 elements while the latter is a cell which has 2 sym objects! We are probably far from fully compatible in this case but still this classdef thing will stretch the gap even farther.

**Comment by cbm755** [comment]
I think your "Moreover" statement is the behaviour we want in both Octave and Matlab...  This is #603.

**Comment by mtmiller** [comment]
> too late to ask

It's never too late to ask.

Personally I have no idea how Matlab `sym` works, but I do think compatibility is important.

Agree with @cbm755 on #603, and I think that would be supported with classdef, right?

@genuinelucifer what do you think would not work with classdef? Is there a Matlab syntax calling `sym` that returns something that is not a `sym` object?

**Comment by cbm755** [comment]
Just to note there are two different forms of compatibility being discussed:
1.  Running Octsympy on Matlab.
2.  Octsympy being similar to SMT.
I think @genuinelucifer was referring to 1.

Note no one is working on 1., so it continually bit-rots due to, among other factors, Octave's "embrace and extend" of m-file syntax ;-)

**Comment by cbm755** [comment]
Why don't `sym([])` and `sym([1 2])` work?  Those are supposed to return `sym` objects, so its not the "sym must return sym" problem... (perhaps I just haven't re-read this thread carefully enough?)

File: inst/@sym/size.m (hunk 1 above)
Comment (cbm755) [inline inst/@sym/size.m]: Is that because `@sym/subsref` explicitly forbids public access of `x.size`?

File: inst/@sym/sym.m [hunk 5]
```diff
@@ -156,257 +156,312 @@
...
-        % ---------------------------------------------
       end
 
-      return
+      isnumber = isnumeric (x) || islogical (x);
+      assert (isempty (asm) || ~isnumber, 'Only symbols can have assumptions.')
 
-    else % S() in other case
+      if (~isscalar (x) && isnumber)  % Handle octave numeric matrix
+        error ('conversion to numeric array of symbols from the constructor is no longer supported. Use sym.symarray instead.');
```
cbm755: I replaced this error with `s = sym.symarray(x);` and it seems fine?  Can you look into this?  Lots of tests changed b/c of this...

genuinelucifer:
>  think @genuinelucifer was referring to 1.

I was actually referring to your point 2! Because in MATLAB we can do all of `sym([])` `sym({1 2})` and similar. While if we change to classdef we can only return 1 sym object and no arrays or cells.

cbm755:
> no arrays

This is incorrect, see my comment somewhere else.  But yes we'll need to change `sym({1 2})` to give `(sym) [1  2]   1x2 matrix`.

inst/@sym/sym.m
cbm755: @genuinelucifer see this please

genuinelucifer:
Thanks. I don't know why I did that earlier (giving error at 3 places). We just need error for conversion to cell array. I removed the `error` from other two places. I'll try to complete this PR with classdefs first. Then we can change the behaviour of cell->matrix in another PR.

genuinelucifer:
Now most of the tests are fixed Except:  
* 1 test in `findsymbols` as it requires `symfun`
* 10 tests in `dsolve` for `symfun`
* 1 test in `isNone`. I can't understand why this is failing. But probably we can't use classdef objects for indexing
* 5 tests in `subsindex` . Again I haven't look closely but seems same as `isNone`
* 2 tests in `sym` because of problem when writing a classdef object using `save`  
  
I have left the above just for my reference and if anyone wishes to test using the code till now. I will update `symfun` to classdef tomorrow and then target it's tests.

inst/@sym/mat_rclist_access.m
```diff
@@ -53,7 +53,7 @@
...
 % However, note these are generally tested elsewhere indirectly.
 
-%!test
-%! B = [1 2 3; 5 6 7];
-%! A = sym(B);
-%! assert (isequal (mat_rclist_access(A,[1 2],[2 3]), [B(1,2); B(2,3)]))
+%%!test
+%%! B = [1 2 3; 5 6 7];
+%%! A = sym(B);
+%%! assert (isequal (mat_rclist_access(A,[1 2],[2 3]), [B(1,2); B(2,3)]))
```
genuinelucifer: I commented out the tests in all the private function files because I use `runtests` command to run the tests in a directory and it tries to run the tests in these files too resulting in error. I hope it's ok.

cbm755: Yes, I think this is unavoidable if we move them out of `private/`.  This is b/c runtests and doctest are currently based on directory listings not `methods` calls.

cbm755:
Thanks for the fix!  Nice job so far.

genuinelucifer:
I am stuck now. Actually if I derive `symfun` from `sym` (which is in my last commit and I think it should be the right way) then octave always calls `subsasgn` of `sym` class in the constructor of `symfun` class and hence it gives error that the property `vars` is not found. Moreover, as can be seen, `f` is `sym` and NOT `symfun` in the constructor of `symfun` from the start. And hence the error!!! I think this has something to do with `superiorto` call which was earlier present but I removed it as it doesn't work in the constructor of a classdef. I think it is an upstream issue.  
.  
While if I derive `symfun` from `handle` then I can create a symfun but its display is not working and even if I define a display.m and do `display(f.sym)` it still doesn't work! Well, I didn't try this much as the first way must be preferred in my opinion.

cbm755:
If you explicitly (rather than implicitly) call the superclass constructor, the type of `f` is at least initially `symfun`.  Something like `f@sym(expr)` should call the superclass ctor.

But after the superclass ctor, `f` seems to be a `sym`; not sure if that is the correct behaviour or not.

Perhaps we need to make a minimal nonworking example to determine if its an octave bug or us?

cbm755:
Ok, I narrowed this down.  The issue is that the `sym(x)` ctor assigns to the input to the output whenever `x` is already a sym.  I think the workaround is that the symfun ctor needs to call the `%% The actual class constructor, Tempting to make a private constructor` bit.

cbm755:
Here's my minimal example.  Rename .txt to .m to play with these.
[myclassA.txt](https://github.com/cbm755/octsympy/files/709246/myclassA.txt)
[myclassB.txt](https://github.com/cbm755/octsympy/files/709247/myclassB.txt)

genuinelucifer:
great! Thanks for narrowing it down. I'll try to fix this tomorrow :)

cbm755:
A little more detail: I had to add `symsize` to `@sym/subsref`.  Then I changed `@symfun/symfun` to be:
````
 
   properties (Access = private)
     vars
-    sym
   end
 
   methods (Static, Access = private)
@@ -156,7 +155,6 @@ classdef symfun < sym
 
   methods
     function f = symfun(expr, vars)
-      class(f)
       if (nargin == 0)
         % octave docs say need a no-argument default for loading from files
         expr = sym(0);
@@ -208,8 +206,11 @@ classdef symfun < sym
         assert (isa (vars{i}, 'sym'))
       end
 
+      class(f)
+      f@sym([], expr.pickle, expr.symsize, expr.flat, expr.ascii, expr.unicode);
+      class(f)
+
       f.vars = vars;
-      f.sym = expr;
     end
````
Lots of things broken but at least it can make symfun's ;-)  Will let you think about this for a bit now.  For example, the other approach might be to do this in the case `sym(x)`, but I think this amounts to making copies so perhaps cannot/should not be a subset of `handle`.  Anyway, OO hurts my brain so enough for now ;-)

**Comment by genuinelucifer** [comment]
Thanks @cbm755! Finally symfun is working now :D  
Only the tests need to be edited now.

**Comment by genuinelucifer** [comment]
With my latest commit almost all tests in `@symfun` dir are now passing! :dancer:   
There are just problems with a few tests which I will comment inline as I don't understand why they should pass!

File: inst/@symfun/minus.m [hunk 7]
```diff
@@ -74,14 +74,14 @@
...
-%! h = x ./ f; assert(isa(h, 'symfun') && isequal(h.sym, x ./ g))
-%! h = x .^ f; assert(isa(h, 'symfun') && isequal(h.sym, x .^ g))
+%! h = f - x;  assert(isa(h, 'symfun') && isequal(h.symbol, g - x))
+%! h = f + x;  assert(isa(h, 'symfun') && isequal(h.symbol, g + x))
+%! h = f * x;  assert(isa(h, 'symfun') && isequal(h.symbol, g * x))
+%! h = f / x;  assert(isa(h, 'symfun') && isequal(h.symbol, g / x))
+%! h = f ^ x;  assert(isa(h, 'symfun') && isequal(h.symbol, g ^ x))
+%! h = f .* x; assert(isa(h, 'symfun') && isequal(h.symbol, g .* x))
+%! h = f ./ x; assert(isa(h, 'symfun') && isequal(h.symbol, g ./ x))
+%! h = f .^ x; assert(isa(h, 'symfun') && isequal(h.symbol, g .^ x))
```
genuinelucifer: since earlier it was `sym - symfun` so `minus` of `sym` was being called, so I changed it to `symfun - sym` to allow for proper call of out function. I don't know how the conversion/calling worked without classdef. If required, we can add an if statement in `sym` to convert results of `sym - symfun` to `symfun` too.

inst/@symfun/mtimes.m
```diff
@@ -58,7 +58,8 @@
 
 function h = mtimes(f, g)
   [vars, s1, s2] = helper_symfun_binops(f, g);
-  h = symfun(s1 * s2, vars);
+  res = s1 * s2;
+  h = symfun(res, symvar(res));
```
genuinelucifer: Well, this is most confusing part for me... When we are doing any arithmetic operator. The result will definitely have variables from both the symbols involved in that operation. In `helper_symfun_binops`, it's not allowed to do operations on 2 symfuns if they dont' have exactly same variables!! Also, the variables of the `symfun` are only considered as final variables in any operation!! This definitely looks wrong to me. I was tempted to change `helper_symfun_binops` to take a union of the sets of variables in each of the operands involved but I currently did this only for `mtimes`.  
Am I missing something? Why isn't operations for symfuns with different variables allowed? Why are variables only from the symfun operand considered as variables in final expression?

inst/@symfun/diff.m
```diff
@@ -78,7 +78,7 @@
 
 function z = diff(f, varargin)
 
-  z = diff(f.sym, varargin{:});
+  z = diff(f.symbol, varargin{:});
```
genuinelucifer: Since I could not find any direct way of converting a classdef to its parent class, I had to store expr in a `symbol` property. Also, If I name it `sym` then it doesn't work! So I named it `symbol`...

genuinelucifer:
Also, @mtmiller , I think I found a bug with classdefs. The files in a `@class` folder are not properly updated even after doing `rehash` or `clear classes`!! To reproduce do:
* Open octave and cd to `inst` directory of octsympy
* In linux terminal do: `mv @symfun/isequal.m ../../`
* Then use `isequal` on 2 `symfun` objects. Octave will try to call `@symfun/isequal.m` and since it doesn't exist, there will be error! I tried doing `rehash` `clear all` `clear classes` but nothing worked. I had to close octave and reopen it for this to work...

mtmiller:
There are quite a few bugs with classdef and `@class` scoped function files. I'm not quite following what you think the bug is, obviously if you mv the file out of the directory it won't exist anymore. Do you want it to revert to the builtin isequal function in this case? If so, then that looks like https://savannah.gnu.org/bugs/?36230, there's a workaround posted there.

inst/@symfun/diff.m
mtmiller: Or maybe in this case you could call `diff@sym(f, varargin{:})`. That is supposed to be the proper OO syntax for dispatching to a base class method.

inst/@symfun/mtimes.m
cbm755: I can't find a reference but IIRC, SMT did this too.  I agree it does seem reasonable to take the union of the inputs as the vars of the result.  But anyway this has little to do with this PR right?

>The result will definitely have variables from both the symbols involved in that operation

Hmm, but this may not be the correct result.  Consider:
````
octave:9> f(x,y) = 2*x
f(x, y) = (symfun) 2⋅x
octave:10> g(x,y) = 1
g(x, y) = (symfun) 1
octave:11> f*g
ans(x) = (symfun) 2⋅x
````
(Previous to this PR, that would be `ans(x,y)`.)

>  Also, the variables of the symfun are only considered as final variables in any operation!!

Not sure what you mean.

File: inst/@symfun/minus.m (hunk 7 above)
Comment (cbm755) [inline inst/@symfun/minus.m]: These tests should stay `sym - symfun`: the whole point of this test is the Octave bug #42735.  You can mark it `xtest` if you can't make it work rather than changing it.

I notice the older workaround gives infinite recursion so we may need a new workaround!  In the workaround in `@sym/plus`, I tried `plus@symfun(x, y)` but that gave an error message than `symfun` is not a subclass of `sym` (which is true enough).

In some cases, such as ``@sym/minus.m` we could do:
````
  % Dear hacker from the distant future... maybe you can delete this?
  if (isa(y, 'symfun'))
    warning('OctSymPy:sym:arithmetic:workaround42735', ...
            'worked around octave bug #42735')
    z = plus(-y, x);
    return
  end
````
(but I don't know we can do that for every case).

Comment (genuinelucifer) [inline inst/@symfun/minus.m]: Well, in minus atleast, we are calling minus on the `symbol` of symfun and the sym and then converting the result to symfun. If it's ok I can add conditions in `minus` of `sym` so that it converts the result to `symfun` if either of the operands is `symfun`.. Is that ok? Or should I mark it as xtest?

Comment (genuinelucifer) [inline inst/@symfun/minus.m]: Moreover it seems that `superiorto` can not be called in a classdef! So, I don't know if we can fix this with superiorto even if upstream is fixed.

File: inst/@symfun/mtimes.m (hunk 8 above)
Comment (genuinelucifer) [inline inst/@symfun/mtimes.m]: > But anyway this has little to do with this PR right?

Well, partly yes. But some tests are failing because they are not getting correct vars in result. Such tests are very less in number so I can leave them now and fix them in next PR.

> Hmm, but this may not be the correct result. Consider

Well, in the example you gave, the previous approach would work because both are `symfun` and both have the same variables.  

> Not sure what you mean

If we consider a symfun `f(x) = x^2` and a sym `y` then `f(x)*y` should have both `x` and `y` in its `vars`. But since the code of `helper_symfun_binops` considers only the variables of the symfun involved in such cases,the result will have only `x` and not `y` included in the `vars`.  
  
I see that my current approach of `symvar` is wrong too. I can change it to take union of the variables from both the operands involved. I think that would definitely work.

File: inst/@symfun/minus.m (hunk 7 above)
Comment (cbm755) [inline inst/@symfun/minus.m]: Its not called "superiorto" in classdef: I've posted at the [upstream bug]  (https://savannah.gnu.org/bugs/?42735).

@genuinelucifer Maybe you're the right person, at the right time, at the right place to try to fix this upstream bug?  I'm sure @mtmiller would be very willing to review a patch for such thing!

File: inst/@symfun/diff.m (hunk 9 above)
Comment (genuinelucifer) [inline inst/@symfun/diff.m]: Oh thanks! Will do.

File: inst/@symfun/minus.m (hunk 7 above)
Comment (cbm755) [inline inst/@symfun/minus.m]: >  If it's ok I can add conditions in ...

Dunno, try I guess.  My impression was if we cannot call the subclass method than all the subclass code will have to be duplicated in `@sym`... ick!  Fix upstream would be best of all of course!

**Comment by genuinelucifer** [comment]
> Do you want it to revert to the builtin isequal function in this case? 

Yes. As it would if there was no such file to begin with.

> If so, then that looks like https://savannah.gnu.org/bugs/?36230, there's a workaround posted there.

Oh. Thanks. That worked for me too.

File: inst/@symfun/minus.m (hunk 7 above)
Comment (genuinelucifer) [inline inst/@symfun/minus.m]: > Maybe you're the right person, at the right time, at the right place to try to fix this upstream bug?

Sure. After finishing this PR, I will try to submit a patch upstream...

File: inst/@symfun/mtimes.m (hunk 8 above)
Comment (cbm755) [inline inst/@symfun/mtimes.m]: > I can change it to take union of the variables from both the operands involved.  

Let's take this conversation to #738.  I don't see it being relevant to this PR.

**Comment by genuinelucifer** [comment]
Finally now most of the tests in `symfun` are working! I have made a few tests in `minus` and `symvar` as `xtest` since as per me they ought to fail. I can't understand why they aren't failing without classdef.  
  
Also the test in `findsymbols` is fixed automatically.  
  
But more problems have arrised..... Now `dsolve` has 14 fails and `solve` has 10 fails. Both are failing because the equations are automatically evaluated and not being solved as equations and hence they cannot be solved!  
  
Moreover, it seems that, for classdef, callling `a(x)` where `x` is the object of `sym` and `a` is a normal vector/integer variable is NOT allowed at all. As in such cases the call gives error without going to subsref at all... So there are now failing tests in `@sym/inNone` and `@sym/subsref`...
  
Lastly, there are 2 failing tests in `@sym/sym`!! One is because, apparently, `save` cannot be called on a classdef object!! And the other one is some error in assumptions. I marked both of them as xtests. I will look into them later. Currently I am just amazed at how many more errors there are to debug!!! :(

File: inst/@symfun/mtimes.m (hunk 8 above)
Comment (genuinelucifer) [inline inst/@symfun/mtimes.m]: Sure

**Comment by mtmiller** [comment]
> apparently, `save` cannot be called on a classdef object

https://savannah.gnu.org/bugs/?45833

**Comment by cbm755** [comment]
One cause of some of the failing tests is because evaluating a symfun at a variable should give a sym:
````
>> syms x
>> f(x) = 2*x
f(x) = (symfun) 2⋅x
>> f(x)
ans = (sym) 2⋅x
````
But this PR currently gives a symfun `ans(x) = (sym) 2⋅x`.

File: inst/@symfun/subsref.m [hunk 10]
```diff
@@ -43,19 +43,20 @@
 
   switch idx.type
     case '()'
-      out = subs(f, f.vars, idx.subs);
+      out = subs (f.symbol, f.vars, idx.subs);
+      out = symfun (out, symvar (out));
```
cbm755: this `out = symfun (out, ...)` is incorrect.

genuinelucifer:
Thanks to @cbm755 , all the tests are passing now! :D  
This PR is hence complete and ready for final review, in my opinion. So, please test it and review it once more.  
  
I have just a couple of questions left now:
* What would be the best way to resolve the conflicts without having to rewrite all the changes of this PR on top of master? Till now, to resolve conflicts, I have simply been deleting my branch and creating a new branch from master and rewriting all the required changes. Since this PR spans so many files with little changes in each one, it doesn't look feasible to me.
* Should I squash some of the starting comments? The ones like `Fix almost all symfun tests!` and `Fix more symfun problems`?

cbm755:
I would `git merge origin/master` into this branch, then fix the conflicts.

Before doing that, maybe you could review #731?  If we merge that to master, that should make a bunch of the `sym.symarray` stuff unneeded.

(this classdef change is not going to be merged before 2.5.0 anyway).

genuinelucifer:
> I would git merge origin/master into this branch, then fix the conflicts.

Thanks I tried this with `meld` as mergetool and looks so easy. Good to know this.

> this classdef change is not going to be merged before 2.5.0 anyway

Oh ok. Then I would wait for that as the new master will have to be merged again anyway.  
  
Also, Currently we have following 3 outstanding errors due to use of classdef (just to sum it up):
* Define `symfun` as superior to `sym` so as to call its functions like that in case of `minus`
* Allow `save`ing of `sym` and `symfun`
* Indexing of constants with classdef is not supported in octave. Either get rid of those tests or leave them as `xtests` till octave supports it.  
  
I am assuming we will translate to classdef hoping that all these will soon be supported by octave and we will get rid of all the `xtest`s. In the mean time, I will try to provide patches for fixing some/all of these points in octave. But, this will still bump the minumum required version of octave for the next release of symbolic to atleast `4.4`. I am assuming that it's not a problem...

cbm755:
Re: your points:

1.  please at least make our class have the inferior classes thing (`classdef {SOMETHING ABOUT sym GOES HERE} symfun < sym`)
2.  saving: seems we cannot merge this until this is fixed.
3.  "Indexing of constants with classdef is not supported in octave": citation needed, do you have an upstream bug report?   Definitely leave them as xtests for now.  If there is no upstream report, we'll need to make a minimal example and file it.

> final review

Wishful thinking ;-)  It seems quite likely we will not merge this until 4.4 is released :(

cbm755:
I think master is reasonably stable; would be good to merge to here.  We also need a clear idea of what needs fixed in upstream octave to get this merged.

genuinelucifer:
Well, the 3 things that I mentioned in my last comment seem to be the bottleneck currently.
* defining a classdef superiorto some other classdef
* saving and loading classdef objects
* indexing constants with a classdef object

cbm755:
Do those all have upstream issue numbers?

mtmiller:
* https://savannah.gnu.org/bugs/?45833 - load and save on classdef

I am not sure what is meant by "indexing constants with a classdef object". Does this mean using a classdef object as the index of a subsref on a builtin matrix? Is there some builtin conversion rules that are supposed to be used? Do we have a Matlab doc describing this?

I thought there were some other bugs. Are there other classdef issues that you were able to find workarounds for, but it would still be better to not have to use a workaround? For example method name resolution problems, problems with private methods, etc?

cbm755:
> "indexing constants with a classdef object"

@genuinelucifer do you mean this?
````
>> a = [1.2 3.4]
a =

         1.2         3.4

>> a(sym(3)) = 42.42
a =

         1.2         3.4       42.42
````

genuinelucifer:
> @genuinelucifer do you mean this?

Yes, exactly.  
  
Sorry,  I've been busy with my exams and couldn't reply.

Issue #591, opened 2016-10-01
fix euler number in sym (and probable other too)


latot:
Missed symbols:

```
sym(e)
warning: Using rat() heuristics for double-precision input (is this what you wanted?)
warning: called from
    sym at line 225 column 7
ans = (sym)

  1457
  ────
  536 
```

latot:
from https://www.gnu.org/software/octave/doc/v4.0.1/Mathematical-Constants.html
seems e is the only left.

cbm755:
We could do this, similar to our special treatment for pi...  My thoughts:

(+) It is of no concern for SMT compatibility b/c AFAIK `e` is not a command in Matlab (you have to say `exp(1)`).

(-) One downside is you cannot do `syms a b c d e` to get some vars...  I remember something similar that used to frustrate me in Maple, where `D` was a command.

(-) If we do this, why not make `sym(sqrt(2))` work, `sym(sqrt(2)/2)` etc.  Its kind of a slippery slope...

I'm undecided.  Why did you want this?  Just because Octave has `e` in the document you linked above?

latot:
mm, i use a lot the number e, but when we work with octave we have written the scripts only with that letter, like ```e^x``` but when we convert the expressions to sym we found e its a symbol and its not detected like pi..
about ```sym(sqrt(2))``` i think sym shouldn't trate with that type of expressions, like only symbols, i only think this because the sym file fo this..
Octave vs SMT?

mtmiller:
My first inclination would be to use

    e = exp (sym (1))

if I needed `e` as a constant.

I originally thought this would be a good idea to match `pi`, but since it's a single letter I agree with @cbm755 that it would be bad to make it hard-to-impossible for users to define a sequence of symbolic vars `a b c d e f g`.

latot:
mm, okis, the only think i'm little worried is the octave doesn't have how to know this behavior.

cbm755:
Perhaps the real issue here is that `sym(char)` should generally, in most cases, just make variables whereas `sym(double)` could have magic behaviour (like `sym(pi)`).  That seems closer to how SMT is these days.  I've filed #604.

If `syms e` could still make a variable called `e`, then I think I'm +1 for having `sym(e)` (and `e^x`) do the right thing...

latot:
only for the record, if the decision is keep e as variable, maybe send a warning to say the difference?
```Warning: the e expression don't is the euler number.```

latot:
[...]

PR #592, opened 2016-10-01
fix euler number

latot:
Fixes https://github.com/cbm755/octsympy/issues/591

### DIFF
--- a/inst/@sym/sym.m
+++ b/inst/@sym/sym.m
@@ -299,6 +299,12 @@ function s = sym(x, varargin)
       x = 'lamda';
     elseif (strcmp(x, 'Lambda'))
       x = 'Lamda';
+    elseif (strcmp(x, '-e'))  %%Special case, don't assign e to E.
+      s = python_cmd ('return -E');
+      return
+    elseif (strcmp(x, 'e'))
+      s = python_cmd ('return E');
+      return
     end
 
     % Decide whether to pass to S() or Symbol()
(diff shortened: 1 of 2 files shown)


latot:
test pass in python 3.4, 2.7 timeout.

inst/@sym/sym.m
```diff
@@ -299,6 +299,12 @@
       x = 'lamda';
     elseif (strcmp(x, 'Lambda'))
       x = 'Lamda';
+    elseif (strcmp(x, '-e'))  %%Special case, don't assign e to E.
+      s = python_cmd ('return -E');
```
cbm755: why not `x = '-E'` like other cases

inst/@sym/sym.m
```diff
@@ -299,6 +299,12 @@
...
+      return
+    elseif (strcmp(x, 'e'))
+      s = python_cmd ('return E');
+      return
     end
 
     % Decide whether to pass to S() or Symbol()
```
cbm755: possible need to add `E` below?

inst/@sym/sym.m
```diff
@@ -693,3 +699,9 @@
 %! else
 %!   delete([myfile '.mat'])
 %! end
+
+%!test
+%! a = sym(e);
```
cbm755: This `sym(e)` test will fail on Matlab: needs wrapped in the `if exist OCTAVE_VER` construct.

latot:
hi, i considere this pr as deprecated, please check this: https://github.com/cbm755/octsympy/pull/594
we can keep this open to check or decide some things

cbm755:
@latot are you likely to redo this on top of master?

cbm755:
Closing in favour of #824, which is similar to this but done on top of master.

PR #558, opened 2016-09-18
disable symfuns with the same names of python funcs env.

latot:
Fixes https://github.com/cbm755/octsympy/issues/290

To know details about this check the same issue.

Basically, disable the symfuns with a equal name of a python function.

Cya.

Issue #421, opened 2016-04-21
support `assume x real` as well as `assume(x, 'real')`


cbm755:
```
syms x
assume x positive
assumeAlso x integer
```

Perhaps this makes it a little more obvious that `assume` is going to muck around in your namespace!

Note: SMT doesn't seem to do this.

cbm755:
I'd like to make a decision on this for 2.5.0.  Main arguments in favour:

1.  `syms x y clear` is deprecated so it would be nice for people to change their code to `assume x y clear`.  If we don't do this, they would need to write `x = assume(x, 'clear'); y = assume(y, 'clear')`.
2.  As I mentioned above, this command form "makes it a little more obvious that `assume` is going to muck around in your namespace!"

Main argument against:

1.  SMT does not do this.

@genuinelucifer @mtmiller any objections to doing this?

@latot you already had a PR that did this so I assume you agree.  Can you do a PR that does *only* this?

genuinelucifer:
+1
It seems perfectly reasonable to me that allowing this will help the end users.

mtmiller:
I'm not necessarily in favor, but another pro:

1. tab-completion and help on assume would work, since it would have to be an ordinary function, not a method

I see no problem with supporting this syntax as an extension.

I guess that this command form would be a separate function that only operates on strings, with an arbitrary number of sym names? While the method form `@sym/assume` would still operate on one sym variable at a time?

cbm755:
Thank you both for the quick replies!  @mtmiller your last comment gives me pause; had not thought of that.  Another con:

2. `assume x y positive integer` would imply that `assume(x, y, 'positive', 'integer')` should be supported (?) and we might not want to do the latter.

@mtmiller's suggestion that `inst/assume.m` could be a separate function from `inst/@sym/assume.m` is one resolution.  (I was originally assuming we just move `inst/@sym/assume.m` up one directory, but maybe not).

mtmiller:
Yeah, I'm thinking it would be safer to have `inst/assume.m` only operate on strings, and dispatch to `@sym/assume.m` in a loop over all given symbol names. This way the true compatible `@sym/assume.m` can stay as it is and only operate on a single sym object as in Matlab, avoiding any possible confusion or complications of argument handling there.

PR #594, opened 2016-10-02
Fix sym things and other things

latot:
Hi, well this time i have a rewrite of the sym function, its because in some time i found problems, parts of redundant code, etc. I try in this cover all i can to improve it fixing things, sadly when i try rewrite it i found a lot of hidden problems, with this i can fix it but i forgot more than 50% of the problems, when i remember someone i'll update this message, theoretically this have less lines than the original but with the fixes now are relative similar (like 57 lines less), one of the others objetives is do some code more flexible and centralize some concepts (obs i try optimize this).

Other point, when i try fix things sadly i must start fixing other things to do this works, i'll try explain why of every part, i put comments to do this more readable, and the code is simplified.

Can be little hard in some way because i don't format this code yet.

I need this pr https://github.com/cbm755/octsympy/pull/577 for the flatten function, i'll need update it, but only for avoid branch conflicts.

This cover this prs:
  Fixes https://github.com/cbm755/octsympy/pull/558
  Fixes https://github.com/cbm755/octsympy/pull/559
And this issues:
  Fixes https://github.com/cbm755/octsympy/issues/628
  Fixes https://github.com/cbm755/octsympy/issues/627
  Fixes https://github.com/cbm755/octsympy/issues/626
  Fixes https://github.com/cbm755/octsympy/issues/625
  Fixes https://github.com/cbm755/octsympy/issues/624
  Fixes https://github.com/cbm755/octsympy/issues/623
  Fixes https://github.com/cbm755/octsympy/issues/622
  Fixes https://github.com/cbm755/octsympy/issues/621
  Fixes https://github.com/cbm755/octsympy/issues/620
  Fixes https://github.com/cbm755/octsympy/issues/619
  Fixes https://github.com/cbm755/octsympy/issues/618
  Fixes https://github.com/cbm755/octsympy/issues/617
  Fixes https://github.com/cbm755/octsympy/issues/290
  Fixes https://github.com/cbm755/octsympy/issues/421
  Fixes https://github.com/cbm755/octsympy/issues/612

Others issues i found and i can remember:
  `sym({'x', 'y', 'z'}, 'positive')` don't set the assumptions correctly
  `sym('x', 'clear')` removed from matlab some time ago, moved to assume for SMT compat.
  `assume('x', 'real')` now create correctly the var in the workspace, and don't was handling correctly the char input
  Simplify handle octave symbols with python.
  All assumptions are checked now, with struct, cells and chars.
  `sym('a', sym([1 2]))` don't create the right matrix.
  `sym('a', [9999999 999999999])` that should not be showed and changed to return a symbolic matrx, actually max is 20x20
  Fixed format in negative assumptions
  Enabled negative assumptions in format `assume('x', 'polar', 'complex', false)`
  Fix `sym("1.1") = 1.100000000000000`
  Send a warning when we try make a sym the a system var name - `sym('beta(a, b)')`
  Warning if you try make a operation from sym function

Well, this is all i can remember.

The base concept of this changes is ready!

Hi, now i think the focus of this pr will be end like here, i don't want change other things and extend more this.

Now only remains wait the pr 577.(i need check and fix the assumptions format with that)

Cya.


inst/@sym/private/cell_array_to_sym.m
```diff
@@ -16,7 +16,7 @@
 %% License along with this software; see the file COPYING.
 %% If not, see <http://www.gnu.org/licenses/>.
 
-function s = cell_array_to_sym (L)
+function s = cell_array_to_sym (L, varargin)
 %private helper for sym ctor
```
latot: Fix assumptions for cell syms

inst/@sym/private/check_assumptions.m
```diff
@@ -0,0 +1,62 @@
...
+%% @deftypefun check_assumptions (@var{x}, @var{dot})
+%% Check if the input have valid assumptions.
+%%
+%% Private helper function.
+%%
+%% @seealso{sym}
+%% @end deftypefun
+
+function check_assumptions(varargin)
+
```
latot: Helper to avoid write this in the sym file, and we can extend this for better functionality # # @

inst/@sym/private/magic_double_str.m
```diff
@@ -28,27 +29,58 @@
 %% @seealso{sym, vpa}
 %% @end deftypefun
 
-function [s, flag] = magic_double_str(x)
+function [s, flag] = magic_double_str(x, in)
+
```
latot: Handle comparison of symbols octave/python in one place.

inst/@sym/private/make_sym_matrix.m
```diff
@@ -19,35 +19,32 @@
...
 
-  assert (ischar (As));
-  assert (isequal (size(sz), [1 2]));
-  if (isa(sz, 'sym'))
+  s = size(sz)(2);
+  assert (s <= 2, 'SymbolicMatrix actually only support max 2 dimesions');
+
+  sz = sym([sz zeros(1, 2-s)]);
+
+  if ~isempty(findsymbols(sz))
```
latot: fixes ```sym('a', sym([1 2]))```

inst/@sym/sym.m
```diff
@@ -190,179 +191,132 @@
...
+             '        x2 = [p for p in x2 if p]'
+             '        for i in x2:'
+             '            try:'
+             '                if eval("callable(" + i + ")"):'
+             '                    lis.add(i)' 
+             '            except:'
+             '                pass'
+             '    if len(lis) > 0:'
+             '        return (1, "\", \"".join(str(e) for e in lis))'
+             '    return (2, x)'};
```
latot: Fixes ```FF(w)```

inst/@sym/sym.m
```diff
@@ -487,7 +441,7 @@
 %! f = {x {2*x}};
 %! asm = assumptions();
 %! assert ( ~isempty(asm))
-%! x = sym('x', 'clear');
+%! assume('x', 'clear');
```
latot: ```assume('x', 'clear')``` is fixed, so now we can use it.

inst/assume.m
```diff
@@ -105,18 +106,67 @@
...
+        ca.(cond) = varargin{n};
+      else
+        ca.(cond) = true;
+      end
+    else
+      if (n == 2 && islogical(varargin{1})) || islogical(varargin{n - 2})
+        print_usage();
+      end
+    end
+  end
```
latot: enable negative assumptions

inst/assume.m
```diff
@@ -105,18 +106,67 @@
...
+
+  if isa(x, 'sym')
+    xstr = x.flat;
+  else
+    xstr = x;
```
latot: fix string/sym input

inst/assumptions.m
```diff
@@ -111,10 +111,8 @@
       'd = x._assumptions.generator'
       'if d == {}:'
       '    astr = ""'
-      'elif all(d.values()):'  % all True so list them
-      '    astr = ", ".join(sorted([str(i) for i in d.keys()]))'
-      'else:'  % more complicated case, just the raw dict
-      '    astr = str(d)'
+      'else:'  % all True so list them
+      '    astr = ", ".join(sorted([(("" if d.values()[i] else "~") + str(d.keys()[i])) for i in range(len(d.keys()))]))'
```
latot: fix format for negative assumptions.

inst/@sym/sym.m
cbm755: Maybe its `assume(x, 'clear')` for SMT compat?

inst/assume.m
```diff
@@ -105,18 +106,67 @@
...
+function s = assume(x, varargin)
+
+  if nargin == 0
+    error ('Send an argument.')
+  elseif nargin == 1
+    error ('Not supported yet.')
+  end
+
+  if isa(varargin{1}, 'char') && strcmp(varargin{1}, 'clear')
+
```
cbm755: "moved"?  `a = sym('a', 'clear')` should still work right?

inst/assumptions.m
cbm755: can you give me an example of a "negative assumption"?

latot: Hi, actually are two ways, i fix that concept too enabling it from sym, first the normal example:
```
octave:6> a=python_cmd('return Symbol("a", complex=True, real=False)')
a = (sym) a
octave:7> assumptions(a)
ans = 
{
  [1,1] = a: complex, ~real
}
```
and with the possible syntax:
```
octave:8> b=sym('b', 'complex', 'real', false)
b = (sym) b
octave:9> assumptions(b)
ans = 
{
  [1,1] = b: complex, ~real
}
```
i need check some things about it for the moment, this will be better with the pr https://github.com/cbm755/octsympy/pull/577 to normalize the input.

latot: only as a note, this expressions works fine from assume and assumeAlso, i need finish sym expression yet.

latot:
@cbm755 Hi,I finally, finally finish this, i search and fix every single detail, compact code, fix behavior, and all that things i can found, i know this seems too big, but only seems, the only big rewrite is the sym function, the others are helpers functions (some one in /private and sym/private) to don't repeat code everywhere and formatted tests, for that reason this have a lot of lines, and for some reason git detect the assume file as deleted and created again... (so more lines)

Obvs, all of this is open to discuss, i recommend read the main files directly, don't from diff of git.
Some parts of python scripts can be compacted with the pr 577 but if you finish this first, that can be in other pr.

One of the fixes i like here is sym now support correctly the char decimals, ```sym('1.3484293847982')```, now return the correct value,  the main message of the issue have a lot of info and a list of main changes.

I use one concept when i check the sym function, the sym function its the user input data, so is the instance to do all checks we can do to avoid bad behaviors in the rest of the program.

I try the code be intuitive, and is commented if something is don't clear.

About the expression ```sym('2*1.0')``` are locked because ```S()``` of some expression don't is calculated correctly, and when the expression have an operator don't is easy split every op and call ```Rational``` in every single case and replace it, for now it send an error, and you try do a operation in the sym function send a warning.

Check related issues of Sympy:
https://github.com/sympy/sympy/issues/11703
https://github.com/sympy/sympy/issues/11704

Thx. Cya.

latot:
Hi, should i add new issues for all things i try fix here?

cbm755:
> should i add new issues for all things i try fix here?

Probably a good idea.

inst/test_bugs.tst
```diff
@@ -223,3 +223,14 @@
 %! % FIXME: not same as an freshly created 5 x 6.
 %! C = sym('B', [5 6]);
 %! assert (isequal(B, C))
+
+%!test
+%! f1 = fileread ('private/magic_double_str.m');
+%! f2 = fileread ('@sym/private/magic_double_str.m');
+%! assert (strcmp (f1, f2))
```
genuinelucifer: Do we need the exact same files at 2 places? I mean what if something changes at one place, is it necessary to change it at another place too?

inst/assumeAlso.m
```diff
@@ -66,29 +66,51 @@
 %% Author: Colin B. Macdonald
 %% Keywords: symbolic
 
-function varargout = assumeAlso(x, varargin)
```
genuinelucifer: Sorry if this is obvious but why did you move this file?

inst/@sym/sym.m
```diff
@@ -190,239 +189,231 @@
...
+        case 2  %%Something else
+          disp (['Python: ' err]);
+          error (['You can not use var name "' s '" for a error, if is a bug please report it.']);
+      end
+      return
+    end
   end
 
-  s = python_cmd ({cmd 'return z,'});
+  error ('Input not supported yet.')
```
genuinelucifer: Although this is ok too but how about the original text `conversion to symbolic with those arguments not (yet) supported`?

inst/@sym/sym.m
```diff
@@ -190,239 +189,231 @@
...
+      s = x;
+    else
+      s = sym (x.flat, varargin{:});
+    end
     return
+  end
 
-  elseif (iscell (x)  &&  nargin==1)
-    s = cell_array_to_sym (x);
+  if iscell(x)  %%Handle Cells
```
genuinelucifer: space after `iscell`?

inst/@sym/private/symsnotfunc.m
```diff
@@ -0,0 +1,54 @@
...
+        symsnotfunc (varargin{i}{j});
+      end
+    elseif isa (varargin{i}, 'char')
+      p = regexp (varargin{i}, '\(', 'split'){1};
+      [x, flag] = magic_double_str (p);
+      if ~flag
+        k = exist (p);
+        if k == 5 || k == 8 || k == 2
+          disp (['warning: The expression "' varargin{i} '" its a system expression.']);
+          warning ('You can without intention overload it.');
```
genuinelucifer: Probably a more assertive statement here? Like you "are" overloading/hiding it?

inst/@sym/private/symsnotfunc.m
```diff
@@ -0,0 +1,54 @@
...
+%%
+%% @seealso{sym}
+%% @end deftypefun
+
+function [s, flag] = symsnotfunc (varargin)
+
+  for i=1:length (varargin)
+
+    if iscell (varargin{i})
+      for j=1:length (varargin{i})
```
genuinelucifer: space around `=`

inst/@sym/private/symsnotfunc.m
```diff
@@ -0,0 +1,54 @@
...
+%% Detect if @var{x} is a octave function.
+%%
+%% Private helper function.
+%%
+%% @seealso{sym}
+%% @end deftypefun
+
+function [s, flag] = symsnotfunc (varargin)
+
+  for i=1:length (varargin)
```
genuinelucifer: space around `=`

inst/@sym/private/check_assumptions.m
```diff
@@ -0,0 +1,53 @@
...
+      fields = fieldnames (x);
+      for j=1:numel (fields)
+        check_assumptions (fields{j})
+      end
+    elseif iscell (x)
+      for j=1:length (x)
+        check_assumptions (x{j})
+      end
+      else
+      error('sym: assumption must be a string or struct or cell')
```
genuinelucifer: space before `(`

inst/test_bugs.tst
latot: yes, actually vpa(is out of ```@sym```) use this file too, so we need keep update both files.

inst/assumeAlso.m
latot: Hi, issue https://github.com/cbm755/octsympy/issues/421

inst/@sym/sym.m
latot: okis!

latot:
Hi! @genuinelucifer thx for your comments and revision!, i update the code with your recommendations.

Feel free to ask all you want about the code, if something don't is clear i think would be better add more comments.

About ```(``` and ```)```, i normally try skip redundant expressions,like ```if (ismatrix(K))```, because what is the idea of that extra?, obvs if you think for a esthetic is necessary or can help to understand better the code i'll change it, well out of this, in what part you think its necessary use it?

Thx. Cya.

genuinelucifer:
As per http://wiki.octave.org/Octave_style_guide#ending_blocks we should use `( )` around `if`s... Also do not use `%%` for inline comments. Use them only for block comments. And have a space between `~` and the operand. It's okay if you don't fix it for now but we should probably follow this in future commits.
@cbm755 @mtmiller Opinion?
.
Other than these things, +1 from me!

mtmiller:
@genuinelucifer Agree with your style comments (and keeping with `%%`, `%`, and `~` for Matlab compatibility).

I agree with the parentheses around the full conditional for all if and while statements, and parentheses around sub-expressions where it helps with clarity. For example I always use parentheses for compound expressions like `if ((nargin < 1) || (nargin > 2))` even if they are not strictly needed.

latot:
Hi!, well after talk lets use this format for now, i reformat the files (only related files i don't will touch other files here), if someone wanna recommend something here, this is always open.

Thx. Cya.

latot:
@genuinelucifer, if you like i can change part of the code to be more easy move it to classdef, do i do it?

genuinelucifer:
@latot about your last commit, why did u need to have a separate variable? `x` was never being changed and basically you are setting the new variable to `~isreal(x)`.... It did work for me without a separate variable.

latot:
Hi, yes it works in the two forms, i change it only because i don't like call one function more than 1 with the same var, i prefer store it and use that var instead recall the functions.
i did a little change in the last commit - rebased - maybe with it its more clear my intentions with that commit.

Thx. Cya.

genuinelucifer:
@latot I started to convert sym to classdef again and now I get your point. The changes that you made make it even better now as absolutely no calls to constructor are made within itself. Great catch! :+1: 
Also after your changes we don't need the `create_sym_with_s.m` anymore. We should probably drop it and move the code back to `sym.m`!

@cbm755 When can you review and possible merge this? I am basing my sym to classdef PR on this.

latot:
Hi, i should have merged upstream/master before you start working from here.. u.u
well, i merge it from here, and i do 2 little changes, restoring some lines and move create_sym_with_s file to the original place you should be able to merge this changes without start again.

Thx. Cya.

cbm755:
Ok, I will try to review.  But I really think this needs to be split up.  Can we leave this PR as-is and start to rebase bits of it for review/merge?

`@sym/private/check_assumptions.m`:  Could this be part of your other PR?


The move of assume  and assumeAlso to `inst/`, and the `norm_logic_strings` bit: can this be a separate PR?


`magic_double_str`: does different things than it used it.  Its docs still say "Recognize special double values."


`make_sym_matrix`: different behaviour for size below/above 20.  I think this should be a different PR/discussion.


`assert (nargin < 3, 'MatrixSymbol do not support assumptions')`
this needs a test


`symsnotfunc (x);  % Warning if you try make a sym with the same name of a system function.`
Nice idea but can be separate PR?


re: future classdef: maybe this doesn't directly call itself, but it certainly calls other functions which call the constructor.   I don't quite see the distinction.  Anyway, that doesn't effect this PR.

genuinelucifer:
> future classdef: maybe this doesn't directly call itself, but it certainly calls other functions which call the constructor. I don't quite see the distinction.

Well, currently there are 3 places where it does so and finally returns something which is not sym but either array/matrix/cell of syms. When we change sym to classdef those 3 places will issue an error and those things will have to be done via `sym.symarray` or whatever we rename the new static function too. This will be required (and is done on my PR) because from constructor of a classdef we cannot return anything except the object of that class.

latot:
Hi!, mmm:

```@sym/private/check_assumptions.m``` && ```norm_logic_strings``` &&  ```symsnotfunc (x);```
This 3 files i can split it, the only difference will be the bugs will be there, if i move the file 1 we in that time can't check assumptions, with the second if the user overload an expression with a assumption name sym will send a nice error, the last is more safe split in a new pr.

```make_sym_matrix```, you are right, i'll change it to can discuss it properly.

about the files i move from ```inst/@sym``` to ```inst```, will do any difference if we have a pr to only move 2 files?

The other points i need cover it, names, descriptions and tests.

Taking in consideration this points, are you sure you want me split the helpers in that way?

Thx. Cya.

cbm755:
Well, its more that it will take me *sooo* long to review otherwise.  And will probably need some rebasing anyway (so many commits here).   Do any of them have "Fixes #xyz" messages for example?

cbm755:
I'm not sure I approve of this change, may need some other discussion:
````
sym('1.1')
ans = (sym)

  11
  ──
  10

````

latot:
mm, what do you think sym should return?, because return a float i think is a very bad idea.....
Explaining a little that point:
Normally the expectation for the octsympy is can work with the best precision possible, if we want works with floats or doubles in first place we don't need this package, except if you need some function from here but in that case i think its better improve the float function of sympy instead mix the precision of sym with it.
One day i open a bug in octave, requesting some way to have a better precision, and the answer was the recommendation for use this package, i don't wanna say we always should do the common expectation, but at least with this we can have a perspective of other of what want or why others uses this package.
Now, if we check the sym function, the doubles section, we can see you did 2 functions to can have the better precision possible of it, so i think i can suppose your expectation from this function is be precise too.
For what reason we send an expression to the sym function?, i think basically 2 reasons, can use a specific function or get the exact expression we want, obvs with the limitations of the function, but for that you have the ```rationalapprox``` warning in doubles, so following the same logic, if we send ```sym('1.1')``` i think sym should return ```11/10```.
Is this useful?, in my personal experience yes, when i normally import data from other programs, but i have a super array with a lot of decimals, and obvs i need the exact expression, actually what we need to do in this cases?, if we have ```1.5937462092```, we manually with every expression need transform to ```sym(15937462092)/10e10```, obvs we can make a new program/script to do this, but that have sense for a user?

If i'm wrong in any point please say it, and if you think sym should return other type of data, what should be?

following SMT, it accepts char inputs to represent exact expressions, with vars and numbers:
https://es.mathworks.com/help/symbolic/sym.html

About the commits here, sincerely, its too hard split this in a rewrite of this type, if you rewrite a function to solve bugs, you will leave intentionally bugs everywhere? i think not, at least in the code i read in others prs, when we rewrite we try solve all possible with it, but that implies the rewrite don't have an order to solve it, because we don't are fixing the original function, we are rewriting it, as i can see i split some themes in the commits, but their structure don't is absolute, because if we found other type of problem how we are rewriting it so we can change anything to this can be better, and the next commits can rewrite the past commits.

In this case i think i can split some changes out of the sym function, specially in assumptions theme, but to can do that i'll need a new pr, and one commit for the entire sym function, and then split commits for assumptions, even the sym function have part of it, so about it self i can't split that part, only in assume, assumptions, and assumeAlso, even in it are some parts can be hard to split, because some rewrite sections solve it but the structure don't change with some features...

To review this i recommend don't do this from the git diff, or web, i think is better read directly the files and use logic with it, the main changes are in sym, assume and assumeAlso, others are minor changes for some concepts or some decisions we are arguing.

Thx. Cya.

cbm755:
Maybe we do want `sym('1.1')` to give rational (esp. if SMT does that).  I'm just saying its yet another change, that ideally would in principle should its own PR, with an addition to NEWS file, etc, etc.

cbm755:
I also took a look through the commit messages: many (most?) look something like:
````
other minor changes

other minor changes

other minor changes

other minor changes

fix nan

other fixes

other minor fix

fix make_sym_matrix

better error message

other fix
````
And note that is all one commit message.  This is not good enough.  It doesn't use the correct format (one line summary, followed by one paragraph of details, including explicit links to bugs fixed).  And I have no idea from reading what was fixed.

So there is some serious work to rebase this down to a bunch of sensible-looking, nicely formatted changes.  And if we're doing that anyway, why not split into more than one PR?

latot:
Hi, i know this commit are all mixed up, but for that reason i explain part of this above, this don't is exactly to solve bugs, is a rewrite, with what that implies, when we rewrite, we can do the first commit, then the second, and then third, but its very probable have future commits to rewrite the previous commits, its part of the process, for that reason the commits text can't say if you are really fixing something, and the structure of the file changes every time, so if you want check what does the commits you will start checking rewrites of rewrites, and that will be too confuse.

Only to remark this, a rewrite don't is the same as a fix, the concepts and how works file change, so sadly split the code its very hard because the bugs don't are fixed like line by line, or minor changes, its reconstructed to avoid that type of problems, but this means you can't extract/split the fix for  specific bug, because now is part of the function of the file, if you remove some feature that affect all others and the system can stop working, obvs are parts we can split in some commits, but i think split in multiples prs this can have a lot of problems.

Now, i'm asking to me a little how split this in multiples prs?

One way is provoke bugs intentionally in the code, so we can split this merge and check part by part, but in that time the upstream/master will be expose to the problems.

Other way is split part of this but with tests to intentionally to fail, and then check part by part and then merge all together.

Personally i don't like to much this options, in the first we will expose master (bad idea), in the second we will can't check if we do new changes it will pass or not.

Maybe the best to do is split this in related sym rewrite and assumptions.
But it will need an order, first sym and then assumptions, but probably are some changes we can't avoid, basically the related parts will need be in the first pr.

Other way, i can comment all and how the issues are fixed in this code, so with it you will can see the specific changes for that parts.

Thx. Cya.

cbm755:
Some bits are in, many trees have diverged, closing for now.  If there are still *simple* changes from this, they can be other small, *focused*, PRs.

Issue #539, opened 2016-09-06
add chain rule example (to syms?)


cbm755:
See #520, which gives an example (https://github.com/cbm755/octsympy/issues/520#issuecomment-244669492) of nesting functions using symfuns.  Useful to add this example to our docs somewhere.

Perhaps just in `syms` because some confusing stuff can happen there if you try to accomplish this in some other somewhat intuitive ways.

Issue #518, opened 2016-08-31
Creating symfun with `syms` can overwrite existing variables


latot:
Hi, this problem, basically we suppose a f(t) symfun, if we make a new symfun like g(f), f(t) its transformed to sym(f):

```
>> syms t f(t) g(t) r(f, g)
>> t
t = (sym) t
>> f
f = (sym) f
>> g
g = (sym) g
>> r
r(f, g) = (symfun) r(f, g)
```

Cya.

mtmiller:
According to https://www.mathworks.com/help/symbolic/syms.html, this looks like it's working correctly. When using the `syms` shortcut, any `f(x, y, z)` also automatically defines `x`, `y`, and `z` as independent variables.

latot:
if you write it in reverse order it works:

```
>> syms f(x, y) x(t) y(t)
>> f
f(x, y) = (symfun) f(x, y)
>> x
x(t) = (symfun) x(t)
>> y
y(t) = (symfun) y(t)
```

cbm755:
Oops, I think I discovered this while looking at one of your other bugs.  A @mtmiller says `syms` is a convenience function which just automates a bunch of stuff.  In this case, I think it has done exactly what we asked it to do.  And yes the order matters: it deals with them left to right.

I'd be surprised if SMT does any better, but it might do some other ill-defined behaviour.

latot:
mm, the description don't match with the help of syms, from the help:
```
     Symfuns represent abstract or concrete functions.  Abstract symfuns
     can be created with ‘syms’:
          syms f(x)
     If ‘x’ does not exist in the callers workspace, it is created as a
     *side effect* in that workspace.
```
as is explained if x exist shouldn't be created again, the other way is remove this section and explain the order  matters.

cbm755:
hmmm, possibly a problem if x already existed with assumptions then it would be overwritten.  I agree the docs are a bit misleading!

latot:
thinking, basically we have two options, first no overwrite the function if exist, and the other, overwrite it, in the last case i think should be a warning.

cbm755:
SMT also overwrites with no warning.

PR #596, opened 2016-10-04
Symfun fixes

cbm755:
@latot, can you review?  This is fixes for #518 and #539.


latot:
Hi, i think its fine.

PR #606, opened 2016-10-05
Rename pytave to native Python/C interface

### LINKED ISSUE #593: decision: rename `pytave` in code, docs, and interface
I would like to avoid exposing the name `pytave` to users, since I don't expect the project to ever be used or installed under that name. I vote for calling it `native` in the docs and in the `sympref` interface ("native" meaning it is linking to / invoking the Python runtime directly). Other suggestions?

mtmiller:
Excludes explicit references and comments about pytave as a project.
Fixes #593.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -53,8 +53,8 @@ octsympy 2.5.0-dev
 
   * `isfinite` behaves correctly for variables (and is documented).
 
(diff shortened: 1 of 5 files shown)


cbm755:
LGTM

PR #607, opened 2016-10-05
Fix typo in heaviside test printing an empty plot

mtmiller:
This looks like it was just a typo/thinko. But the end result for me was a file "skipping test, sympy too old.ps" appearing whenever I ran the test suite.


cbm755:
heh!

PR #608, opened 2016-10-05
[wip] sympref: document experimental "native" ipc option

### LINKED ISSUE #589: add pytave doc to sympref

mtmiller:
Small attempt to fix #589 (assuming that the rename to `native` is acceptable).

Should we also mention that `native` is the default if the pytave functions are found in the load path?

### DIFF
--- a/inst/sympref.m
+++ b/inst/sympref.m
@@ -106,6 +106,10 @@
 %% @itemize
 %% @item @code{sympref ipc popen2}: force popen2 choice (e.g.,
 %% on Matlab were it would not be the default).
+%% @item @code{sympref ipc native}: use native Python/C interface to
+%% interact directly with an embedded Python interpreter.
+%% This is highly experimental and requires functions provided by the
+%% "pytave" project which have not yet been merged into Octave.
 %% @item @code{sympref ipc system}: construct a long string of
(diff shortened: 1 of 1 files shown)


inst/sympref.m
```diff
@@ -106,6 +106,10 @@
 %% @itemize
 %% @item @code{sympref ipc popen2}: force popen2 choice (e.g.,
 %% on Matlab were it would not be the default).
+%% @item @code{sympref ipc native}: use native Python/C interface to
+%% interact directly with an embedded Python interpreter.
+%% This is highly experimental and requires functions provided by the
+%% "pytave" project which have not yet been merged into Octave.
```
cbm755:
````
``pytave''
````
i.e., LaTeX style backtics and apostrophes instead of double-quotes?  Sorry, can't easily check for myself which is usual for texinfo.

cbm755:
If we do a release soon (should) then I think I will not make it default, even if found.

cbm755:
Text itself LGTM

mtmiller:
> I will not make it default, even if found

Sounds good. I'll file an issue to keep track of that.

cbm755:
I'll fix the backtics post merge

Issue #614, opened 2016-10-14
Imag and real can't solve non-linear expressions


latot:
Hi, basically this:
```
octave:74> imag(exp(x*i))
ans = (sym)

     ⎛ ⅈ⋅x    -ⅈ⋅x⎞ 
  -ⅈ⋅⎝ℯ    - ℯ    ⎠ 
  ──────────────────
          2         

octave:75> real(exp(x*i))
ans = (sym)

   ⅈ⋅x    -ⅈ⋅x
  ℯ      ℯ    
  ──── + ─────
   2       2  
```

Thx. Cya.

PR #615, opened 2016-10-14
fix imag and real

latot:
Fixes https://github.com/cbm755/octsympy/issues/614


latot:
to see the expected return check the new tests.
(this behavior is more close to SMT)

cbm755:
Nice!

Issue #629, opened 2016-10-18
zeta function is incomplete


latot:
Hi, well the implementation of the zeta function is incomplete:

```
octave:114> zeta(sym(1), 1)
error: Invalid call to zeta.  Correct usage is:

 -- Method on @sym: zeta (X)

Additional help for built-in functions and operators is
available in the online version of the manual.  Use the command
'doc <topic>' to search the manual index.

Help and information about Octave is also available on the WWW
at http://www.octave.org and via the help@octave.org
mailing list.
```

```
octave:116> python_cmd('return zeta(1/2, 0)')
ans = (sym) ζ(0.5, 0)
```

Probable can good idea check if other functions are incomplete or similars.

Thx. Cya.

cbm755:
Matlab's SMT seems to use `zeta(n, z)` for derivatives of zeta function [1].  But SymPy is Hurwitz zeta.  Do these agree for `n` integer?

[1] https://www.mathworks.com/help/symbolic/zeta.html

cbm755:
Matlab's SMT seems to use `zeta(n, z)` for derivatives of zeta function [1].  But SymPy is Hurwitz zeta.  I haven't checked; are these two things actually the same thing for `n` integer?

[1] https://www.mathworks.com/help/symbolic/zeta.html

cbm755:
stupid browser, sorry for dupe

latot:
mm, searching here i found a bug in sympy, i'll send it, but for now i propose this:
```
zeta(z) -> normal z function
zeta(n, z) -> n derivative of zeta in z
zeta(n, z, i) -> n derivative of hurwitz zeta
```
What do you think?

latot:
answering your question, don't are equals.

cbm755:
How are you going to do the derivatives in SymPy?

latot:
mm, i was thinking calc the derivatives manually while we wait Sympy to support the n derivative, but sadly i found a super bug, so, we can write the ideal code, and the manual code, but it don't will works until sympy fix at least this issue, so for now we can lock ```n ~= 0``` and use it...
or we can just leave python send the error.
https://github.com/sympy/sympy/issues/11802

Thx. Cya.

PR #630, opened 2016-10-21
improve zeta(a, b)

latot:
Fixes https://github.com/cbm755/octsympy/issues/629


latot:
Hi i think this should fix the issue.

Issue #636, opened 2016-10-26
We should return the linerror


latot:
Hi, well with the changes in python_cmd and the error handling now is more easy debug all of this, but now i think we need send what line have the problem, if we get, "occurred at line 58", now we must start counting all the lines until found the line, take in consideration when we use helpers this is hard, we need start counting the lines added in all files....

Well, basically add this to python_cmd:

```
disp (strtrim (cmd {A {3}}));
```

Thx. Cya.

PR #637, opened 2016-10-26
Print linerror

latot:
Maybe if you like, set a `sympref debug 1`?

Fixes https://github.com/cbm755/octsympy/issues/636

Thx. Cya.

### DIFF
--- a/inst/python_cmd.m
+++ b/inst/python_cmd.m
@@ -170,8 +170,9 @@ function varargout = python_cmd(cmd, varargin)
   %% Error reporting
   % ipc drivers are supposed to give back these specially formatting error strings
   if (~isempty(A) && ischar(A{1}) && strcmp(A{1}, 'COMMAND_ERROR_PYTHON'))
-    errlineno = A{3} - db.prelines - LinesBeforeCmdBlock;
-    error(sprintf('Python exception: %s\n    occurred at line %d of the Python code block:\n    %s', A{2}, errlineno, strtrim (cmd {A {3}})));
+    errcmdlineno = A{3} - db.prelines;
+    errlineno = errcmdlineno - LinesBeforeCmdBlock;
+    error(sprintf('Python exception: %s\n    occurred at line %d of the Python code block:\n    %s', A{2}, errlineno, strtrim (cmd {errcmdlineno})));
   elseif (~isempty(A) && ischar(A{1}) && strcmp(A{1}, 'INTERNAL_PYTHON_ERROR'))
     error(sprintf('Python exception: %s\n    occurred %s', A{3}, A{2}));
   end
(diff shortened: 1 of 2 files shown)


latot:
mm, this for a octave bug, is fixed in the actual hg version.

cbm755:
Seems like a good idea.  Please do `strtrim (cmd{A{3}})));`  (i.e., some whitespace problems in the push).

I don't think its an Octave bug: I think its just that on other ipc modes, there are some lines added so you cannot look look it up at index `A{3}`.  Does it work if you use `errlineno` instead?

latot:
Hi, i'm already using ```strtrim```, so should be fine, and you are right, don't is a octave bug, i was testing this bad (i was running python_cmd tests instead of sympref), in popen2 it works in some cases (coincidences), so now works.

Thx. Cya.

inst/python_cmd.m
```diff
@@ -170,8 +170,9 @@
   %% Error reporting
   % ipc drivers are supposed to give back these specially formatting error strings
   if (~isempty(A) && ischar(A{1}) && strcmp(A{1}, 'COMMAND_ERROR_PYTHON'))
-    errlineno = A{3} - db.prelines - LinesBeforeCmdBlock;
-    error(sprintf('Python exception: %s\n    occurred at line %d of the Python code block', A{2}, errlineno));
+    errcmdlineno = A{3} - db.prelines;
+    errlineno = errcmdlineno - LinesBeforeCmdBlock;
+    error(sprintf('Python exception: %s\n    occurred at line %d of the Python code block:\n    %s', A{2}, errlineno, strtrim (cmd {errcmdlineno})));
```
cbm755: my comment was about whitespace `cmd{...}` (no spaces with indexing)

latot:
!! Okis, now i fixed it.

latot:
Travis timeout, but in Python3 (stalled) pass sympref with the relevant tests.

cbm755:
I restarted the CI run.

Probably this should have a test or two.  IIRC, there are tests for such things in `sympref`.

latot:
Okidokis, Tests added!

cbm755:
But these tests do not test the new feature!

I'm trying to see why we need both `errcmdlineno` and `errlineno`, why they are not the same.  Figured maybe a test would help.

Also some commit should have a "Fixes #636" in the commit message.

latot:
Hi, i think the test are fine, because the idea of this is send what line has the error:
```
python_cmd('pass;raise;pass')
```
this should return ```raise``` and some other text, this is what test do.

```errcmdlineno``` don't are equal to ```errlineno``` because, the last one try found the error counting from the input data of python_cmd (```def _fcn(_ins):```), so we need remove some lines and remove the start lines of the default cmd, while in the first, we are trying to get the error not in the input of python_cmd, is in the complete cmd. I think i'm explaining this too bad, so lets use an example:

```
cmd = [{'Hi'
             'O'}
             some user input text
            {'Bye'
             'J'}]
```
well, if i send this text, ```{'pass' ; 'Meh' ; 'pass'}```, in what position is ```Meh```?, is the line number 2, even if is replaced in cmd, should be 4(input text) - 2 (pre lines), this is errlineno, but in what line of cmd is?, here is 4, and this is the errcmdlineno, and as you can see both are related, basically this is.

cbm755:
Thanks, this looks good to me.  I want to add a similar test to the system ipc tests, but I can do that.

cbm755:
I added 6c61547366e1fbce8be20d7bf048737f4246ec48 just so it gets tested with `system` as well.

Issue #639, opened 2016-10-28
Merge (uni,bin,tri)op_helper into one


latot:
Hi, well checking this i found actually octsympy is using 3 helpers for the same operation (and 1 deprecated)....

So, i think would be great merge this into one file and clean this.

Thx. Cya.

Issue #635, opened 2016-10-26
Round inconsistent in py2/py3, sound return a symbolic Integer?


latot:
Hi all, well this happen because round don't return a Sympy object, return an int, and for other weird reasons:

```
octave:7> a=sym([1 2])
a = (sym) [1  2]  (1×2 matrix)
octave:8> round(a)
ans = (sym) [1.0  2.0]  (1×2 matrix)
```

Now the weird things:

```
octave:9> python_cmd('return str(type(round(Integer("3"))))')
warning: called from
    python_cmd at line 186 column 5
ans = <type 'float'>
```

```
>>> str(type(round(Integer("3"))))
"<class 'int'>"
```

For some reason the returned object is different from python to octave.........

Well, Thx. Cya.

mtmiller:
Hm? Are you sure you're comparing the right version of Python?

On my system, with Python 2.7

    >> a = sym ([1, 2])
    >> round (a)
    ans = (sym) [1.0  2.0]  (1×2 matrix)

And with Python 3.5

    >> a = sym ([1, 2])
    >> round (a)
    ans = (sym) [1  2]  (1×2 matrix)

And running the same commands in Python 2 vs 3 confirms the difference.

See also http://python3porting.com/differences.html#rounding-behavior.

latot:
Hi, yes, i test this in octave and python 3.4, so maybe exist a difference between 3.4 and 3.5 too.

Thx. Cya.

mtmiller:
No, rounding in Python 3.4 is the same as in 3.5. I think you're using Python 2 with Octave, can you double check that? What does `x = python_cmd("return sys.version")` give you?

I think the more interesting difference in rounding is the following.

Python 2

    >> round (sym ([2:6]) ./ 2)
    ans = (sym) [1.0  2.0  2.0  3.0  3.0]  (1×5 matrix)

Python 3

    >> round (sym ([2:6]) ./ 2)
    ans = (sym) [1  2  2  2  3]  (1×5 matrix)

Not just the types of the values, but the rounding direction has changed. I would think we want to match Octave rounding behavior, so that `round(2.5) == round(sym(2.5))` is always true.

latot:
:O 

You are right i'm using python 2, i forgot i restart octave with python 2, i was using python 3.
I'm agree to match the behavior with octave, and we need add this to new tests.

Thx. Cya.

cbm755:
[...]

PR #640, opened 2016-10-28
Improve op_helper - Clean helpers

latot:
Fixes https://github.com/cbm755/octsympy/issues/639

Hi, well, now this is a little clean in the helpers, and a generalization of it, so we will no need rewrite new functions for this, this code practically don't affect performance, with relative low interaction (less than 1000 operations) are very similar, some times one is faster other moments the other, while with a 10000 operations we finally can see we have an approximation of 6% of less performance.

Other point is with this is defined the operation element by element to any function, and if we found some problem we only need fix one file instead of 3.

I write this code in this way thinking in Tensor.

Thx. Cya.


latot:
I change the round function, because in this issue https://github.com/cbm755/octsympy/issues/635, we found two problems, and this change expose one of it, so its necessary to pass the tests, and this fix don't solve the the issue completely.

inst/@sym/private/op_helper.m
```diff
@@ -44,19 +43,20 @@
...
-          '    return _op(x, y),' ];
+          'l = (1, 1)'
+          'for i in range(len(_ins)):'
+          '    if isinstance(_ins[i], MatrixBase):'
+          '        if _ins[i].shape != (1, 1):'
+          '            l = _ins[i].shape'
+          '        else:'
+          '            assert l == _ins[i].shape, "Matrix must have equal sizes"'
+          '    else:'
+          '        _ins[i] = Matrix([_ins[i]])'
```
cbm755: I think you should first convert it to matrix, then check the size is either same or (1,1).

inst/@sym/private/op_helper.m
```diff
@@ -44,19 +43,20 @@
...
+          'base = zeros(*l)'
+          'for i in itertools.product(*map(Range, l)):'
+          '    base.__setitem__(i, _op(*[k[0] if k.shape == (1, 1) else k.__getitem__(i) for k in _ins]))'
+          'return base,' ];
 
-  z = python_cmd (cmd, sym(x), sym(y));
+  z = python_cmd (cmd, sym(varargin){:});
```
cbm755: Can't change indexing on Matlab (I think).  Also, IIRC, we might deprecate `sym(cellarray)`.  Can you use `cellfun` or something to do it?

inst/@sym/private/op_helper.m
```diff
@@ -44,19 +43,20 @@
...
+          'for i in range(len(_ins)):'
+          '    if isinstance(_ins[i], MatrixBase):'
+          '        if _ins[i].shape != (1, 1):'
+          '            l = _ins[i].shape'
+          '        else:'
+          '            assert l == _ins[i].shape, "Matrix must have equal sizes"'
+          '    else:'
+          '        _ins[i] = Matrix([_ins[i]])'
+          'base = zeros(*l)'
+          'for i in itertools.product(*map(Range, l)):'
```
cbm755: Can this just be:
````
for i in range(0, len(<something>)):
    base[i] = ...
````

File: inst/@sym/abs.m [hunk 4]
```diff
@@ -40,7 +40,7 @@
   if (nargin ~= 1)
     print_usage ();
   end
-  y = uniop_helper (x, 'Abs');
+  y = op_helper ('Abs', x);
```
cbm755: For many of these files, the change needs to be made in `generate_functions.py` (see note in the help text).

inst/@sym/private/op_helper.m
cbm755: Sorry I meant "can't chain indexing in Matlab".

inst/@sym/private/op_helper.m
latot: Hi, the send line works, while the first sadly not, that only will travel in a 1 dim matrix.
The idea of the line is travel in all values of the matrix, while the matrix can have any number of dims.

cbm755:
> only will travel in a 1 dim matrix

Seems to work for me:
````
In [21]: A = Matrix([[x,2,3],[3,4,5]])

In [22]: B = zeros(*A.shape)

In [23]: for i in range(0, len(A)):
   ....:     B[i] = 2*A[i]
   ....:     

In [24]: B
Out[24]: 
Matrix([
[2*x, 4,  6],
[  6, 8, 10]])
````

File: inst/@sym/private/op_helper.m (hunk 2 above)
Comment (latot) [inline inst/@sym/private/op_helper.m]: Hi, i split the chain, about cellfun sadly i can't use it, at least in a right way, because it converts the cells to matrix! and with it we can't unpack it to a function. Obvs i could researched bad, if you know some way please tell me and i'll update this.
In some way i prefer do this in other pr, because i think we should update all the code with that instead only a few lines.

If don't exist, maybe use a helper instead?

Thx. Cya.

File: inst/@sym/eq.m [hunk 5]
```diff
@@ -88,7 +88,7 @@
     print_usage ();
   end
 
-  t = ineq_helper('[donotuse]', 'Eq', x, y);
+  t = ineq_helper('[donotuse]', 'Eq', sym(x), sym(y));
```
cbm755: this change is unrelated to this PR?

inst/@sym/rewrite.m
```diff
@@ -109,7 +109,7 @@
 
   F = python_cmd(cmd, sym(f), how);
 
-  % maintainer note: binop_helper would sym(how)
+  % maintainer note: op_helper would sym(how)
```
cbm755: ... for example, this is not true any more: maybe just say "op_helper might sym(how)"

latot:
You are right, sorry i misread what you mean, now i change it.

With last change now the helper support more formats, not only sym objects, its useful to can simplify other functions like sym, so now we don't transform the objects to sym in the helper and we need declare it from the functions, this is why i update the functions with ineq_helper (ineq_helper call op_helper).

Thx. Cya.

inst/@sym/isempty.m
```diff
@@ -60,7 +60,6 @@
 %! a = sym ([1 2]);
 %!assert (~isempty (sym (1)))
 %!assert (isempty (sym (se)))
-%!assert (isempty (se == []))
```
cbm755: WTF?

inst/@sym/laguerreL.m
```diff
@@ -63,10 +63,9 @@
 function L = laguerreL(n, alpha, x)
 
   if (nargin == 2)
-    x = alpha;
```
cbm755: I want this back, too hard to read the code otherwise.

inst/@sym/lambertw.m
```diff
@@ -52,17 +52,12 @@
...
-    x = sym(x);
-    k = sym(k);
-    W = binop_helper (x, k, 'LambertW');
-  else
+function W = lambertw(varargin)
+  if (nargin > 2)
     print_usage ();
   end
+  varargin = sym(varargin);
+  W = op_helper ('lambda a: LambertW(*reversed(a))', varargin);
```
cbm755: nice!

inst/@sym/limit.m
```diff
@@ -83,14 +83,7 @@
...
 
-  cmd = { '(f, x, a, pdir) = _ins'
-          '# note, not MatrixExpr'
-          'if isinstance(f, sp.MatrixBase):'
-          '    g = f.applyfunc(lambda b: b.limit(x, a, dir=pdir))'
-          'else:'
-          '    g = f.limit(x, a, dir=pdir)'
-          'return g,' };
-  L = python_cmd (cmd, sym(f), sym(x), sym(a), pdir);
+  L = op_helper('lambda a, b, c, d: a.limit(b, c, d)', sym(f), sym(x), sym(a), pdir);
```
cbm755: this must be another PR

cbm755:
I started reviewed agian, but there is too much new stuff here!!  This was a nice simple PR that I as looking forward to merging.  Please  keep it that way.  

EACH PR DOES ONE SIMPLE THING!  That way, it will get reviewed and merged.  This one will sit here until it looks simple again.

latot:
Okidokis!, splitting....

inst/@sym/isempty.m
latot: that don't have any sense..., because [] is equal to []..., why it should return empty? (se is [])

latot:
Hi!, i split this, when this will be merged i'll send the next pr with the changes.
I add documentation to the helper.

Thx. Cya.

inst/@sym/isempty.m
cbm755: elementwise comparision I suppose.  But even if the test is wrong, its not this PRs position to change it.  If the test is wrong, do another PR.

Unless this PR changes behaviour in some way that the test used to pass and now it fails...  But if so I'm worried because this PR is just a code refactor.

inst/@sym/private/op_helper.m
```diff
@@ -0,0 +1,89 @@
...
+    %assert strncmp(scalar_fcn_str, 'def ', 4)
+    cmd = scalar_fcn;
+  else
+    cmd = {['_op = ' scalar_fcn]};
+  end
+
+  % note: cmd is already cell array, hence [ concatenates with it
+  cmd = [ cmd
+          'q = Matrix([0])'
+          'for i in _ins:'
```
cbm755: maybe `for A in _ins:` better  (`i` is for integer)

cbm755:
Sorry for losing temper earlier: had only a few mins to review and was sad that it looked so different.

Now looks much better!  Any idea why lambertw is failing?

latot:
Hi!, i restore lambertw function but i forgot update the function, now should works, i'm actually with a few mins with internet so i can't check if travis pass, my hope is it works.

Thx. Cya.

inst/@sym/private/op_helper.m
latot: Okidokis

cbm755:
Ok, this looks good.  Feel free to prepare those other fixes as new PRs but let's try to keep each self-contained.

inst/@sym/isempty.m
cbm755: Oops, I missed that this was still here before I merged :(

I will continue this discussion at #650

PR #641, opened 2016-10-30
Added guide to install on Ubuntu

mohtamohit:
There are packages like liboctave-dev whose mention wasn't there in the installation guide. Neither were the commands handy for a noob to set things easily, for a noob who hasn't used Octave ever before. So, just added a few more details to help a beginner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,9 @@ How to install on Ubuntu
 -------------------------
 
 1. If you've not installed Octave yet, use `sudo apt-get install octave`.
-2. Fetch the Debian package liboctave-dev to get the mkoctfile command. Use `sudo apt-get install liboctave-dev` in terminal.
-3. If Python and SymPy not already installed, use `sudo apt-get install python` && `sudo apt-get install python-sympy`.
-4. In Octave prompt, type `pkg install -forge symbolic` followed by last two steps of above-menioned installation guide.
+2. Fetch the Debian package liboctave-dev to get the mkoctfile command. Use `sudo apt-get install liboctave-dev`.
+3. If you've not installed python-sympy yet, use `sudo apt-get install python-sympy`.
+4. In Octave prompt, type `pkg install -forge symbolic` followed by last two steps of previous section.
 
 How to Install on Windows
 -------------------------
--- a/README.md
+++ b/README.md
@@ -22,6 +22,16 @@ Status
 
 "Beta" quality at best!  Contributions welcome.
 
+How to install on Ubuntu
+-------------------------
+
+1. Install Octave first. (Type 'sudo apt-get install octave' in terminal).
+2. Install the Debian package 'liboctave-dev' to get the mkoctfile command. (Type 'sudo apt-get install liboctave-dev' in terminal).
+3. Install other dependecies too like Python and sympy. ('sudo apt-get install python' && 'sudo apt-get install python-sympy').
+4. Now, open Octave and enter 'pkg install -forge symbolic' in its command window.
+5. At Octave prompt, type `pkg load octsympy`.
+6. At Octave prompt, type `syms x`, then `f = (sin(x/2))^3`,
+   `diff(f,x)`, etc.
 
 
 How to Install
--- a/README.md
+++ b/README.md
@@ -42,10 +42,10 @@ How to Install
 How to install on Ubuntu
 -------------------------
 
-1. If Octave not installed yet, use 'sudo apt-get install octave'.
-2. Fetch the Debian package 'liboctave-dev' to get the mkoctfile command. Use 'sudo apt-get install liboctave-dev' in terminal.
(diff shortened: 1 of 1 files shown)


cbm755:
Seems like a good idea!  Is the Symbolic Package packaged for recent Ubuntu?  Maybe we should first tell people to just `apt-get install octave-symbolic`?    And include your instructions for "installing a more recent version" or something like that?

mtmiller:
A couple nitpicks.

* The `liboctave-dev` package is only needed for `mkoctfile` But octsympy doesn't actually have any compiled code. Might we not get rid of the configure script (and the src subdirectory) entirely, and not need any checks in place for a compiler and no dependency on `mkoctfile`?
* Do we really need to explain how to install Octave? Point to http://wiki.octave.org instead?
* Factor out or skip the parts that are common? Why re-state the parts that are already in the section **How to Install**?

In short, I think it's fine to document something like "use `apt install octave octave-symbolic` if you are happy with the version Ubuntu provides, use `apt install octave` and `pkg install -forge symbolic` if not". And maybe requiring `liboctave-dev` is a bug that we can fix easily for the next release.

mohtamohit:
@mtmiller Getting rid of configure script (and src subdirectory) seems to be a better idea. And, it was a one-liner to install octave so not a big deal, I think. Why redirect someone to other pages when all the information that needs is a one-liner. I have updated PR again and factored out the parts that were re-stated.
@cbm755 Is Symbolic package packaged for older Ubuntu versions as well?

mtmiller:
That's fine, like I said, nitpicking. Your changes look better, thanks for moving it below the main installation instructions.

Ubuntu packages for symbolic: http://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=Octave-symbolic&searchon=names

Versions 1.x are not based on SymPy and not recommended. But if a user is on 16.04 or newer, they can use the apt package.

README.md
```diff
@@ -41,7 +39,13 @@ How to Install
 6. At Octave prompt, type `syms x`, then `f = (sin(x/2))^3`,
    `diff(f,x)`, etc.
 
+How to install on Ubuntu
+-------------------------
 
+1. If Octave not installed yet, use 'sudo apt-get install octave'.
+2. Fetch the Debian package 'liboctave-dev' to get the mkoctfile command. Use 'sudo apt-get install liboctave-dev' in terminal.
+3. If Python and SymPy not already installed, use 'sudo apt-get install python' && 'sudo apt-get install python-sympy'.
+4. In Octave promt, type 'pkg install -forge symbolic' followed by last two steps of above-menioned installation guide.
```
genuinelucifer: Just minor spell check "prompt"!

README.md
```diff
@@ -41,7 +39,13 @@ How to Install
 6. At Octave prompt, type `syms x`, then `f = (sin(x/2))^3`,
    `diff(f,x)`, etc.
 
+How to install on Ubuntu
+-------------------------
 
+1. If Octave not installed yet, use 'sudo apt-get install octave'.
+2. Fetch the Debian package 'liboctave-dev' to get the mkoctfile command. Use 'sudo apt-get install liboctave-dev' in terminal.
+3. If Python and SymPy not already installed, use 'sudo apt-get install python' && 'sudo apt-get install python-sympy'.
```
genuinelucifer: Not important but how about using verbs? :P
"are installed" and "is installed" in 1st step?

mohtamohit: Corrected :P

README.md
```diff
@@ -41,7 +39,13 @@ How to Install
 6. At Octave prompt, type `syms x`, then `f = (sin(x/2))^3`,
    `diff(f,x)`, etc.
 
+How to install on Ubuntu
+-------------------------
 
+1. If you've not installed Octave yet, use `sudo apt-get install octave`.
+2. Fetch the Debian package liboctave-dev to get the mkoctfile command. Use `sudo apt-get install liboctave-dev` in terminal.
+3. If Python and SymPy not already installed, use `sudo apt-get install python` && `sudo apt-get install python-sympy`.
```
mtmiller: Minor nit: this can be simplified to `sudo apt-get install python-sympy`, the python package is pulled in automatically.

README.md
```diff
@@ -41,7 +39,13 @@ How to Install
...
+4. In Octave prompt, type `pkg install -forge symbolic` followed by last two steps of above-menioned installation guide.
```
mtmiller: How about "previous section" instead of "above-menioned [sic] installation guide"?

mohtamohit: Updated it :)

README.md
mohtamohit: Done with the changes. :)

mtmiller:
👍

cbm755:
Thank you @mohtamohit and other for review, merging

PR #642, opened 2016-10-30
Make use of SymPy round for Python 2/3 consistency

### LINKED ISSUE #635: Round inconsistent in py2/py3, sound return a symbolic Integer?

mtmiller:
Ensure result is an integer and rounding direction is consistent with
Octave. Fixes #635.

### DIFF
--- a/inst/@sym/round.m
+++ b/inst/@sym/round.m
@@ -25,7 +25,7 @@
 %% @example
 %% @group
 %% y = round(sym(-27)/10)
-%%   @result{} y = -3
+%%   @result{} y = (sym) -3
 %% @end group
 %% @end example
 %%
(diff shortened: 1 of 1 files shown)


cbm755:
There is a doctest that needs to change.

I would have added a new test rather than changing and old one, but that's a nitpick, I don't really care.

mtmiller:
Thanks for the feedback, I updated the doctest, restored the original unit test (I agree, more tests is always better), and squashed.

Issue #611, opened 2016-10-07
do not automatically use experimental native ipc driver by default if functions found in load path


mtmiller:
For the next release, and however many releases until support stablizes, the experimental `native` ipc mechanism should not be used by default if the pytave project happens to be installed or in the load path. It should only be used if the user explicitly selects it with `sympref`.

Just a reminder to change the default ipc detection so that it does not select `native` if the functions are found in the load path.

PR #645, opened 2016-10-31
Fixes for `sympref ipc`

mtmiller:
Fix for #611, and a small improvement to the `sympref ipc` warning message when an invalid argument is passed in.


cbm755:
+1

PR #649, opened 2016-11-01
Fix round to do the right thing on complex numbers

### LINKED ISSUE #643: Make round() work with complex values
While fixing #635, I discovered that `round(sym("5j/2"))` fails. This should return the same value as Octave `round(5j/2)` or SymPy `(5*I/2).round()`.

mtmiller:
Add conditional logic to round all Number values to the nearest Integer,
but still allow round to operate normally on expressions such as complex
numbers. Fixes #643.

### DIFF
--- a/inst/@sym/round.m
+++ b/inst/@sym/round.m
@@ -37,7 +37,7 @@ function y = round(x)
   if (nargin ~= 1)
     print_usage ();
   end
-  y = elementwise_op ('lambda a: Integer(a.round())', x);
+  y = elementwise_op ('lambda a: Integer(a.round()) if isinstance(a, Number) else a.round()', x);
 end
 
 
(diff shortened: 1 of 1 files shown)


cbm755:
Can I get a test please? ;-)

mtmiller:
Absolutely, rebased to new `op_helper` interface and added tests.

cbm755:
Just FYI: not a fan of expressions inside quotes inside sym. 'sym(4)/5 etc' is preferred.  Sort of surprised that works on Python 2 for example.  Maybe it doesn't as we only test it rounds correctly...

cbm755:
Actually maybe I won't merge yet: i don't have a great track record this morning!

Also I'm thinking new commits should just use regular Octave style (except maybe for sym() as it is so commonly inside brackets)

mtmiller:
Sounds good, I can rewrite.

So to be clear, you want to see something like `x = sym(5) / 3 - sym(4j) / 7;` instead? And spaces after function names that are not `sym`, as in `f1 = round (x);`?

cbm755:
almost, `sym(5)/3 - sym(4j)/7`, no spaces around `/`

mtmiller:
I will fix this up and rebase onto master after #656 gets merged.

mtmiller:
Ok, this is rebased to `elementwise_op` and coding style in tests is fixed.

cbm755:
Awesome, thanks for your patience!

Issue #650, opened 2016-11-01
regression: op_helper with empty args: `sym([]) == sym([])` should be empty 


cbm755:
It was recently changed by #640.  I accidentally merged without tracking this down.

cbm755:
Current behaviour in master:
````
octave:9> [] == []
ans = [](0x0)
octave:10> sym([]) == sym([])
ans = (sym) True
````

But previous behaviour, e.g., before merging #640, gave:
````
octave:5> sym([]) == sym([])
ans = (sym) []  (empty 0×0 matrix)
````
and in fact this was unit tested :(

cbm755:
I think the new `op_helper` fails in some way if all (some?) args are empty.  For example, after the first loop, `q` is still the `Matrix([0])` but I guess it should be the empty matrix?

@latot, can you try to fix this and restore the removed test?

cbm755:
This needs to work too:
````
>> sym(zeros(3,0)) == zeros(3,0)
ans = (sym) []  (empty 3×0 matrix)
````
and should be added to the tests  (it currently gives `True`)

latot:
Hi, i'll send a patch, i'm thinking a little thing, some functions need use [] as element but others as matrix, so i need found a way to balance this.

cbm755:
Would this help?

1.  Loop `_ins`, looking for the maximum shape.  (and check that all are same size or scalar).
2.  Create `q` after the loop, based on the maximum shape found.

PR #653, opened 2016-11-02
[WIP] elementwise_op - Support Matrix as Elements

latot:
Fixes https://github.com/cbm755/octsympy/issues/650

Hi, well here the fix, now why i add list?, is because some functions need works with strings in _op, and if we use the normal matrix the strings are converted to symbols....
So list are ideal to handle this type of things.

Thx. Cya.


latot:
Hi, this is relative ready, to simplify read the code, i create 5 helper functions, i think can be useful in other functions too, do you prefer move it to python_headers?
Sadly i should bring back the itertools function, because the actual index method only works in Matrix of sympy, and some types of data can't be stored there, like strings.
I know is seems a lot of more code, but i think don't will affect the performance, because in the most of cases the loops only will be max 2, so should be fast.

Thx. Cya.

cbm755:
This looks more complicated than I was expecting.  I'll look closer later.

latot:
again working, i move some functions and now a lot of fails...

latot:
Ufff, finally, i want explain a little why all this changes, in first place send ```[]``` as matrix present a problem, some functions need use it as element, but actually we don't have a way to support this, because we can't do in octave a matrix of matrix, octave join one each other and again 1 big matrix, so thinking a solution i found we can use cells to handle this, and a nice part cells are transformed to list in python, that means we can send chars directly, this last part is because if we send chars in a matrix, the matrix function convert the chars to symbols..., so to follow the same logic/behavior of matrix, i add lists to support, the idea is basically can send elements and travel the list instead of the element of the element (don't travel the matrix ```[Matrix([1])]```), so with it we can send, matrix, chars, etc. and use it directly in the function.
One of the reasons of why i include cells/lists in the format to travel, is because depending of the input, when _op found a list it will use it in that way... (```['a']```) instead of use only the expression we want (this happen in we send cells directly to python_cmd), and we must start traveling the list to found the element, other point is, what happen if we need send more than one expression in cells? (```{'a', 'b'}```), in that case we will break the concept and behavior of the function, like non-normalize it, and i think that is a bad idea too, so concluding, i include list to travel in the same way as the matrix and we can skip defining new behaviors etc.

Well, you know, recommendations are welcome!

Thx. Cya.

cbm755:
Hmmm, but why we need to do all that?  Only thing elementwise_op should be getting is (a) matrices, (b) scalars and (c) strings (strings are new after your PR #640).

> we use the normal matrix the strings are converted to symbols

Hmmm, I don't see this this in #657: it which passes a string.  Everything that is not a MatrixBase get's treated as a scalar (I think!)

latot:
Hi!, sorry i explain a lot of things but i forgot say the reasons, or why this, so please let me start with this point.
Do you remember when in op_helper pr i send a lot of thigs? and you request me simplify it?
Well, in that time i was inspecting the ```@sym``` files to know what the project need in this function, obvs in this context, element by element. There i found some functions need works some matrix as elements instead of a matrix, for that reason the actual elementwise_op use empty matrix as element, obvs to can pass the actual tests exist a easy fix, but if we only do that we will disable some functionalities for other functions, so trying to cover that is this pr, can fix the empty set, and don't loose the features to can improve in the right way the other functions.

So to handle this, this pr cover two things, first use all matrix with the element by element funtion, and cells (transformed to lists), will use their content as elements, without travel in, ```[Matrix([])]```, this list will return the matrix instead the content of the matrix.

Here i think a ideal example is the limit function, because it uses the empty matrix as a element instead of a matrix.

2 Notes, when i was writing this function i always keep in mind we need some feature like this to cover other functions; Second, i need send a probably last commit, because i found a little problem and i'm writing from a external pc.

If something still don't is clear, feel free to ask.

Thx. Cya.

cbm755:
Ok, I think I see: you are trying to add new features.  Whereas in #665 I'm just trying to fix the regression without adding new features.

The empty matrix in `limit` looks like a bug to me (#669).

I think I plan to merge #665.  (Of course we can still look at improving `elementwise_op` if there is strong case).

cbm755:
I've merged my changes.   I understand elementwise_op could be improved further.  Tensors is one thing.

But also keep in mind that not *every* function that acts elementwise *must* use this!  Its ok if some methods have their own implementations: that could still be easier to maintain than significantly more complicated `elementwise_op` helper.

cbm755:
Hopefully the important bits of this are in.

PR #659, opened 2016-11-02
adopt the maintainers Makefile

### LINKED ISSUE #413: drop configure script in favour of run-time checks
I'm considering dumping all the autoconf stuff from Symbolic: we have
nothing to compile, all its used for is checking that Python and SymPy
are installed.  And it just seems to get in the way on Windows (#371) and
(apparently now) MacOS (#412).

@genuinelucifer 

@carandraug you added this originally and it certainly seemed like a good idea on GNU/Linux.  But now I'm not so sure...  Thoughts?

cbm755:
This is WIP for #254.

TODO list:

- [x] port the matlab package stuff, fixes #413.
- [x] careful, matlab package has different name
- [x] make testing work
- [x] make doctests work
- [x] ~~~possibly squash commits~~~ (its not so bad)
- [x] ~~~check if other packages have more standard names for make targets and adopt those~~~ meh future work
    - [x] ~~~also possibly do same in octave-doctest project~~~ meh not this project
- [x] absorb make_release script
- [x] windows bundle script: move to util, leave for future improvement?


Makefile
```diff
@@ -0,0 +1,131 @@
...
+	@#rm -f fntests.log
+	rm -f octsympy_tests.log
+
+test:
+	@echo "Testing package in GNU Octave ..."
+	@$(OCTAVE) --no-gui --silent --path "${PWD}/inst" \
+		--eval "set (0, 'defaultfigurevisible', 'off'); \
+		 octsympy_tests; \
+		 sympref reset"
+	@! grep '!!!!! test failed' octsympy_tests.log
```
mtmiller: Why not `anyfail = octsympy_tests;` and `exit (anyfail);` here?

cbm755: Could be!  I just copy-pasta from Doctest and/or Interval ;-)  I'll look into it.

Makefile
```diff
@@ -0,0 +1,140 @@
...
+	git archive --output="$@" --prefix=${PACKAGE}-${VERSION}/ HEAD
+	tar --delete --file "$@" ${PACKAGE}-${VERSION}/README.matlab.md
+	tar --delete --file "$@" ${PACKAGE}-${VERSION}/HOWTO-release.md
+	tar --delete --file "$@" ${PACKAGE}-${VERSION}/README.bundled.md
+	tar --delete --file "$@" ${PACKAGE}-${VERSION}/TODO.md
+	tar --delete --file "$@" ${PACKAGE}-${VERSION}/testing
+	tar --delete --file "$@" ${PACKAGE}-${VERSION}/util
+	tar --delete --file "$@" ${PACKAGE}-${VERSION}/.gitignore
+	tar --delete --file "$@" ${PACKAGE}-${VERSION}/.mailmap
+	tar --delete --file "$@" ${PACKAGE}-${VERSION}/screenshot*.png
```
cbm755: Make wizards: any nicer way to list all these files to be excluded?

cbm755:
I propose doing this for now.  Improvements certainly welcome, either now or post merge.

Matlab packaging could be improved but its likely not functional anyway (bitrotted) so not sure its worth worrying too much about the makefile aspects.  I might drop Matlab support at some point (although its kind of nice to test code with a different "compiler")

Makefile
mtmiller: You can have a .gitattributes file that names certain files that should always be excluded from `git archive`, but you may not want that set globally.

You can do the inverse, tell `git archive` to only include `CONTRIBUTORS`, `NEWS`, `README.md`, `inst`, etc.

mtmiller: Or you could use `git archive` and immediately unpack it into a temporary directory, fix up the directory with `chmod` and `rm`, and then tar and compress the result. That's probably what I would do, I try to avoid GNU options like `--delete`.

genuinelucifer:
I tried current source with MATLAB R2013a and it gives a lot of errors.. In my opinion it's probably better to drop MATLAB support.

mtmiller:
Yes let's just get this done, leave the Matlab stuff in there for now as a best effort, and if it's not too much trouble someone may want to get it working again.

@cbm755 has talked about making the sym class derive from pyobject at some point in the future. That may be the point where we lose all hope of Matlab compatibility.

cbm755:
Merging, more PRs to improve Makefile certainly welcome.

Makefile
carandraug: For what is worth, that is exactly what is [bsltl package does](https://sourceforge.net/p/octave/bsltl/ci/master/tree/Makefile) (the only other package in Forge that uses git and has a maintainer `Makefile`):

    git archive --format=tar --prefix="$@/" HEAD | tar -x
    $(RM) "$@/Makefile"
    $(RM) "$@/.gitignore"

It then redoes a tarball and a zip file.

PR #661, opened 2016-11-03
[WIP] travis build cleanup

mtmiller:
There's at least a small conflict here with #659, just want to make sure my cleanups actually work in travis and get some feedback.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,7 @@ install:
         octave -W --no-gui --eval "pkg install -forge doctest";
     fi
   - if [ "x$PYTAVE" = "xyes" ]; then
-        hg clone https://bitbucket.org/genuinelucifer/pytave_main;
-        mv pytave_main pytave;
+        hg clone https://bitbucket.org/genuinelucifer/pytave_main pytave;
         pushd pytave;
         hg checkout travis;
         pwd;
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -32,14 +36,12 @@ before_install:
         popd;
     fi
 
-
 # all commands here must have exit code 0 for the build to be called "passing"
 # debugging: octave -W --no-gui --eval "syms x; A = [x sin(x) x^3]; A; exit(0)"
 script:
   - octave -W --no-gui --eval "ver; pwd; exit(0)"
-  - stty cols 80 rows 40
+  - stty cols $COLUMNS rows 40
```
cbm755: Hmmm, I had an actual problem with this.  I can try to track it down, but I'm scared of this change b/c of some travis bug IIRC.

cbm755: oh wait, $COLUMNS comes from my own file :) In that case, ok!

cbm755:
LGTM so far.  Let me know if you want it merged now, or planning more.

latot:
Hi, little question, when you think is possible enable pytave in travis?

mtmiller:
I think this is about all I was planning on for now, you can merge if you like since it looks like it's working.

@latot I think the main blocker is having Octave 4.2 installable via PPA, once there is an Octave 4.2 package then it should be pretty easy to add pytave with Python 2 and 3 to the CI matrix.

latot:
Hi, sorry ask this but, what is the main blocker?

cbm755:
@latot: We cannot easily install Octave 4.2 on Travis CI.

Its not worth trying to backport Pytave to work on earlier Octave.  (IIRC, @genuinelucifer and I had it working on our own machines but its not a good use of time given 4.2 is coming Real Soon Now).

cbm755:
@mtmiller this reminds me to ask that we can have an "old Octave" PPA so that I can keep running CI tests against Octave 4.0 even after 4.2 is out (and in your PPA).

PR #662, opened 2016-11-03
Fix syntax error in native ipc driver

mtmiller:
Add missing quote from commit eb8076e5f41e3403bc53b2c01af47de295e210d9.

Fixes the native (pytave) interface, apparently broken in PR #583.


cbm755:
oh my!  thanks

PR #665, opened 2016-11-04
Fix the empty elementwise regression

### LINKED ISSUE #650: regression: op_helper with empty args: `sym([]) == sym([])` should be empty

cbm755:
I think this fixes #650, @latot does it look ok to you?

### DIFF
--- a/inst/@sym/private/elementwise_op.m
+++ b/inst/@sym/private/elementwise_op.m
@@ -37,7 +37,7 @@
 %% python function in a cell array and use it:
 %%
 %% @example:
-%% scalar_fcn = @{ 'def _op(a,b,c,d):' '    return a % c + d / b' @};
+%% scalar_fcn = @{ 'def _op(a,b,c,d):'; '    return a % c + d / b' @};
 %% A = 3;
 %% B = [1 2;3 4];
 %% C = inv(B);
--- a/inst/@sym/private/elementwise_op.m
+++ b/inst/@sym/private/elementwise_op.m
@@ -85,10 +85,11 @@ function z = elementwise_op(scalar_fcn, varargin)
           'assert sameshape, "Matrices in input must all have the same shape"'
           '# could be no matrices in the input'
           'shape = shapes[0] if len(shapes) > 0 else (1, 1)'
-          'q = zeros(*shape)'
-          'for i in range(0, len(q)):'
-          '    q[i] = _op(*[A[i] if isinstance(A, MatrixBase) else A for A in _ins])'
-          'return q' ];
+          'L = []'
+          'for i in range(prod(shape)):'
+          '    L.append(_op(*[A[i] if isinstance(A, MatrixBase) else A for A in _ins]))'
+          'assert len(shape) == 2, "Dims > 2 not implemented yet"'
+          'return Matrix(shape[0], shape[1], L)' ];
 
   z = python_cmd (cmd, varargin{:});
 
--- a/inst/@sym/private/elementwise_op.m
+++ b/inst/@sym/private/elementwise_op.m
@@ -87,8 +87,8 @@ function z = elementwise_op(scalar_fcn, varargin)
           'shape = shapes[0] if len(shapes) > 0 else (1, 1)'
           'q = zeros(*shape)'
           'for i in range(0, len(q)):'
-          '    q[i] = _op(*[k[i] if isinstance(k, MatrixBase) and k else k for k in _ins])'
-          'return q,' ];
+          '    q[i] = _op(*[A[i] if isinstance(A, MatrixBase) else A for A in _ins])'
+          'return q' ];
 
   z = python_cmd (cmd, varargin{:});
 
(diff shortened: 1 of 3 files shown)


inst/@sym/private/elementwise_op.m
```diff
@@ -80,13 +80,12 @@
...
-          '        else:'
-          '            assert q.shape == A.shape, "Matrices must have equal sizes"'
+          'shapes = [A.shape for A in _ins if isinstance(A, MatrixBase)]'
+          'sameshape = all([s == shapes[0] for s in shapes])'
+          'assert sameshape, "Matrices in input must all have the same shape"'
+          '# could be no matrices in the input'
+          'shape = shapes[0] if len(shapes) > 0 else (1, 1)'
+          'q = zeros(*shape)'
           'for i in range(0, len(q)):'
           '    q[i] = _op(*[k[i] if isinstance(k, MatrixBase) and k else k for k in _ins])'
```
cbm755: @latot what does `and k` in this "`isinstance(k, MatrixBase) and k`" do?

I assume its something to do with `None`: but if its a `MatrixBase` than it cannot be `None`...

latot:
Hi, i'll check this later, i'm out of time now, for now can you read this comment plis?
https://github.com/cbm755/octsympy/pull/653#issuecomment-258468505

I say there some things you are asking here.

Thx. Cya.

latot:
Hi, i already check this, i like some changes but really others not, point by point:
To fix the issue you need remove ```and A:``` and ```and k```, with that the issue will be done, obvs i don't have any problem updating the error message. That lines was to can use empty arrays as elements, that was the objetive, with your fix in limit i think will be fine.

I don't like to much the changes in the first section, basically 2 reasons:
  1) use zeros will make a new matrix and consume more memory, depending of the size it will use more or less time.
  2) zeros don't support dim > 2

The idea of the actual code is found the element with the matriz size needed for the user, and then store the data there, with this we can avoid make new elements and use more memory, the next point we don't need handle the type of data, this means, if the user send an Array or a Matrix this will use it and we don't need create the object, this is compatible with Tensor and Matrix at the same time.

Thx. Cya.

cbm755:
Thanks for the comments.  I first replaced the `zeros` bit with a list, then build the `Matrix` from the list.  
Then I remembered you can pass a lambda to Matrix; so I re-wrote it using that.

Still doesn't work with Tensors, but I protected it a bit with an assert.  That will do for now.  We can explore improving for Tensors later.

latot:
Hi, i'm confused with this changes.................

with the previous code to support Tensor the only we need to do is change 2 lines, basically add ```NDimArray``` to ```isinstance``` and works....

Now, when we can support Tensors we will need rewrite the entire function instead change 2 lines..., and don't is a rewrite, is back the code to previous this pr.

I still don't like the idea use more memory to compact some lines...............

Maybe keep the previous code in comments lines to prepare it for Tensor?

Thx. Cya.

cbm755:
Well, this code does not rely on adding a bunch of helper functions to python_header, and it fixes the bug in question without adding new feature.  If we need to rewrite it later, that's fine.

Nothing wrong with memory usage here: Matrix is constructed with a lambda.  No temporary arrays of the same shape as the inputs are created.

latot:
Hi, i know i have the other pr with that features, but i mean the previous code don't need it, was only remove the ```and A``` & ```and k``` and the bug was fixed..................
And why rewrite the entire function later if was already working now...? (taking in consideration the minor change needed to fix the issue)

Sorry if i can't express in a assertive way this.

cbm755:
> was only remove the `and A` & `and k` and the bug was fixed

Oh!  This I didn't understand.  I thought the problem was starting with `q = Matrix([0])` so I tried to rewrite it without that part.  I'd have to go back and check more carefully, but if the matrix started as 1x1 I don't see how the old code could have made it smaller...

latot:
mm, little explanation of why it works (with the change), in first place, why was the script using empty arrays as elements?, is exactly for that lines, when you use an array as argument in a if statement only return true when is not empty:
```
>>> a=[]
>>> b=Matrix(a)
>>> if a:
...  1
... else:
...  0
... 
0
>>> if b:
...  1
... else:
...  0
... 
0
```
So, the script use the empty matrix as element because we exclude it to travel it with that condition, following the same, if we remove that condition the empty matrix will be recognized again as a valid matrix and the script will travel in this.
I don't know how do the old code smaller, but i think in this case don't have to much importance, because the difference is only of 2 lines (old code have 10 lines, new code have 8).

Thx. Cya.

cbm755:
Well, we could switch it back to a loop-based instead of list comprehension...    I still don't see how the `0x0` and `0x3` cases will work.  But instead of talking in theory, we should just try it.  Do you want to send a PR for this (and only this)?

latot:
Hi, the only cases when can exist 0xn, is with MatrixExpr, but in the doc it say explicitly this function don't travel in that type of matrix.

I thing i need moderate the level of changes i'm doing in my prs....

cbm755:
Yes, but I said `0 x 3` not `0 x n`.

Anyway my point is, we don't have to argue about it.  If you have an implementation that is nicer (and doesn't do a bazillion other things), show us the code and we can decide which one to use.

PR #667, opened 2016-11-04
sympref: Restore user preferences after unit tests

### LINKED ISSUE #666: `test sympref` resets some user settings to their default values
I want the `sympref` state to be preserved before and after calling `test sympref`. This will allow one to write

    sympref ipc native ## or any other ipc type
    octsympy_tests

to run the entire test suite using a different ipc driver. Right now it hits `sympref` pretty early on in the test run, which resets to default (popen2), and runs the remaining tests using popen2.

This will also help with #461.

I have a fix for this, just reporting to log the bug.

mtmiller:
Fixes #666.

### DIFF
--- a/inst/sympref.m
+++ b/inst/sympref.m
@@ -369,6 +369,9 @@ function r = tf_from_input(s)
 end
 
 
+%!shared sympref_orig
+%! sympref_orig = sympref ();
+
 %!test
 %! % test quiet, side effect of making following tests a bit less noisy!
 %! sympref quiet on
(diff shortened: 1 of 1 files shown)


inst/sympref.m
```diff
@@ -428,7 +431,8 @@
 %!test
 %! syms x
 %! r = sympref('reset');
+%! % restore original sympref settings
+%! sympref ('ipc',   sympref_orig.ipc);
+%! sympref ('quiet', sympref_orig.quiet);
 %! syms x
 %! assert(r)
-%! % ok, can be noisy again
-%! sympref('quiet', 'default')
```
mtmiller: @cbm755 Not sure about this, was the purpose of this last test block to simply restore default state? Or am I screwing something up, should I add another block after this one to reset yet again and restore the settings?

cbm755: I think its ok.  Maybe line 434 is not needed as resetting the IPC does that anyway, but its not a big deal.

cbm755:
If a test fails within that shared block, then subsequent runs within the same session may not have the desired IPC.

(I'm ok with this, just noting for the record).

PR #670, opened 2016-11-04
limit: use a dummy var if neither symvar nor user give one

### LINKED ISSUE #669: limit uses symvar but doesn't check for empty return
Currently it seems to work but I don't see that its supported from reading the SymPy `limit` documentation: in particular, it only works because `limit` has `Limit(e, z, z0, dir).doit(deep=False)`.  If you change that to `deep=True` then it fails if `z` is empty.

I think our limit should just use some dummy variable like `x` if `symvar` returns empty.

cbm755:
Fixes #669.

### DIFF
--- a/inst/@sym/limit.m
+++ b/inst/@sym/limit.m
@@ -85,7 +85,8 @@ function L = limit(f, x, a, dir)
   end
 
   if (isempty (x))
-    x = sym('x');
+    L = f;
+    return
   end
 
   cmd = { '(f, x, a, pdir) = _ins'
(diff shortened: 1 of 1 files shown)


inst/@sym/limit.m
```diff
@@ -83,6 +83,10 @@
       print_usage ();
   end
 
+  if (isempty (x))
+    x = sym('x');
```
latot: if x is empty, can we just return f?, without a limit var, the limit will be the same function.

cbm755: That makes sense, done.

cbm755:
I fixed another small bug I found.  Should have done a separate PR to practice what I preach but here we are.

cbm755:
Will merge assuming tests pass.

PR #672, opened 2016-11-05
Add test for multirow char arrays

### LINKED ISSUE #664: string matrix don't are passed correctly to python - only pass the first row
Hi all:
```
octave:133> a=['a', 'b';'c', 'd'];
octave:134> python_cmd('return _ins[0]', a)
warning: called from
ans = ab
```
And as a note, i can confirm in ```_ins``` only exist one element, so don't exist ```_ins[1]```

Thx. Cya.

cbm755:
Fixes #664.

### DIFF
--- a/inst/python_cmd.m
+++ b/inst/python_cmd.m
@@ -442,3 +442,17 @@ end
 %! % complex output
 %! z = python_cmd ('return 3+2j');
 %! assert (z, 3+2i)
+
+%!test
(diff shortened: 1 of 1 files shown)


cbm755:
I think I have a +1 from @mtmiller in #664 discussion so merging for now.  If behaviour changes, test should be modified.  Ideally the comment in the test should note that, but I'm lazy.

PR #676, opened 2016-11-07
Elliptic

latot:
Hi!, now the elliptic functions.

As you can see some functions have less or more tests, this is because some approximations of SMT don't match with Sympy, and checking don't match with WolframAlpha too, take in consideration the approximations of Sympy are too close to WolframAlpha, so in that cases i use less tests.

Little example:
EllipticCPi(-1, 1/3)
  WA:      1.205059303480956918275419123818767032685642224169380264249...
  Sympy: 1.205059303480956918275419123818767032685642224169380264249...
  SMT:     1.370337322....

you can use this commands directly in this platforms.

Thx. Cya.

### DIFF
--- a/inst/@sym/ellipke.m
+++ b/inst/@sym/ellipke.m
@@ -1,0 +1,97 @@
+%% Copyright (C) 2016 Lagu
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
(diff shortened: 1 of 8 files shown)


cbm755:
I started looking at this today---sorry for the long delay.

Can you cherry-pick this commit?   127971248fe271e300687d0b14e891154029b36c

1.  I think we should use Sympy's elliptic integral functions instead of the integral.
2.  It would also be great to have sympy make the integrals for us.  I think that needs an upstream change similar to https://github.com/sympy/sympy/pull/11014  (which could use a review btw!)

latot:
ups, when i search it for some reason i could not find it, i'll rewrite this, Thx.

latot:
Hi, checking the elliptic function from sympy, sadly don't are from there, is from the mpmath, and that function don't works with symbols, i test it....

latot:
here i found something, https://github.com/sympy/sympy/blob/master/sympy/functions/special/elliptic_integrals.py

i'll try use this instead

cbm755:
Please "git cherry-pick 1279712".

cbm755:
My commit uses those special functions that you mentioned; that's why I mentioned it.  I would've sent a PR, but this branch is too far behind master.

latot:
Okidokis, all updated and working.

cbm755:
For next time, I think a "cherry-pick" of my commit, followed by a "merge origin/master" would've been better than a rebase.  Best practice is rebase after review, not during.

latot:
Hi, i was trying to check this but the branch was to far of master, and for that i can't merge upstream/master, and that branch had problems with octave so i can't test it to check the functions, for that reason i reset the branch.

cbm755:
> Little example:

Perhaps there was a mixup or typo about `CPi` versus `Pi` that lead to this?  Can you double check my commit a66602fcb698e75c1e05087dbb4b908993edd6d9?

PR #677, opened 2016-11-07
elementwise_op pre-tensor

latot:
Hi, here is the code, can i put 2 comments for the future?

Thx. Cya.

### DIFF
--- a/inst/@sym/private/elementwise_op.m
+++ b/inst/@sym/private/elementwise_op.m
@@ -64,7 +64,6 @@
 %%   This function doesn't work with MatrixSymbol.
 %%   If you send matrices as arguments, all must have equal sizes
 %%   except if have a 1x1 size, in that case it always will ise that value.
-%%   Empty arrays are treated as elements.
 %%
 %% @end defmethod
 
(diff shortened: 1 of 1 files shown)


cbm755:
Suppose the inputs are `Matrix([1])`, `Matrix(2,2,[...])`, `Matrix([1])`.  Then this will raise the error.

But suppose the inputs are `Matrix([1])`, `Matrix([1])`, `Matrix(2,2,[...])`.  No error will be raised.

Our code will not hit that because we do not use 1x1 matrices.  But makes me uncomfortable and makes it difficult for me to understand the code.

How about?
````
# Make sure all matrices in the input are the same size, and set q to one of them
q = None
for A in _ins:
    if isinstance(A, MatrixBase):
        if q is None:
            q = A
        else
            ...
# in case all inputs were scalars:
if q is None:
    q = Matrix([0])
...
````

cbm755:
>  can i put 2 comments for the future?

Do you mean, "write an actual commit message"?  Yes please do!  Rebase is fine here.

latot:
Hi, i test the code and send the error when we send matrix with different sizes, anyway, i update the code to be more clear and don't cause confusion, i think is a nice change use None, i add some comments line and a better description in the commit.

Thx!

cbm755:
This is in.  I rewrote the commit msg.  I reformatted a bit (no `#` comments to be used).  Also `if q is None` instead of `==`.  Moved the tensor comments to one place.

latot:
Hi!, i'm little confused, the pr is closed, but i don't see any change in master with this..
¿?

Thx. Cya.

cbm755:
Oops I should have linked to the commit: its 525e16e504957a0ee071c40d8ec0f69cf5cb6d7e.

As I explained I changed it a bit before merging.  Probably I should do that; I should try to do everything through PRs only.

latot:
Hi, my git was little weird, for that i could not see your changes, now i found the commits!

Thx!

Issue #91, opened 2014-09-21
converting a symbolic variable using char() fails


liondog:
In MATLAB commands using the char function like

```
clear all
syms t x(t) y z
N = sym(char([x, y, z]))
```

work but will fail in Octave/octsympy. (Tested on Windows.)

PR #680, opened 2016-11-11
char: gives human-readable string not sympy srepr

### LINKED ISSUE #375: maybe `char(x)` should do something different
Someone contacted me on sourceforge suggesting using it to expose the underlying `srepr` might not be ideal....

TODO: find the conversation, post here.

cbm755:
This is more useful, e.g., for labelling a plot and I think its
what people would expect to happen if they cast a object to char.
It does mean that `sym(char(x))` doesn't always reproduce `x` with
the fidelity that it used too because previously `x` gave the
implementation (srepr) of the object.  That change is documented in
char.m.

Fixes #375, also effects (or affects) #91.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -58,6 +58,9 @@ octsympy 2.5.0-dev
 
   * Bug fix: assign `[]` to row/column removes that that row/column.
 
+  * `char(x)` now gives a more-human readable output, useful for example
+    in labelling a figure.
+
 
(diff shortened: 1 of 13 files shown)


mtmiller:
👍 LGTM so far.

cbm755:
Thanks @mtmiller, merging

PR #681, opened 2016-11-12
Remove the "snippet" display feature

### LINKED ISSUE #651: remove "snippets" display feature
The "snippets" feature (`sympref snippet on`) is cute but does anyone care?  I never use it.  One can just do `char(f)`.

Its visual clutter and makes the display code harder to maintain (see e.g., #583).

@genuinelucifer and @mtmiller many remember that pytave had something like this---and we also removed it.

Any objections to removing this?  If not, patches happily accepted

cbm755:
I don't think it is very useful and has a significant amount of
code that exists only to support it.  Fixes #651.


mtmiller:
👏 Good clean up of unnecessary logic.

inst/sympref.m
```diff
@@ -249,17 +230,6 @@
         settings.digits = arg;
       end
 
-    case 'snippet'
```
mtmiller: How much do you care about deprecation vs deleting? Should sympref warn with "the snippet option has been removed"?

cbm755: Good point.  I wasn't going to bother with formal deprecation (keeping functionality for for one version) because its a debug thing, but I could at least warn as you suggest.

PR #685, opened 2016-11-12
More simplifying of equationsToMatrix

cbm755:
@latot, is this ok?  AFAICT, `Matrix` doesn't do anything here.


latot:
mm, as i remember, not all objects from poly was a sympy object, so some times the command to get the symbols fails, for that reason i use the matrix, maybe instead of matrix use ```S```?

latot:
```
>>> srepr(S(Poly.from_expr(S(1), x).all_coeffs()).free_symbols)
AttributeError: 'list' object has no attribute 'free_symbols'
```
```
>>> srepr(Matrix(Poly.from_expr(S(1), x).all_coeffs()).free_symbols)
'set()'
```
Testing and remember this, S don't works too, need to be matrix, well my fail here is don't add a test for this, can you add it please?

latot:
mmm, mmm, mm, rerereching, i think we can now remove safely the matrix section, other part of the code handle this possible problem so should be fine, sorry all previous comments.

cbm755:
thanks!

Issue #674, opened 2016-11-06
[maint] Makefile should make a zip file


cbm755:
Traditionally we released tarballs and zip files.  Makefile should make a zip too.

Also, see [this comment from @mtmiller](https://github.com/cbm755/octsympy/pull/659#discussion_r86474354) to improve/change how we make the tarball (this is relevant to making a zip file, as we'd be untarring the one git makes anyway).

cbm755:
Fixed in #691.

PR #689, opened 2016-11-13
Add the packaging of the realease as a zip file too.

genuinelucifer:
Fixes issue #674  
Add the commands to make a zip file while packaging for a release. Also modify the current packaging procedure to avoid gnu options.  
.  
.  
There are 2 questions from my side for this:  
* Do we want to create the zip file with the tarball or as a separate rule?  
* Should we change the $(BUILD_DIR) from 'tmp' to something else like 'build_dir'?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ BUILD_DIR = tmp
 MATLAB_PKG_DIR=${MATLAB_PACKAGE}-matlab-${VERSION}
 OCTAVE_RELEASE_TARBALL = ${BUILD_DIR}/${PACKAGE}-${VERSION}.tar
 OCTAVE_RELEASE_TARBALL_COMPRESSED = ${OCTAVE_RELEASE_TARBALL}.gz
+OCTAVE_RELEASE_ZIP = ${BUILD_DIR}/${PACKAGE}-${VERSION}.zip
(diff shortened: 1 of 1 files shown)


genuinelucifer:
I hope I understood the essence of #674 correctly. I have no clue about standards of Makefiles and hence I might have done something horrible (hopefully not). Suggestions/edits are welcome.

cbm755:
I haven't looked at your PR yet, but @carandraug suggested we look at https://sourceforge.net/p/octave/bsltl/ci/master/tree/Makefile

Maybe take a look?

carandraug:
The `zip` command updates an existing archive instead of creating a new one. This means that consecutive calls to the `zip` target would not remove files that had been removed from the repository.

I just made pull request #691 which does this in a different way. It adds a target to export the package and prepare all files right before the archiving and compression step. This means that recipes to make the `tar.gz` and `zip` files can do the archiving and nothing else. It seems cleaner than having `OCTAVE_RELEASE_TARBALL` effectively doing two things, creating the release directory and the tarball, and then having the zip target removing the directory.

genuinelucifer:
Closing as #691 does it better.

cbm755:
Make scares me, so thanks @genuinelucifer for your efforts and your review of #691 (I'm taking your comment as a "+1" for that)

genuinelucifer:
> I'm taking your comment as a "+1" for that

Definitely.

Issue #598, opened 2016-10-04
subsasgn of symbolic array with logical indexing fails when zero or one elements affected


mtmiller:
Assignment to a symbolic array using logical indexing fails when the assignment would affect either zero or one elements of the array.

Simple example of assignment to zero elements:

```
>> A = sym ([0 0])
>> indices = [false false]
>> A[indices] = []
error: Python exception: TypeError: cannot add matrix and <type 'int'>
    occurred at line 1 of the Python code block
error: called from
    python_cmd at line 174 column 5
    mat_replace at line 40 column 15
    subsasgn at line 99 column 13
```

Simple example of assignment to one element:

```
>> A = sym ([0 0])
>> indices = [true false];
>> A(indices) = 1;
error: Python exception: AttributeError: 'One' object has no attribute 'T'
    occurred at line 5 of the Python code block
error: called from
    python_cmd at line 174 column 5
    mat_mask_asgn at line 88 column 5
    mat_replace at line 71 column 7
    subsasgn at line 99 column 13
```

If `indices` contains two or more `true` values, the assignment succeeds.

genuinelucifer:
This also occurs if we try to assign `[]` to any number of elements (may be more than 1)...  
```
>> A = sym([0 0 0]);
>> indices = [true false true];
>> A(indices) = [];
error: Python exception: TypeError: cannot add matrix and <type 'int'>
    occurred at line 1 of the Python code block:
    _ins[0].col_del(_ins[1] - 1); return _ins[0],
error: called from
    python_cmd at line 175 column 5
    mat_replace at line 40 column 15
    subsasgn at line 99 column 13
```

cbm755:
This is fixed, thanks @genuinelucifer.

(it didn't auto-close: I think github is rather particular about the string "Fixes #598")

PR #690, opened 2016-11-13
Fix subsasgn behaviour when logical indices are used.

genuinelucifer:
Fixes issue #598
When logical indices are used, pass the valid indices (if any). If there are
no valid indices then return the input as is.

### DIFF
--- a/inst/@sym/private/mat_replace.m
+++ b/inst/@sym/private/mat_replace.m
@@ -1,5 +1,6 @@
 %% Copyright (C) 2014-2016 Colin B. Macdonald
 %% Copyright (C) 2016 Lagu
+%% Copyright (C) 2016 Abhinav Tripathi
 %%
 %% This file is part of OctSymPy.
 %%
@@ -28,6 +29,18 @@
 
 function z = mat_replace(A, subs, b)
 
+  if (length (subs) == 1 && islogical (subs{1}))
+    %% A(logical) = B
+    subs{1} = find (subs{1});
+    if isempty (subs{1})
+      z = A;
+      return;
+    end
+    if (~ isscalar (b) && ~ isvector (b) && ~ isempty (b))
+      warning ('OctSymPy:subsagn:rhs_shape', ...
+            'B neither vector nor scalar in indexed A(I)=B: unusual, did you intend this?')
+    end
+  end
   %% Check when b is []
   if (isempty(b))
     switch length(subs)
@@ -66,12 +79,7 @@ function z = mat_replace(A, subs, b)
     end
   end
 
-  if (length(subs) == 1 && islogical(subs{1}))
-    %% A(logical) = B
-    z = mat_mask_asgn(A, subs{1}, b);
-    return
-
-  elseif (length(subs) == 1 && strcmp(subs{1}, ':') && length(b) == 1)
+  if (length(subs) == 1 && strcmp(subs{1}, ':') && length(b) == 1)
     z = python_cmd('return ones(_ins[0], _ins[1])*_ins[2],', A.size(1), A.size(2), sym(b));
     return
 
--- a/inst/@sym/private/mat_replace.m
+++ b/inst/@sym/private/mat_replace.m
@@ -94,6 +94,9 @@ function z = mat_replace(A, subs, b)
     else
       % linear indices into 2D array
       [r, c] = ind2sub (size(A), subs{1});
+      % keep all the indices in a row vector
+      r = reshape (r, 1, []);
+      c = reshape (c, 1, []);
     end
 
   elseif (length(subs) == 2)
--- a/inst/@sym/subsasgn.m
+++ b/inst/@sym/subsasgn.m
@@ -522,6 +522,14 @@ end
 %% Tests from mat_replace
 
 %!test
+%! % 2D indexing with length in one dimension more than 2
+%! a = sym ([1 2; 3 4; 5 6]);
+%! indices = [1 4; 2 5; 3 6];
+%! b = [10 11; 12 13; 14 15];
+%! a(indices) = b;
+%! assert (isequal (a, sym (b)));
+
+%!test
 %! A = sym ([0 0 0]);
 %! indices = [false true false];
 %! A(indices) = 1;
--- a/inst/@sym/subsasgn.m
+++ b/inst/@sym/subsasgn.m
@@ -1,5 +1,6 @@
 %% Copyright (C) 2014-2016 Colin B. Macdonald
 %% Copyright (C) 2016 Lagu
+%% Copyright (C) 2016 Abhinav Tripathi
 %%
 %% This file is part of OctSymPy.
 %%
@@ -516,5 +517,64 @@ end
 %!error <null assignment>
 %! a = sym([1, 2; 3, 4]);
 %! a(1, 2) = [];
+
+
+%% Tests from mat_replace
+
+%!test
+%! A = sym ([0 0 0]);
+%! indices = [false true false];
+%! A(indices) = 1;
+%! assert (isequal (A, sym ([0 1 0])));
+%! A(indices) = [];
+%! assert (isequal (A, sym ([0 0])));
+%! indices = [false false];
+%! A(indices) = [];
+%! assert (isequal (A, sym ([0 0])));
+
+%!shared a, b
+%! a = [1 2 3 5; 4 5 6 9; 7 5 3 2];
+%! b = sym (a);
+
+%!function test_bool (a, b, c)
+%!  a(c) = 0;
+%!  b(c) = 0;
+%!  assert (isequal (a, b))
+%!endfunction
+
+%!test
+%! c = false;
+%! test_bool (a, b, c)
+%! test_bool (a, b, c | true)
+
+%!test
+%! c = [false true];
+%! test_bool (a, b, c)
+%! test_bool (a, b, c | true)
+%! test_bool (a, b, c & false)
+
+%!test
+%! c = [false true false true; true false true false; false true false true];
+%! test_bool (a, b, c)
+%! test_bool (a, b, c | true)
+%! test_bool (a, b, c & false)
+
+%!test
+%! c = [false true false true false];
+%! test_bool (a, b, c)
+%! test_bool (a, b, c | true)
+%! test_bool (a, b, c & false)
+
+%!test
+%! c = [false; true; false; true; false];
+%! test_bool (a, b, c)
+%! test_bool (a, b, c | true)
+%! test_bool (a, b, c & false)
+
+%!test
+%! c = [false true; false true; true false];
+%! test_bool (a, b, c)
+%! test_bool (a, b, c | true)
+%! test_bool (a, b, c & false)
 
 %% End of mat_* tests


genuinelucifer:
Hopefully there aren't any hidden cases which would break by this. As far as I have tested this seems a reasonable solution.

genuinelucifer:
There's a test failure in subsasgn.. When I ran the test in the file, I just got a warning of failure which I stupidly ignored. I'll fix it tomorrow.

latot:
Hi, can you add this tests plis? i know this seems a lot, but every test confirm differents parts of this, although likes similars, and this need have this minimum size of the matrix to can test all of this:
```
%!shared a, b
%! a = [1 2 3 5;4 5 6 9;7 5 3 2];
%! b = sym(a);

%!function test_bool (a, b, c)
%!  assert (a(c), double (b(c)))
%!  a(c) = 0;
%!  b(c) = 0;
%!  assert (nnz (a == 0), nnz (b == 0))
%!endfunction

%!test
%! c = false;
%! test_bool (a, b, c)
%! test_bool (a, b, c | true)

%!test
%! c = [false true];
%! test_bool (a, b, c)
%! test_bool (a, b, c | true)
%! test_bool (a, b, c & false)

%!test
%! c = [false; true];
%! test_bool (a, b, c)
%! test_bool (a, b, c | true)
%! test_bool (a, b, c & false)

%!test
%! c = [false true false true; true false true false; false true false true];
%! test_bool (a, b, c)
%! test_bool (a, b, c | true)
%! test_bool (a, b, c & false)

%!test
%! c = [false true false true false];
%! test_bool (a, b, c)
%! test_bool (a, b, c | true)
%! test_bool (a, b, c & false)

%!test
%! c = [false; true; false; true; false];
%! test_bool (a, b, c)
%! test_bool (a, b, c | true)
%! test_bool (a, b, c & false)

%!test
%! c = [false true; false true];
%! test_bool (a, b, c)
%! test_bool (a, b, c | true)
%! test_bool (a, b, c & false)

%!test
%! c = [false true false; false true false];
%! test_bool (a, b, c)
%! test_bool (a, b, c | true)
%! test_bool (a, b, c & false)

%!test
%! c = [false true; false true; true false];
%! test_bool (a, b, c)
%! test_bool (a, b, c | true)
%! test_bool (a, b, c & false)
```
I think the issue will be solved when all this test pass.

Thx. Cya.

latot:
Only as a note, this need a lot of test because the behavior is very different with logic index and non-logic index..........

genuinelucifer:
@latot I added your tests with following 2 changes:  
* I removed the first line `assert (a(c), double (b(c)))` from `test_bool` as it exposed the problem with `subsref` and not `subsasgn`. I will fix the problems with `subsref` in next PR.
* I removed 2 of the tests where you mixed logicals with integers in `c` as this is not supported in octave for any type of matrix anyway

latot:
Hi!, i'm agree to remove the subsref tests, we will need open a new issue for subsref, its possible keep the assert as a comment? only to remember it.

The test mixing logicals and integers is supported for octave:
```
octave:15> a=[1, 2, 3;4, 5, 6]
a =

   1   2   3
   4   5   6

a([true 1;true 2])=0
a =

   0   2   3
   0   5   6
```
But i think we can do it later.

Thx. Cya.

genuinelucifer:
I also removed some of your tests which were superseded by some other tests.  
  
> The test mixing logicals and integers is supported for octave

I see. It works only if all the logicals are `true`. We can probably support this then. I'll check.

latot:
mm, some test are lost...., was this intentional?

latot:
comment removed, let me check something...

inst/@sym/private/mat_replace.m
```diff
@@ -28,6 +28,17 @@
...
 
+  if islogical(subs{1})
+    subs{1} = find(subs{1});
+    if isempty(subs{1})
+      z = A;
+      return;
+    end
+    if (~isscalar(b) && ~isvector(b))
+      warning('OctSymPy:subsagn:rhs_shape', ...
+            'B neither vector nor scalar in indexed A(I)=B: unusual, did you intend this?')
```
genuinelucifer: @cbm755 Is this warning important?

genuinelucifer:
5 doctests are failing because `evalpy` is deprecated!!.. I'll see if I need to rebase!

latot:
I'm with the same problem!, should that be fixed from master and merge here?

inst/@sym/private/mat_replace.m
cbm755: I'm not sure.  If Octave double allows it, I guess we should too.

cbm755:
Ok, I reverted the commit for evalpy, sorry about that.

cbm755:
It might not have been that commit, it might've been the new doctest 0.5.0 coming in, which needed changes.  Anyway, master is fine again on CI and I've restarted jobs.

inst/@sym/private/mat_replace.m
genuinelucifer: Octave double allows it without any warning. Should I remove the warning then? (from here and from the original file `mat_mask_asgn`)

genuinelucifer:
@latot I have checked with different values and saw that octave actually replaces all the `true` in indices to `1` if the indices consist of integers too! And `false` is not at all allowed in this case. Thanks for the test cases...  
  
I also saw that currently it was an error if the index used had more than 2 rows because of how `ind2subs` works. So, I fixed that in here too.
  
@cbm755 I think this is now ready. Please review it. Also, I think the 3 commits are fine. Should I squash the first 2 or is it ok like this?

cbm755:
Also: please add yourself to the copyright header of these two files

inst/@sym/subsasgn.m
```diff
@@ -513,8 +513,78 @@
...
+%! test_bool (a, b, c & false)
+
+%!test
+%! c = [false true; false true; true false];
+%! test_bool (a, b, c)
+%! test_bool (a, b, c | true)
+%! test_bool (a, b, c & false)
+
+%!test
+%! c = [2 true; true 3; true true];
```
latot: Hi, octave support doubles and bolean mixed for index, i think is a valid test.

cbm755: No it doesn't.  `[true 2]` is cast to double.  `whos` will show you.

latot: uu, true, when we mix logical with doubles octave transform logical values to that type.

mmm, i think we need to be careful and check how this works when we send a symbolic index matrix.

cbm755: Good point, file another issue so we don't loose it?

inst/@sym/subsasgn.m
```diff
@@ -513,8 +513,78 @@
 %! a(1) = [];
 %! assert( isequal( a, b));
 
+%!test
+%! % indexing with a dimension more than 2
```
genuinelucifer: I actually got confused while choosing the label. This test actually is for indexing with a mXn matrix where either one of m or n is more than 2. So probably something along the lines of "indexing with a matrix with size of a dimension more than 2"!?

inst/@sym/subsasgn.m
```diff
@@ -513,8 +513,78 @@
...
+%! a = sym ([1, 2; 3, 4; 5 6]);
+%! indices = [1 2; 3 4; 5 6];
+%! b = [10 11; 12 13; 14 15];
+%! a (indices) = b;
```
genuinelucifer: Oh thanks! Will do.

inst/@sym/subsasgn.m
```diff
@@ -513,8 +513,78 @@
...
+%! assert (isequal (A, sym ([0 0])));
+
+%!shared a, b
+%! a = [1 2 3 5;4 5 6 9;7 5 3 2];
+%! b = sym(a);
+
+%!function test_bool (a, b, c)
+%!  a(c) = 0;
+%!  b(c) = 0;
+%!  assert (nnz (a == 0), nnz (b == 0))
```
genuinelucifer: That would work too. I thin this function,  in current state, just checks for the changes in a, b and not the exact values.
Since @latot wrote it, probably he can comment here!

inst/@sym/subsasgn.m
genuinelucifer: After this PR is merged I am planning to fix issues related to subsref too.  Actually quite similar issues are there with subsref too!

inst/@sym/subsasgn.m
cbm755: Ok, or "2D indexing with length in one dimension more than 2".

inst/@sym/subsasgn.m
latot: Hi, note, i only choose nnz because there was 2 options, transform all the matrix to double, or count the number of zeros, i choose the last, if you think can help the other option no problem changing it.

Note 2, how the input matrix don't have any zero, count the number of zeros is the same as check the values changed in the matrices.

Thx. Cya.

cbm755: Hmm `nnz` is much faster than `double` currently.  Speeding up double would be a good idea, easier after pytave I'm sure, so I wasn't going to bother.

Actually just calling `isequal(A,B)` is twice as fast as converting to double (this will convert them both to sym).  Maybe that's the way to go here.

genuinelucifer:
@cbm755 the extra warnings were due to a mistake on my side. I issued the warning even when rhs was empty. I've fixed it now.
.
I think we should drop the warning as octave's double inherently allows subsasgn with rhs as a matrix.
.
Also, I have fixed the spacing issues now. Please review it.

inst/@sym/subsasgn.m
```diff
@@ -517,4 +518,71 @@
 %! a = sym([1, 2; 3, 4]);
 %! a(1, 2) = [];
 
+
+%% Tests from mat_replace
```
genuinelucifer: I have added this as al the tests here on after mostly relate to the changes done in mat_replace only.

inst/@sym/private/mat_replace.m
```diff
@@ -66,12 +79,7 @@
     end
   end
 
-  if (length(subs) == 1 && islogical(subs{1}))
-    %% A(logical) = B
-    z = mat_mask_asgn(A, subs{1}, b);
-    return
-
-  elseif (length(subs) == 1 && strcmp(subs{1}, ':') && length(b) == 1)
```
genuinelucifer: I have remove the first if as that condition is now taken care at the start of this method.

inst/@sym/private/mat_replace.m
```diff
@@ -28,6 +29,18 @@
...
 function z = mat_replace(A, subs, b)
 
+  if (length (subs) == 1 && islogical (subs{1}))
+    %% A(logical) = B
+    subs{1} = find (subs{1});
+    if isempty (subs{1})
+      z = A;
+      return;
+    end
+    if (~ isscalar (b) && ~ isvector (b) && ~ isempty (b))
```
genuinelucifer: Added an isempty condition to block the warnings in that case.

cbm755:
LGTM!  Merging.

Can we drop `private/mat_mask_asgn` now?

PR #691, opened 2016-11-14
Makefile changes

carandraug:
Main changes:

* use `:=` instead of `=` which does [lazy assignment](http://stackoverflow.com/a/448939/1609556)
* `make dist` instead of `make pkg` so it matches what other packages are doing (see [proposal](https://lists.gnu.org/archive/html/octave-maintainers/2016-11/msg00065.html) to make this target a requirement for for packages)
* the changed recipe for Octave package release and html is mostly to match the code from the other packages so that they all work in similar way which makes it easier when copying other recipes from other Makefiles of Octave Forge.


carandraug:
Added one more change that adds a recipe for the zip file.

cbm755:
This also closes #674.

cbm755:
Thanks @carandraug.

PR #692, opened 2016-11-14
594 Part 1

latot:
Hi!, well i try split that big pr, this is the first part..... still seems big...........

And i try split the commits, sadly i can't split it right, in every commit exist the main part of it, but in someones exist other parts i could not split.......

I don't split more this or we will start having problems with the rewrite, you know, basically we can't split a rewrite of it self........

Thx. Cya.


latot:
Hi, updated!, i remove some code for the future, and i put again the clear function but with deprecate message.

Thx. Cya.

cbm755:
Ok, so some of this is merged by #703 and #704.  Anything else from here needed?  I suggest other *small* PRs for remaining changes.

inst/@sym/assume.m
```diff
@@ -107,6 +107,8 @@
 
 function varargout = assume(x, varargin)
 
+  varargin = expr_to_str (varargin);
```
latot: ups test missed

latot:
Hi!, i think now i can start sending mini-prs, and some things from here i think we will need.

Thx- Cya.

PR #694, opened 2016-11-15
sympref: show argument value in error message

mtmiller:
Show a more explicit error message with the unrecognized argument value.

This looks more helpful to me than the standard "usage" message. In Octave we typically reserve `print_usage` for when the number or types of arguments are blatantly wrong, but not when the value is out of range or an unrecognized string enum like this function.


cbm755:
LGTM

PR #704, opened 2016-12-17
sym constructor rewrite (594 Part 1)

cbm755:
I did some refactoring of @latot's commits in #692 and #594.

I've commented a bunch of things out, and removed some other commits: those things can be proposed as new PRs: I'm trying to be conservative here while still moving these changes forward.


inst/@sym/sym.m
```diff
@@ -189,179 +188,224 @@
...
+  end
 
+  if (isa (x, 'char'))
+    % We now have a char; need to decide whether to use S() or Symbol() on it.
+
+    if (check)
+      % TODO: Warning if you try make a sym with the same name of a system function.
+      %symsnotfunc (x);
+
+      % TODO: tests pass without this?  Is there a example where this is needed?
```
latot: Hi, i don't know how do this test, basically, if you do this ```sym(-1)``` should not send warning, but i think we can do that in other pr, only lets hold this for the future (warning when we do ops from sym)

Thx. Cya.

cbm755: thanks, ok we'll just leave that comment for some future PR.

You ok with merging this?

inst/@sym/private/const_to_python_str.m
```diff
@@ -0,0 +1,99 @@
...
+%% @end deftypefun
+
+function [s, flag] = const_to_python_str (x)
+
+  persistent list
+  persistent const
+
+  if (isempty (list))
+    % Table of special doubles and special strings
+    % format: number, {octave strings}, resulting python expression
```
latot: if i remember right, here was a comment about always the python expression should be in the octave strings for can eval the expression in self, like ```sym(sympy(a))==a``` Are you sure you want remove that comment?

inst/@sym/sym.m
```diff
@@ -189,179 +188,224 @@
...
+      [err flag s] = python_cmd (strrep (cmd, '{s}', x));
+
+      switch (flag{1})
+        case 1  % Bad call to python function
+          error (['Python: %s\n' ...
+                  'Error occurred using "%s" Python function, perhaps use another variable name?'],
+                 err, s);
+        case 2  % Something else
+          error (['Python: %s\n' ...
+                  'Seems you cannot use "%s" for a variable name; perhaps this is a bug?'],
```
latot: Hi!, with this text i think is possible simplify the flag structure, just to flag:
```
+             '        return (str(e), (1, "" if len(lis) == 1 else "s"), "\", \"".join(str(e) for e in lis))'
 +             '    return (str(e), (2, 0), x)' };
```
to
```
+             '        return (str(e), 1, "\", \"".join(str(e) for e in lis))'
 +             '    return (str(e), 2, x)' };
```
and change ```flag{1}``` to ```flag```

Thx. Cya.

latot:
Hi, i think is almost ready, i put some comments, and for last, is possible keep some test for the FF issue?, like error when we try ```sym(FF(a))```.

Thx. Cya.

inst/@sym/private/const_to_python_str.m
cbm755: Hmm maybe I did by accident.  Not sure what comment you mean?  Was it in your #594 or in master?

cbm755:
There is a test for `sym(FF(a))` failing.

I fixed up the the `flag` error handling, thx.

inst/@sym/private/const_to_python_str.m
latot: Hi, in the pr, here the comment would be like:
```
% {octave string} need contain the python expression for sympy function
```
is in https://github.com/latot/octsympy/blob/200c70e48977c244561a82fc59a19bc369803e01/inst/%40sym/private/magic_double_str.m:36

Thx. Cya.

cbm755: Ok I see now.  Thanks, done.

latot:
Hi, i think this is ready, if i found any detail i think we can send other pr.

Thx. Cya.

cbm755:
Merging!

PR #708, opened 2016-12-17
Improve coeffs

cbm755:
improve compatibility with SMT: return "wrong" order, and implement "all" keyword.


cbm755:
@latot can you review this?

inst/@sym/coeffs.m
```diff
@@ -202,7 +248,7 @@
 %! assert (isequal (t, s))
 
 %!test
-%! % empty same as no specifying
+%! % empty same as not specifying
```
genuinelucifer: @cbm755 Hi, I was just trying to run these tests on MATLAB R2016a and it seems that MATLAB gives different outputs if the second argument is empty vs when it's not there.  
Example:  
```matlab
>> [c, t] = coeffs(6*x*x + 27*y*x  + 36, {})
c =
6*x^2 + 27*y*x + 36
t =
1
>> [c, t] = coeffs(6*x*x + 27*y*x  + 36)
c =
[ 6, 27, 36]
t =
[ x^2, x*y, 1]
>> 
```
( I just reducing the spacing in the output above).  
As it seems that the order of `c` and `t` are reversed in this case. Also, MATLAB doesn't give any error/warning if the `all` is misspelled. It just treats it as giving `{}` as the argument. In all cases where there is an argument and it is not exactly `all` then it gives the same result. So, `[c, t] = coeffs(6*x*x + 27*y*x  + 36, {});` gives the same result as `[c, t] = coeffs(6*x*x + 27*y*x  + 36, 'some weird stuff');`

cbm755: SMT is probably taking the 'someweirdstuff' and making it into a sym; that is, the same as `[c, t] = coeffs(6*x*x + 27*y*x + 36, sym('s'))`.  But SMT has a legacy of accepting `char` for what should be sym inputs that we don't necessarily need to support.  I agree giving an error seems more appropriate.

That empty case is kind of annoying!  I prefer our behaviour (empty is fairly commonly the same as omitting the argument in Octave/Matlab code). I will think about this and perhaps file a separate bug.

While testing these, I noticed that the new `'all'` doesn't work with multivariate polynomials :(

cbm755: Or, giving an error for empty 2nd argument seems ok to me too

inst/@sym/coeffs.m
```diff
@@ -78,75 +79,96 @@
...
+%% @strong{Note:} The @qcode{'all'} feature does not yet work with
+%% multivariate polynomials (https://github.com/cbm755/octsympy/issues/720).
 %%
 %% @seealso{@@sym/sym2poly}
 %% @end deftypemethod
 
-function [c, t] = coeffs(p, x)
+function [c, t] = coeffs(p, x, all)
 
-  if (nargin > 2)
```
latot: mm, if declare ```all=false;``` here and change it only when the conditions need it? (less lines)

inst/@sym/coeffs.m
```diff
@@ -78,75 +79,96 @@
...
+  if (nargin == 1)
+    % don't use symvar: if input has x, y we want both
+    x = {};
+    all = false;
+  elseif (nargin == 2)
+    if (ischar (x))
+      if (~ strcmpi (x, 'all'))
+        error ('coeffs: invalid 2nd input: if string, should be "all"')
```
latot: if replace this with an assert?

inst/@sym/coeffs.m
```diff
@@ -78,75 +79,96 @@
...
+      end
+      x = {};
+      all = true;
+    else
+      x = sym(x);
+      all = false;
+    end
+  elseif (nargin == 3)
+    if (~ strcmpi (all, 'all'))
+      error ('coeffs: invalid 3rd input: should be string "all"')
```
latot: same above

inst/@sym/coeffs.m
```diff
@@ -78,75 +79,96 @@
...
   if ~(isscalar(p))
     error('coeffs: works for scalar input only');
   end
 
-  cmd = { 'f = _ins[0]'
-          'xx = _ins[1]'
+  cmd = { '(f, xx, all) = _ins'
           'if xx == [] and f.is_constant():'  % special case
           '    xx = sympy.S("x")'
           'try:'
```
latot: maybe here instead a try use ```isinstance(xx, MatrixBase)```?

latot:
Hi, i write some comments, other things i saw you are already working on it.

inst/@sym/coeffs.m
cbm755: good idea, but we try to keep these m-files using Matlab dialect, no Octave extensions.

inst/@sym/coeffs.m
cbm755: ok, will do

inst/@sym/coeffs.m
cbm755: ok

inst/@sym/coeffs.m
cbm755: Could do.  I think xx can also be a list.  I think I was trying to be more pythonic, but I agree, this code looks wonky.  Will consider.

inst/@sym/coeffs.m
latot: mm, maybe:
```
all=false;
if (...)
  all=true;
```
???

cbm755: But its also a string input...  I guess I could use `varargin`.  I don't think it matters too much.

latot:
mm, the last thing, maybe add some warnings or error for functions actually don't works completely fine? (like mutivar, symvar, etc), i don't check exactly this, so only if is missed.

cbm755:
good idea, I added an `%!error` test which ensures the user gets a good error message.  A specific warning/error would be better but not so easy to do (because in this version we may or may not know if its multivariate until we're in Python).  

The `symvar` version being discussed in #721 would make this easy.

Issue #709, opened 2016-12-22
Octave's subsref seems to return weird sized empty matrices


genuinelucifer:
I was trying to fix the subsref behaviour for logical indices and ran into the following problem.
.
If we return empty matrix on this `a([false false])` where `a` is a `sym` then the size of that matrix will be `0x0` while octave returns (for `a` as `double`) a matrix of size `1x0`!! If the matrix is empty then does `1` in `1x0` make sense? Moreover, I can't replicate such matrices without hardcopying them, i.e., `sym(a[false false])`....
After the changes to fix logical indices, the results are:
```
>> a = [1 2 3 5; 4 5 6 9; 7 5 3 2];
>> b = sym(a);
>> c = [false false];
>> a(c)
ans = [](1x0)
>> b(c)
ans = [](0x0)
```

genuinelucifer:
Octave actually returns a "0x0" empty matrix if a scalar false is used in subsref.  
"1x0" empty matrix is returned if a row vector of all false is use.   
"0x1" empty matrix is returned for all other cases where all the values are false.   
Should we also follow this? Also how to produce an empty sym of such dimensions?

cbm755:
`sym('a', [0 1])` and `sym('a', [1 0])` and `sym('a', [0 0])`.

Actually, maybe `sym(zeros(0,1))` is nicer?

cbm755:
> Should we also follow this?

probably we should.  There are some failing tests: are those related?

genuinelucifer:
> There are some failing tests: are those related?

Yes three tests are failing because the result of same operation on double is not equal to the sym counterpart. I'll fix those and add the tests.

cbm755:
Is this closed?  I don't think its a problem in core Octave but we still have errors here right?

genuinelucifer:
I thought it's weird (and probably erroneous) behavior of octave. So I opened this. Since we decided to follow the same here so I closed this. I'll update #710 by tomorrow to follow this.

cbm755:
Fixed by 479522393a909ba1f199eddb77b576b960b4ccd1, thanks!

PR #710, opened 2016-12-22
Fix subsref behaviour when logical indices are used 

genuinelucifer:
Fixing this lead me to an issue which I have described in #709 .. I will add tests to this PR after we come to a decision on how to fix the issue as the tests will fail because of that...
.
Moreover, I have deleted the 2 files that are no more needed (both because of logical indexing). One was leftover from my PR for subsasgn fix.


inst/@sym/subsref.m
```diff
@@ -1,4 +1,6 @@
 %% Copyright (C) 2014, 2016 Colin B. Macdonald
+%% Copyright (C) 2016 Lagu
+%% Copyright (C) 2016 Abhinav Tripathi
```
genuinelucifer: @latot Since you wrote the tests so I added your name in the copyright.

inst/@sym/private/mat_access.m
```diff
@@ -1,4 +1,5 @@
-%% Copyright (C) 2014 Colin B. Macdonald
+%% Copyright (C) 2014-2016 Colin B. Macdonald
+%% Copyright (C) 2016 Abhinav Tripathi
```
genuinelucifer: @cbm755 I updated the copyright from `2014` to `2014-2016`.. I hope it's okay.

genuinelucifer:
I just wanted to get it clarified. I forgot what we exactly decided to use here but I think we decided to not use braces on `if` if there is only 1 condition. So I have used `if ~ is_same_shape (A, subs{1})` .... Please let me know if this should have braces (probably because if contains `~`).

inst/@sym/private/mat_access.m
cbm755: I list copyright for years where the file was actually modified.  So I would leave this as "2014 Colin..." because I didn't modify in 2015 and made only trivial changes in 2016.

cbm755:
Hmm, I don't recall.  We should probably use upstream Octave convention (whatever that is) for new code.

inst/@sym/private/mat_access.m
genuinelucifer: Oh ok. I will fix this by tonight and update the PR.

genuinelucifer:
As per http://wiki.octave.org/Octave_style_guide#ending_blocks we should use braces for `if` and not for the iteration in `for`. I will update this PR with this style.

genuinelucifer:
@cbm755 I have done the required changes. Please review.

cbm755:
Great! Merging

PR #715, opened 2017-01-11
Fixes for Octave >= 4.3

cbm755:
@latot and @mtmiller do these look ok?


latot:
Hi, i think is fine.

cbm755:
thanks, merging

mtmiller:
👍 from me post hoc

PR #721, opened 2017-01-14
use symvar in coeffs

cbm755:
On top of PR #708.


inst/@sym/coeffs.m
```diff
@@ -78,75 +79,102 @@
...
-          'if xx == [] and f.is_constant():'  % special case
-          '    xx = sympy.S("x")'
+  p = sym(p);
+
+  if (isempty (x))
+    x = symvar (p);
+    if (isempty (x))
+      x = sym('x');  % any symbol
+    end
+  end
```
latot: i think can be better do this in python, create a var from octave is more expensive than do it in the code.

cbm755: True, but logic is in one little block here...  And moving this block to Python tricky b/c `symvar` is non-trivial.

latot: mm, and sens symvar as a new param to python and then decide if use it or not?

cbm755: could do, not sure it would be easier to understand or not.

Bigger picture here is whether we want to do this change or keep `coeffs` behaving more like Sympy `Poly` objects.  SMT compat says we should make this change... but I'm not sure.

cbm755:
I rebased and force pushed.

cbm755:
I got bit my this doing `coeffs(pi*x + x)` and having it use `pi` as a part of the field instead of a coeff.  I don't think this is how SMT users expect this to work, so I think I will merge this.

Dealing with Polynomials in a more general way (as SymPy) does is larger in scope than just this function: it would be a bigger project.  So if we have to revisit this later so be it.

Will when tests pass.

Issue #627, opened 2016-10-17
assume(x) and assumeAlso(x) don't return the correct error


latot:
Hi:

```
octave:24> assume(a)
error: 'ca' undefined near line 116 column 20
error: called from
    assume at line 116 column 8
octave:24> assumeAlso(a)
error: sym: assumption must be a string
error: called from
    assert at line 94 column 11
    sym at line 347 column 11
    assumeAlso at line 92 column 8
```

Thx. Cya.

cbm755:
Is this supposed to:
A. do the same as `assume(x, 'clear')`
or 
B. list the assumptions on x?

latot:
C. Return an error of not supported yet, send 1 var implies set the condition with 1 expression, ex:

```
assume(x>0)
```

cbm755:
Ah I see.   Unlikely to be supported either, given differences in assumptions system between SymPy and MuPAD.

Yes, error message sounds good, something like "general algebraic assumptions are not supported".

latot:
pr for this https://github.com/cbm755/octsympy/pull/724

PR #724, opened 2017-01-15
assert for unsupported command

latot:
Fixes https://github.com/cbm755/octsympy/issues/627


cbm755:
Great!

Just to note: we could probably support `assume (x < 0)`.  Much harder would be `assume (3*x < sin(y))`.

Issue #621, opened 2016-10-17
Negative assumptions don't are showed correctly from assumptions.


latot:
Hi:

```
octave:12> a=python_cmd('return Symbol("a", polar=False)')
a = (sym) a
octave:13> assumptions
ans = 
{
  [1,1] = a: {'polar': False}
}
```

Thx. Cya.

latot:
pr for this https://github.com/cbm755/octsympy/pull/725

PR #725, opened 2017-01-15
show negative assumptions correctly

latot:
Fixes https://github.com/cbm755/octsympy/issues/621

The proposed format to negate assumptions is add the prefix ```~``` to the assumption, i think can follow the octave logic.

### DIFF
--- a/inst/assumptions.m
+++ b/inst/assumptions.m
@@ -111,10 +111,8 @@ function [A, B] = assumptions(F, outp)
       'd = x._assumptions.generator'
       'if d == {}:'
       '    astr = ""'
-      'elif all(d.values()):'  % all True so list them
-      '    astr = ", ".join(sorted([str(i) for i in d.keys()]))'
-      'else:'  % more complicated case, just the raw dict
-      '    astr = str(d)'
+      'else:'
+      '    astr = ", ".join(sorted([("" if v else "~") + str(k) for (k,v) in list(d.items())]))'
       'if outputdict:'
       '    return (astr, d)'
       'else:'
@@ -250,3 +248,8 @@ end
 %! % multiple assumptions: eqn neither true nor false
 %! n = sym('n', 'negative', 'even');
 %! assert (~isequal (n, sym(true)) && ~isequal (n, sym(false)))
+
+%!test
+%! %% TODO: rewrite later with https://github.com/cbm755/octsympy/issues/622
+%! a = python_cmd ('return Symbol("a", real=False)');
+%! assert (strcmp (assumptions (a), {'a: ~real'}))


inst/assumptions.m
```diff
@@ -111,10 +111,9 @@
       'd = x._assumptions.generator'
       'if d == {}:'
       '    astr = ""'
-      'elif all(d.values()):'  % all True so list them
-      '    astr = ", ".join(sorted([str(i) for i in d.keys()]))'
-      'else:'  % more complicated case, just the raw dict
-      '    astr = str(d)'
+      'else:'  % all True so list them
+      '    val, key = list(d.values()), list(d.keys())' % In Python3 d.values() don't support indexing
+      '    astr = ", ".join(sorted([(("" if val[i] else "~") + str(key[i])) for i in range(len(d.keys()))]))'
```
cbm755: I forget if Sympy assumptions have to be either true or false... Can they be None?

I think this code assumes each value of the dict is either true or false.

latot: Yes, theoretically only can be true or false, and that says the doc, i test this with none and only causes irregular things and bugs...., so i don't think we will get that value in the future.

cbm755: I see.

The comment "all True so list them" isn't correct anymore: remove it.

cbm755: Instead of indexing with `i` can this be something like `[... for (v, k) in zip(list(d.values(), list(d.keys()))]`?

or `[("" if d[k] else "~") + str(k) for k in d]`?

or `[("" if v else "~") + str(k) for (k,v) in d.items()]`?

(haven't tested which are portable between 2 and 3)

inst/assumptions.m
```diff
@@ -250,3 +249,8 @@
 %! % multiple assumptions: eqn neither true nor false
 %! n = sym('n', 'negative', 'even');
 %! assert (~isequal (n, sym(true)) && ~isequal (n, sym(false)))
+
+%!test
+%! %% rewrite later: https://github.com/cbm755/octsympy/issues/622
```
cbm755: stick a `TODO` in this comment so we find it later.

latot:
okidokis!

inst/assumptions.m
```diff
@@ -111,10 +111,8 @@
       'd = x._assumptions.generator'
       'if d == {}:'
       '    astr = ""'
-      'elif all(d.values()):'  % all True so list them
-      '    astr = ", ".join(sorted([str(i) for i in d.keys()]))'
-      'else:'  % more complicated case, just the raw dict
-      '    astr = str(d)'
+      'else:'
+      '    astr = ", ".join(sorted([("" if v else "~") + str(k) for (k,v) in list(d.items())]))'
```
cbm755: `list()` probably unnecessary here, but its not a big deal

PR #731, opened 2017-01-17
sym constructor: add deprecation warning for sym(cell)

cbm755:
Basically this PR is to top using `varargin = sym(varargin)` and make calls to `sym(cell)` give a warning.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -36,6 +36,9 @@ octsympy 2.5.0-dev
           fresnelc      sinint
           fresnels      zeta
 
+  * Deprecation notice: `sym(cell)` will soon create a sym array
+    (currently it creates a cell array of sym objects.)
+
   * The full set of assumptions implemented by Sympy can now be used.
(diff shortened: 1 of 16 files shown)


cbm755:
@genuinelucifer and @latot please review.

@latot I imagine you would prefer a helper function called something like `cell_to 
_cell_of_syms`.  I'd probably consider a PR that does that, but for now, here is my solution.

genuinelucifer:
:+1:  
Looks good to me. Will be quite handy to finally get rid of `symarray`.

cbm755:
ok,merging.

latot:
mm, related with this, maybe improve num2cell to can do ```num2cell(sym(varargin))```?

if i remember right in SMT cells are transformed to arrays.

Thx. Cya.

cbm755:
In fact, core Octave's num2cell seems to DTRT for sym arrays!  A pleasant surprise.  I will add a test for it.

Thanks for this.

PR #732, opened 2017-01-17
sympref: removed references to whichpython (fixes #723)

### LINKED ISSUE #723: sympref: remove references to `whichpython`
I think this is leftover from previous functionality, should be removed.

nvs-abhilash:
I just removed the references in inst/sympref.m Is there some more stuff to remove or this is it for #723  ?


cbm755:
That's all as far as I know, thanks!

PR #736, opened 2017-01-17
subsref: make output dims agree if input is vector

### LINKED ISSUE #735: non-deterministic test failures (again)
Deja vu, but I'm seeing non-deterministic test failures from `@sym/subsref`:

````
>>>>> processing @sym/subsref.m

***** test

 % 1D arrays, does right with despite warning

 r = [1:6];

 ar = sym(r);

 c = r';

 ac = sym(c);

 Ir = rand(size(r)) > 0.5;

 Ic = rand(size(c)) > 0.5;

 assert (isequal (ar(Ir), r(Ir)))

 assert (isequal (ac(Ic), c(Ic)))

 s = warning ('off', 'OctSymPy:subsref:index_matrix_not_same_shape');

 assert (isequal (ar(Ic), r(Ic)))

 assert (isequal (ac(Ir), c(Ir)))

 warning (s)

!!!!! test failed

assert (isequal (ac (Ir), c (Ir))) failed
````
Last time, this was caused by `rand` choosing some "corner case" (e.g., such as all true or all false).  Maybe @nvs-abhilash (or someone else) feels like playing detective to track this down to a simpler case?

mtmiller:
Something like this should fix #735.

Needs some tests added in subsref.m for the degenerate cases that are actually causing the random failures. Any improvements or suggestions to what I have so far?

### DIFF
--- a/inst/@sym/private/mat_access.m
+++ b/inst/@sym/private/mat_access.m
@@ -38,8 +38,17 @@ function z = mat_access(A, subs)
                'A and I in A(I) not same shape: no problem, but did you intend this?')
     end
     subs{1} = find (subs{1});
-    if (isempty (subs{1}))
-      z = sym (zeros (size (subs{1})));
+    idx = subs{1};
+    if (isempty (idx))
+      % fix the dimensions when both A and idx are vectors
+      if (max (size (idx)) > 0)
+        if (iscolumn (A))
+          idx = idx(:);
+        elseif (isrow (A))
+          idx = idx(:)';
+        end
+      end
+      z = sym (zeros (size (idx)));
       return;
     end
 
--- a/inst/@sym/subsref.m
+++ b/inst/@sym/subsref.m
@@ -336,3 +336,18 @@ end
 %! test_bool (a, b, c)
 %! test_bool (a, b, c | true)
 %! test_bool (a, b, c & false)
+
+%!test
+%! % Orientation of empty results of logical indexing on row or column vectors
+%! r = [1:6];
+%! c = r';
+%! ar = sym(r);
+%! ac = sym(c);
+%! s = warning ('off', 'OctSymPy:subsref:index_matrix_not_same_shape');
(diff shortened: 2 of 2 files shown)


cbm755:
@genuinelucifer last changed that code so let's get his opinion.

Tests look good to me.

genuinelucifer:
I haven't tested it but It looks good.  
  
I have just one question though, I used size of find which is actually what octve would return in such cases. Why isn't that enough? Sorry for not testing anything. You ca merge this, I'll test it in evening and post here if needed.

cbm755:
@genuinelucifer no hurry, I'll wait until you've looked it over.

mtmiller:
@genuinelucifer the size of find is correct in most cases.

    >> size (find (false (1, 10)))
    ans =
    
       1   0
    
    >> size (find (false (10, 1)))
    ans =
    
       0   1
    

That is correct and good enough when `A` is a two-dimensional matrix.

The problem that this change is fixing is when `A` happens to be a row or column vector, then the dimensions of it take precedence over the dimensions of the logical indexing array. The singleton-ness of `A` dictates what the shape of the empty output array should be.

    >> size ((ones (10, 1))(false (10, 1)))
    ans =
    
       0   1
    
    >> size ((ones (10, 1))(false (1, 10)))
    ans =
    
       0   1
    

Your change was correct and you didn't do anything wrong. The only thing missing was tests for the case when a vector is indexed by another logical vector, and the shapes either agree or don't agree.

mtmiller:
@cbm755 I plan on squashing these commits, so don't merge until I remove WIP from the title.

genuinelucifer:
> The problem that this change is fixing is when A happens to be a row or column vector, then the dimensions of it take precedence over the dimensions of the logical indexing array.

Thanks. I get it now. Looks great! :+1:

mtmiller:
Rewritten and ready to merge, thanks all

Issue #616, opened 2016-10-17
assume([x y], 'real') don't works


latot:
Hi:

```
octave:4> assume([x y], 'real')
error: assert (isempty (asm)) failed
error: called from
    assert at line 92 column 11
    sym at line 327 column 7
    assume at line 116 column 8
```

Thx. Cya.

cbm755:
Do you know if SMT supports this?

mtmiller:
Someone should test to confirm, but looks like it from the [docs](https://www.mathworks.com/help/symbolic/assume.html):

> If `expr` is a vector or matrix, then `assume(expr,set)` sets an assumption that each element of `expr` belongs to `set`.

cbm755:
might be nice for 2.5.0 to help with the deprecation of `syms x y clear`.

@latot do you have time to do a PR for this?  Or maybe @nvs-abhilash wants to try a deep-dive into our messy horrible no-good assumptions hackery.

PR #737, opened 2017-01-18
[WIP] add vector support to assume and assumeAlso

latot:
Fixes https://github.com/cbm755/octsympy/issues/616

Hi all, assume now works, but sadly i can't found the problem with assumeAlso....

this don't works
```
syms x y positive
assumeAlso([x y], 'real')
```

and when we check the change we can see don't happen anything....

The new sym is created correctly, and the right assumptions, but don't are assigned to the workspace for some reason.

Thx. Cya.

### DIFF
--- a/inst/@sym/assumeAlso.m
+++ b/inst/@sym/assumeAlso.m
@@ -64,52 +64,55 @@
 %% @seealso{@@sym/assume, assumptions, sym, syms}
 %% @end deftypemethod
 
-%% Author: Colin B. Macdonald
-%% Keywords: symbolic
 
 function varargout = assumeAlso(x, varargin)
 
   assert (nargin > 1, 'General algebraic assumptions are not supported');
 
-  [tilde,ca] = assumptions(x, 'dict');
+  for i = 1:length (x)
 
-  if isempty(ca)
-    ca = [];
-  elseif (length(ca)==1)
-    ca = ca{1};
-  else
-    ca
-    error('expected at most one dict')
+    xstr = subsref(x, substruct('()', {i}));
+    [tilde,ca] = assumptions(xstr, 'dict');
+
+    if isempty(ca)
+      ca = [];
+    elseif (length(ca)==1)
+      ca = ca{1};
+    else
+      ca
+      error('expected at most one dict')
+    end
+
+    for n=2:nargin
+      cond = varargin{n-1};
+      ca.(cond) = true;
+    end
+
+    xstr = xstr.flat;
+    newx = sym(xstr, ca);
+
+    if (nargout > 0)
+      varargout{1} = newx;
+      return
+    end
+
+    % ---------------------------------------------
+    % Muck around in the caller's namespace, replacing syms
+    % that match 'xstr' (a string) with the 'newx' sym.
+    %xstr =	
+    %newx =
+    context = 'caller';
+    % ---------------------------------------------
+    S = evalin(context, 'whos');
+    evalin(context, '[];');  % clear 'ans'
+    for i = 1:numel(S)
+      obj = evalin(context, S(i).name);
+      [newobj, flag] = symreplace(obj, xstr, newx);
+      if flag, assignin(context, S(i).name, newobj); end
+    end
+    % ---------------------------------------------
+
   end
-
-  for n=2:nargin
-    cond = varargin{n-1};
-    ca.(cond) = true;
-  end
-
-  xstr = x.flat;
-  newx = sym(xstr, ca);
-
-  if (nargout > 0)
-    varargout{1} = newx;
-    return
-  end
-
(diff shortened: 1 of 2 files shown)


cbm755:
I don't think this your fault: same things happens with just one variable in 2.4.0.

 (I guess because positive implies real, although `syms x real positive` does have both true.)

cbm755:
@latot, what do you think `q = assumeAlso([x y], 'real')` should do?  I think it gives `q = x` but should it be `q = [x y]`?

latot:
mm, Hi, i was asking the same to my self, that don't is documented in smt, but i don't know very well the behavior, some part lost sense, i'm thinking 3 possibilities, one ```q=[x y]``` other ```[x y] = [x y]``` personally i think have more sense the last option, because assume don't use the matrix as element, only to collect the vars to work on it, and the last option is maybe smt don't do this any of this and maybe something "undefined"? like an error or something.

checking assume i found this:
```
octave:2> assume(a, 'positive' ,'real')
octave:3> assumptions(a)
ans = 
{
  [1,1] = a: positive, real
}
```

when i check the bucle from assumeAlso the assignation of the assuptions works, but fails when the script need assign the var to the workspace, i don't know why fails, checking i found other problem when we assign more than one var to more than one assumption, assume and assumeAlso fails with that, and both have the same behavior i write above..., i'll need check better this and add more tests.

but still i think is weird only fails to assign to the workspace but internally works fine, and the conditions to this i feel are too irregular.

Thx. Cya.

inst/@sym/assumeAlso.m
```diff
@@ -64,53 +64,56 @@
...
-%% Keywords: symbolic
 
 function varargout = assumeAlso(x, varargin)
 
   assert (nargin > 1, 'General algebraic assumptions are not supported');
 
-  [tilde,ca] = assumptions(x, 'dict');
+  for i = 1:length (x)
+
+    xstr = subsref(x, substruct('()', {i}));
```
cbm755: can we do this as:
````
for i = 1:length (xx)
  x = subsref(xx, ...
  [tilde,ca] = assumptions(x, 'dict');
````
that way, no changes are made to the code below.

File: inst/@sym/assumeAlso.m [hunk 2]
```diff
@@ -165,3 +168,9 @@
 %!error <General algebraic assumptions are not supported>
 %! syms a
 %! assumeAlso (a>0)
+
+%!test
+%! syms x y positive
+%! assumeAlso ([x y], 'real')
```
cbm755: So for now, let's use something like `'even'` here.

cbm755:
I agree it doesn't look like SMT's `assume` has outputs at all (if so its not documented).

Maybe `[x y] = assume([x y], 'even')` should work?

There  is a bug where `return` is called too early.  I suggest:

````
  if (nargout > 0)
    varargout{i} = newx;
  else
    % ---------------------------------------------
    % Muck around in the caller's namespace, replacing syms
    ...
  end
````
(note: no mucking around when there are outputs).

**Comment by cbm755** [comment]
> only fails to assign to the workspace but internally works fine

I suggest you test with something other than `real` and `positive` b/c of #739.

**Comment by latot** [comment]
Hi, you are right, testing it with even pass and works, now i don't know why now fails the test in travis, locally works fine here...

And about the bug, i check the code and the return i think is ok, your proposes structure is equal to the actual code, are you sure want to change it?

File: inst/@sym/assumeAlso.m (hunk 1 above)
Comment (latot) [inline inst/@sym/assumeAlso.m]: mm, taht will change all the other syntax.., or add (at least to me) useless lines, because finally this will be:
```
x = ...
[tilde, ca] ...
xstr = x
```
or if i replace all xstr to x i will need change the names of the vars in assume and assumeAlso, i think you prefer the last option? (or keep how is now)

cbm755:
works fine for me too.  Maybe some octave 4.0.2 specific problem.  your strcmp compares a string and a cell array of a string.  Maybe try:
````
strcmp('x: even', assumptions(x){1})
````

> your proposes structure is equal to the actual code, are you sure want to change it?

Sorry, I don't understand.

cbm755:
> your proposes structure is equal to the actual code, are you sure want to change it?

Actually, I understand now.  Yes I like it better because the original calls `ca = assumptions (x, dict)` and your version calls `ca = assumptions (xstr, dict)`.  Maybe that's about the same, but simpler to not change it.

inst/@sym/assumeAlso.m
cbm755:
````
x = xx{i}
[tilde, ca] = ...
xstr = x.flat
````
(xstr is the string of x)

cbm755:
Fixed by #746 which used a rebased version of these commits---thanks!

Issue #603, opened 2016-10-04
Why is `cell = sym(cell)` supported?


cbm755:
I added this in 49a45ddc.  No explanation why.  I wonder if it was for internal use. and then figured it was cute/useful to expose it.

The commit also has BIST that are still around in `@sym/sym` I think.

@mtmiller thinks this is a bad idea/illegal.  Perhaps we should try removing it and see what breaks?

It is documented in `@sym/sym` so it should probably be deprecated rather than dropped outright.

cbm755:
Interestingly, SMT seems to allow flat cells of strings, and the output is a symbolic array not a cell array:
````
>> sym({'a' 'b'}, 'positive')          
 
ans =
 
[ a, b]
 
>> assumptions
 
ans =
 
[ 0 < a, 0 < b]
 
>> 
````
So there we could paint this as an SMT compatibility issue...   E.g., I think @latot was trying to make assumptions work with cells.

latot:
Hi, yes, my actual pr support it, but exist some point to discuss, like, in the classdef we move ```sym(cell)``` to ```sym.symarray()```, so in some way we can't keep SMT compat with that, the other point is the return is a array, we actually use it to normalize the varargin and will break the behavior, so, lets do two functions with different outputs? or just disable it from octsympy and keep the normal way to varargin?

mtmiller:
I think doing a cellstr version of the constructor for undocumented Matlab compatibility looks good.

We could keep around a static method for building a cell array of syms from a cell array of names/values, even though I'm not sure what the use case would be.

cbm755:
f3d18449 abuses `cell = sym(cell)` for at least horzcat and vertcat, maybe others.  The original loops that commit replaced are more correct.

cbm755:
Basically `varargin = sym(varargin);` should be a loop.

cbm755:
I think for 2.5.0, we'd better put a deprecation notice in place so we can change this behaviour soon after.

cbm755:
TODO: `help @sym/sym` should include an example of the new behaviour.

cbm755:
Ok, I've done most of this and left a deprecation warning for 2.5.0: removing the 2.5.0 milestone.

After 2.5.0, we will change it to output a sym array, then we can close this.

PR #743, opened 2017-01-23
Docs and tests

cbm755:


latot:
what do you think to replace
```
for i...
  varargin{i} = sym(varargin{i});
end
```
to
```
varargin=num2cell(sym(varargin))
```
?

cbm755:
I support we could do that, after #603 (right now it will give a warning)...

But:
1.  its kind of messy: cell -> array -> cell.
2.  would need to be careful about what happens to matrices in the varargin.

So maybe its not worth it.

Issue #739, opened 2017-01-19
subs sometimes replaces assumptions with (equivalent?) ones


cbm755:
See https://github.com/cbm755/octsympy/pull/737#issue-201506688.  I tracked this down to Sympy's subs command:

In Sympy:
````
In [63]: x = Symbol('x', even=True, positive=True)

In [64]: x2 = Symbol('x', even=True, positive=True, real=True)

In [65]: srepr(x)
Out[65]: "Symbol('x', even=True, positive=True)"

In [66]: srepr(x2)
Out[66]: "Symbol('x', real=True, even=True, positive=True)"

In [67]: f = 2*x

In [68]: srepr(f.subs(x, x2))
Out[68]: "Mul(Integer(2), Symbol('x', even=True, positive=True))"
````
Note that the redundent `real=True` is gone.

If you really do change the symbol, it doesn't happen:
````
In [69]: x3 = Symbol('x', odd=True, positive=True, real=True)

In [70]: srepr(f.subs(x, x3))
Out[70]: "Mul(Integer(2), Symbol('x', real=True, positive=True, odd=True))"
````

I should probably track down whether this is deliberate optimization or a bug...

cbm755:
I think this is something to do with caching in sympy.  Basically `x == x2` is True and I think this might imply x2 and x can get mixed up in the cache (they are equivalent).

Here's a simple example:
````
In [1]: from sympy import *

In [2]: x = Symbol('x', positive=True)

In [3]: x2 = Symbol('x', positive=True, real=True)

In [4]: f = 2*x

In [5]: srepr(f)
Out[5]: "Mul(Integer(2), Symbol('x', positive=True))"

In [6]: srepr(f.func(2,x2))
Out[6]: "Mul(Integer(2), Symbol('x', real=True, positive=True))"

In [7]: srepr(2*x2)
Out[7]: "Mul(Integer(2), Symbol('x', positive=True))"
````
note last result is giving `2*x2`.

PR #746, opened 2017-01-25
add vector support to assume and assumeAlso

cbm755:
This is a rebase and bug fix of #737.

### DIFF
--- a/inst/@sym/assume.m
+++ b/inst/@sym/assume.m
@@ -21,8 +21,10 @@
 %% @documentencoding UTF-8
 %% @deftypemethod  @@sym {@var{x} =} assume (@var{x}, @var{cond}, @var{cond2}, @dots{})
 %% @deftypemethodx @@sym {@var{x} =} assume (@var{x}, 'clear')
-%% @deftypemethodx @@sym {} assume (@var{x}, @var{cond1}, @var{cond2}, @dots{})
+%% @deftypemethodx @@sym {[@var{x}, @var{y}] =} assume ([@var{x} @var{y}], @dots{})
+%% @deftypemethodx @@sym {} assume (@var{x}, @var{cond}, @var{cond2}, @dots{})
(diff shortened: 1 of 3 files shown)


cbm755:
I also rewrote the commits to make it clearer this is a minimal change.  I did this by doing the pure indent changes as separate commits.

cbm755:
@latot  ok to merge?

inst/@sym/assume.m
```diff
@@ -116,47 +116,48 @@
...
     end
-    newx = sym(xstr, ca);
-  end
 
-  if (nargout > 0)
-    varargout{1} = newx;
-    return
-  end
+    if (nargout > 0)
+      varargout{i} = newx;
```
latot: little question here, if we do
```
syms x
f=x*2;
x = assume(x, 'positive')
```
```f``` will be updated with the new assumptions?

if not will be updated i think is missed the ```return```, and if not is fine.

(This for assume and assumeAlso)

Thx. Cya.

Comment (cbm755) [inline inst/@sym/assume.m]: `f` should not be updated.  `assumptions(f)` should be empty.  I believe this is tested.

I don't think the return is missing: the "mucking around" bit is in a "else" block.

Comment (cbm755) [inline inst/@sym/assume.m]: Thanks for looking at this.  I've double-checked and those cases are indeed tested.

**Comment by latot** [comment]
Hi, the last thing, can you add one more test to check assume and assumeAlso with 2 new assuptions please?

```
assume(x, 'even', 'positive')
assumeAlso(x, 'even', 'positive')
```

like that, i remember there was some weird things, only to test it.

Thx. Cya.

cbm755:
Ok, done.

(The "weird things" are probably `'real', 'positive'` and `'even', 'real'` where `'real'` is redundant and sometimes non-deterministically disappears #739.)

PR #747, opened 2017-01-25
assume: support command form ("assume x y ...")

### LINKED ISSUE #421: support `assume x real` as well as `assume(x, 'real')`

cbm755:
This is done by adding a new assume.m that is not a method of the sym
class.  This must be the case because in "assume x y real", x and y
are strings.  But why not remove the method "@sym/assume" with the new
"assume.m"?  Well, for now "assume(x, y, 'real')" gives an error, which
is compatible with SMT behaviour.  It also means the assume *method*
does not need to locate the boundary between variables and conditions,
a process which could introduce bugs.  Maybe in some future commit we
will unify the two .m files, but for now they are separate.

Fixes #421.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -46,7 +46,8 @@ octsympy 2.5.0-dev
     Previously undocumented commands such as `sym(x, 'clear')` and
     `syms x clear` will be deprecated in a future version.
 
-  * `assume` supports vector inputs.
+  * `assume` supports vector inputs.  Also supported are alternative
+    "command forms" such as `assume x y real` and `assume x y clear`.
 
   * `solve` now works with inequalities.
 
--- a/inst/@sym/assume.m
+++ b/inst/@sym/assume.m
@@ -127,7 +127,7 @@ function varargout = assume(xx, varargin)
     assert (ischar (varargin{n-1}), 'assume: conditions should be specified as strings')
   end
 
-  for i = 1:length (xx)
+  for i = 1:numel (xx)
     x = subsref (xx, substruct('()', {i}));
     xstr = x.flat;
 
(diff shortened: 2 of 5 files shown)


cbm755:
@latot what do you think of this?

latot:
Nice!, the only question what should do this ```assume [x y] positive```?

cbm755:
That's a good point.  I think its a syntax error.  Either that or it will call `assume('[x', 'y]', 'positive')`, for which we will give an error.  This is the job of the Octave parser; outside of our control.

This will fail:
````
>> assume('[x y]', 'positive')
error: assume: only symbols can have assumptions
````
(in contrast with `assume([x y], 'positive')` which will work.

mtmiller:
> `assume [x y] positive`?

Yes, that is a syntax error, just as `assume {x y} positive` or `assume (x y) positive` would be. These bracket operators all bind to the first token `assume` the same as if there were no space at all.

latot:
Hi, so is necessary send error for ```a=[x y];assume a real````

cbm755:
Hmm, that is a tricky one.  `assume(a, 'real')` will make both x and y real.  But `assume a real` will quietly do nothing.

Consider:
`````
f = 2*sym('a');
a = [sym('x') sym('y')]
assume a real
assumptions(f)   % will say a is real
assumptions(x)    % empty
assumptions(y)    % empty
````
(contrast that with `assume(a, 'real')` which would be the opposite).

So I guess the question is, are these potentially confusing cases worth it?  (see pros/cons in https://github.com/cbm755/octsympy/issues/421#issuecomment-273264769)

latot:
personally is confusing, i would like support matrix in both.

mtmiller:
Yes I also find this confusing, I would expect

    assume a real
    assume (a, 'real')

to do the same thing.

cbm755:
Hmmm, so by extension, I guess you'd both expect both of these to do the same thing:
````
syms x
y = x
assume (y, 'real')
assume y real 
`````
(currently, `assume y real` means all sym('y') become real).

What would you expect to happen here?
````
f = 2*sym('a')
assume a real
````
Currently, `f` would then depend on a real symbol.  But based on above, it should be either a no-op or an error b/c there is no variable *named* `a`.

mtmiller:
Yeah absolutely. I guess it makes more sense to me that the command syntax is operating on the names of variables in the workspace, not on symbol names that may or may not exist in the workspace.

    a = sym('x');
    assume x real

This doesn't seem right to me. Should be

    a = sym('x');
    assume a real

cbm755:
Ok, I've implemented this and I think I've come around to this way of thinking!

PR #749, opened 2017-01-27
added isequal for @symfun class (fixes #740)

### LINKED ISSUE #740: isequal for symfun should check variables
As @genuinelucifer has noted elsewhere, `symfun` may need its own `isequal`.  Here is a buggy case:
````
>> syms x y
>> f(x) = 2*x;
>> g(x,y) = 2*x;
>> isequal(f, g)
1
````
(those should probably not be equal).

When tackling this, be careful with `isequaln` and `isequal`: perhaps its enough to overload one of them and rely on the superclass (`@sym`) implementation for the other?

### PR DESCRIPTION
**Pull request by nvs-abhilash**
Added isequal to @symfun to compare the `vars` of two `symfun` if they are not equal will return false otherwise will delegate the task of further checking to `isequaln`

### DIFF
--- a/inst/@symfun/isequal.m
+++ b/inst/@symfun/isequal.m
@@ -50,16 +50,11 @@
 
 function t = isequal(x, y, varargin)
 
-  t = logical(1);
-  if isa (x, 'symfun') && isa (y, 'symfun')
-    t = isequal (argnames(x), argnames(y));
+  t = isequal(formula(x), formula(y)) && isequal(argnames(x), argnames(y));
 
-    if nargin >= 3 && t == 1
-      t = t & isequal(x, varargin{:});
-    end
+  if nargin >= 3 && t
+    t = t & isequal(x, varargin{:});
   end
-
-  t = t & isequal(formula(x), formula(y));
 end
 
 
--- a/inst/@symfun/isequal.m
+++ b/inst/@symfun/isequal.m
@@ -1,0 +1,96 @@
+%% Copyright (C) 2017 NVS Abhilash
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defmethod  @@symfun isequal (@var{f}, @var{g})
+%% @defmethodx @@symfun isequal (@var{f}, @var{g}, @dots{})
+%% Test if contents of two or more arrays are equal.
+%%
+%% Example:
+%% @example
+%% @group
+%% syms x
+%% f(x) = x + 1;
+%% g(x) = x + 1;
+%% isequal(f, g)
+%%   @result{} 1
+%% @end group
+%% @end example
+%%
+%% Note: two symfuns are equal if they have the same formula @emph{and}
+%% same arguments:
+%% @example
+%% @group
+%% syms x y
+%% f(x) = x + 1;
+%% g(x, y) = x + 1;
+%% isequal(f, g)
+%%   @result{} 0
+%% @end group
+%% @end example
+%%
+%% @seealso{@@sym/isequal, @@symfun/formula, @@symfun/argnames}
+%% @end defmethod
+
+function t = isequal(x, y, varargin)
+
+  t = logical(1);
+  if isa (x, 'symfun') && isa (y, 'symfun')
+    t = isequal (argnames(x), argnames(y));
+
+    if nargin >= 3 && t == 1
+      t = t & isequal(x, varargin{:});
+    end
+  end
+
+  t = t & isequal(formula(x), formula(y));
+end
+
+
+%!test
+%! syms x y
+%! f(x) = 2*x;
+%! g(x) = 2*x;
+%! assert (isequal (f, g))
+
+%!test
+%! syms x
+%! f(x) = 2*x + 1;
+%! g(x) = 2*x + 1;
+%! h(x) = 2*x + 1;
+%! assert (isequal (f, g, h))
+
+%!test
+%! syms x
+%! f(x) = 2*x + 1;
+%! g(x) = 2*x + 1;
+%! h(x) = 2*x;
+%! assert (~ isequal (f, g, h))
+
+%!test
+%! syms x y
+%! f(x) = 2*x;
+%! g(x, y) = 2*x;
+%! assert (~ isequal (f, g))
+
+%!test
+%! syms x y
+%! f(x) = symfun(nan, x);
+%! g(x) = symfun(nan, x);
+%! assert (~ isequal (f, g))
--- a/inst/@symfun/isequal.m
+++ b/inst/@symfun/isequal.m
@@ -50,13 +50,19 @@
 
 function t = isequal(x, y, varargin)
 
+  if (nargin <= 1)
+    print_usage ();
+  end
+
   t = isequal(formula(x), formula(y)) && isequal(argnames(x), argnames(y));
 
   if nargin >= 3 && t
     t = t && isequal(x, varargin{:});
   end
+
 end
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by nvs-abhilash** [comment]
Hope this will do the job. I am a bit confused about the coding style of the code, is it right? @cbm755 please review the code and do see if it might create hidden bugs in the future (I hope not) :)

File: inst/@symfun/isequal.m [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+%%
+%% @end defmethod
+
+function t = isequal(x, y, varargin)
+
+  if isa (x, 'symfun') && isa (y, 'symfun')
+    vars_x = x.vars;
+    vars_y = y.vars;
+
+    if length (vars_x) ~= length (vars_y)
```
cbm755: Maybe `isequal(argnames(x), argnames(y))` will do all these length checks etc?

inst/@symfun/isequal.m
```diff
@@ -0,0 +1,87 @@
...
+      for i = 1:length(vars_x)
+        v_x = vars_x{i}.flat;
+        v_y = vars_y{i}.flat;
+
+        if ~strcmp (v_x, v_y)
+          t = false;
+          return;
+        end
+      end
+      t = isequaln(x, y, varargin{:});
```
cbm755: I guess this is supposed to call the superclass method?  Might be clearer to do `isequal (formula (x), formula (y))`.

inst/@symfun/isequal.m
```diff
@@ -0,0 +1,87 @@
...
+%% f(x) = x + 1;
+%% g(x, y) = x + 1;
+%% isequal(f, g)
+%%   @result{} 0
+%% @end group
+%% @end example
+%%
+%% Note that this function is implemented to avoid issue #740.
+%% It just checks the equality for variables of the @@symfun, if not equal,
+%% returns false otherwise delegates further checking to @@sym/isequaln.
```
cbm755: I think `@seealso{@@sym/isequal}` would be enough, don't need this message.

inst/@symfun/isequal.m
```diff
@@ -0,0 +1,87 @@
...
+%% @example
+%% @group
+%% syms x
+%% f(x) = x + 1;
+%% g(x) = x + 1;
+%% isequal(f, g)
+%%   @result{} 1
+%% @end group
+%% @end example
+%%
```
cbm755: Before your second example, say

`%% Note: two symfuns are equal if they have the same formula @emph{and} same arguments:`

inst/@symfun/isequal.m
cbm755: Or maybe `@seealso{@@sym/isequal, @@symfun/formula, @@symfun/argnames}`?

cbm755:
I made a few suggestions that should result in shorter code, but the overall idea looks fine except for:
1.  comparing `.flat` isn't quite right b/c of assumptions.
2.  this isn't really careful about `isequaln` versus `isequal`.

Regarding `nan`, here is a test:
````
syms x y
f(x) = nan
g(x) = nan
h(y) = nan
isequal (f, g)     % should be false
isequaln (f, g)    % should be true
isequaln (f, h)    % should be false
````
Please add that as a test (and double-check I haven't gotten them mixed up!)

**Comment by cbm755** [comment]
Probably we should test some `isequal (f, g, h)` case to make sure the varargin bit works.

**Comment by cbm755** [comment]
Finally, let's try to get reasonable commit messages: "sympy developers guide" has some advice on this I think.

Your second commit message in this PR is not accurate for example, should be something like:
````
isequal: make some noisy tests quieter
````

**Comment by nvs-abhilash** [comment]
@cbm755 Thanks for your suggestions. I incorporated most of it. I have some doubts.

1. Should I add parameter checks like `if nargin <= 1` because I noticed something strange (to me, might be obvious to you):

I did not put any error checks
```
syms x
f(x) = x + 1
isequal f
```
The above code throws error
```
error: Invalid call to isequal.  Correct usage is:
 -- isequal (X1, X2, ...)
```

But 
```
syms x
f(x) = x + 1

isequal(f)
```
It gave: y undefined, which should happen in my case.

Also, I don't see any error checks in `@sym/isequal` , what am I missing?

**Comment by nvs-abhilash** [comment]
@cbm755 check if all the requirements are met, or there is something more to do.

**Comment by cbm755** [comment]
'isequal f' calls isequal('f'), so this is not our problem.

Some nargin check would be good.

Please try to write nicer commit bags.

**Comment by cbm755** [comment]
Oh I see you debased it, ok but next time don't do that until the review is finished

File: inst/@symfun/isequal.m [hunk 5]
```diff
@@ -0,0 +1,96 @@
...
+%% @seealso{@@sym/isequal, @@symfun/formula, @@symfun/argnames}
+%% @end defmethod
+
+function t = isequal(x, y, varargin)
+
+  t = logical(1);
+  if isa (x, 'symfun') && isa (y, 'symfun')
+    t = isequal (argnames(x), argnames(y));
```
cbm755: Argnames should work for sym too, return empty, so do we need this isa symfun check?

inst/@symfun/isequal.m
```diff
@@ -0,0 +1,96 @@
...
+
+    if nargin >= 3 && t == 1
+      t = t & isequal(x, varargin{:});
+    end
+  end
+
+  t = t & isequal(formula(x), formula(y));
```
cbm755: Should this come before the varargin chaining?  Maybe its same but I worry about corner cases...

nvs-abhilash:
Oh yes I rebased it. Sorry, will take care of that next time.

nvs-abhilash:
I think @symfun/symfun.m failed because it has a test 
```
%!test
%! % defining 1D symfun in terms of a 2D symfun
%! syms x y t
%! syms 'g(x,y)'
%! f(t) = g(t,t);
%! f(5);
%! assert (length (argnames (f)) == 1)
%! assert (isequal (argnames (f), t))
%! assert (isequal( diff(f,x), sym(0)))
```
Here `diff(f, x)` returns a symfun.
So `argnames (diff(f, x))` gives `t`, whereas `argnames (sym(0))` gives an empty matrix.

I can't say weather the test is wrong (comparing a `symfun` with `sym`) or my code needs to be modified?
@cbm755 Need your help to decide.

cbm755:
I suspect the test is wrong; should compare `formula (diff (f, x))` against `sym(0)`.  Maybe it should also `assert (isequal (argnames (diff (f,x)), sym(t))`

Nice catch.

nvs-abhilash:
@cbm755 So, what to do now? Should that be changed in this PR or some other one? 

Again in  `@symfun/int.m` g(x) is compared with a (sym) x^2*y
```
%!test
%! % same as above, but concrete symfun
%! syms x y
%! f(x) = x^2;
%! g = int(f, y);
%! assert (isa (g, 'symfun'))
%! assert (isequal (argnames (g), x))
%! assert (isequal (g, x^2*y))
```

nvs-abhilash:
Also, I was wondering,
```
syms x y
f(x) = nan
g(x) = nan
h(y) = nan
isequal (f, g)     % should be false
isequaln (f, g)    % should be true
isequaln (f, h)    % should be false
```
The last one `isequaln (f, h)` is still true because isequaln still does not care about argnames of f and h
Should isequaln be also overloaded?

cbm755:
> again in `@symfun/int.m`
Yes probably need to fix that too.  Maybe this assumption is made in a lot of places.

cbm755:
We need that nan test to work yes.  Overloading `@symfun/equaln` should work.

(It might also be possible to only overload one of them and have the superclass deal with the other, but I'm not sure.)

cbm755:
> Should that be changed in this PR

It will need to be here, b/c we will not merge this until the tests pass.

nvs-abhilash:
@cbm755 I added the patch for both function's tests. Now all tests are passing. Please review it.

cbm755:
There's a few things discussed above that aren't addressed yet:
1.  I think we discussed adding a nargin check with `print_usage()`?
2.  how about `equaln`?  Does it work?  Is it tested?

nvs-abhilash:
@cbm755 I will do the `print_usage()` , and for `isequaln` I have tested and it not works, should I implement isequaln for `@symfun` or there is some other workaround for this? I cannot understand what you mean by  "It might also be possible to only overload one of them and have the superclass deal with the other."

cbm755:
> I cannot understand what you mean

I haven't looked at it carefully so its hard for me to explain, but there are two superclass fcns `@sym/equaln`, `@sym/equal`.  I meant, it may not be necessary to implement both `@sym/equaln` and `@sym/equal`.  But as I said, I have not looked at it yet.

cbm755:
if you're still unsure, just add the failing `nan` test as an `xtest`.  Or implement `@symfun/isequaln` as you think it should be done: can always be improved later.

nvs-abhilash:
@cbm755 I have added isequaln, see if it is the correct way(for now).

cbm755:
This looks good to me!  Will merge soon.

PR #751, opened 2017-01-31
Fixes for native/pytave ipc

cbm755:
I haven't been testing with the pytave ipc for a long time; will run tests now and try to fix anything that comes up...


cbm755:
Wow, everything passed after this commit!

PR #758, opened 2017-02-06
tests: don't use plain "clear", for __run_test_suite__

cbm755:
Octave has a __run_test_suite__ command which doesn't like "clear"
being done during tests (because it relies on some global variables
I think).  Usually we can just clear the specific variable we need
cleared.  In other cases "clear variables" should do and seems
portable between Octave and Matlab.

### DIFF
--- a/inst/@sym/sym.m
+++ b/inst/@sym/sym.m
@@ -494,7 +494,7 @@ end
 
 %!test
 %! %% assumptions and clearing them
-%! clear  % for matlab test script
+%! clear variables  % for matlab test script
 %! x = sym('x', 'real');
 %! f = {x {2*x}};
 %! asm = assumptions();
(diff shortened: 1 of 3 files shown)


latot:
mm, with this changes we will can't check the clear effect in all the user env....

cbm755:
`clear variables` clears all local variables so I think that is sufficient.  Or did you mean something else?

latot:
right!, that is, only if is possible add or changes the test with that please.

cbm755:
Sorry I don't understand what you mean.

1.  `clear` is bad for upstream Octave's test tooling (because `clear` seems to clear global variables.)
2.  Most tests don't need `clear`.
3.  A few tests do need `clear` (in fact, only on Matlab but that isn't relevant).
4.  This PR changes those tests to use `clear variables` instead of `clear`.

latot:
ups, i was confused, clear is a octave function, sorry.

cbm755:
ok, thanks for checking.

PR #760, opened 2017-02-06
subsasgn: fix to define a matrix while also growing it

cbm755:
Support "clear a; a(3) = sym(1)" which should pad with zeros.
Inspired by upstream https://savannah.gnu.org/bugs/index.php?50202


cbm755:
Will merge in 24 hours or so

cbm755:
or 20 days or so :(

PR #761, opened 2017-02-07
Add new functions

cbm755:


cbm755:
will merge in 24 hours if no concerns

inst/@double/harmonic.m
```diff
@@ -0,0 +1,72 @@
...
+  end
+  cmd = { 'L = _ins[0]'
+          'A = [complex(harmonic(complex(x))) for x in L]'
+          'return A,' };
+  c = python_cmd (cmd, num2cell(x(:)));
+  assert (numel (c) == numel (x))
+  y = x;
+  for i = 1:numel (c)
+    y(i) = c{i};
+  end
```
latot: maybe here use cell2sym instead?

inst/@sym/kroneckerDelta.m
```diff
@@ -0,0 +1,82 @@
...
+
+function a = kroneckerDelta (n, m)
+
+  if (nargin == 1)
+    m = sym(0);
+  elseif (nargin == 2)
+    %% no-op
+  else
+    print_usage ();
+  end
```
latot: and here use:
```

 +  if (nargin == 1)
 +    m = sym(0);
 +  elseif (nargin > 2)
 +    print_usage ();
 +  end
```

??

inst/@double/harmonic.m
cbm755: Yes, that does look poor.  Outputs here are doubles (not sym) but I think `cell2mat` will help.

cbm755:
thanks!  Will merge when tests pass.

Issue #772, opened 2017-04-03
IPC differences: int return types converted to Octave


mtmiller:
The following difference remains between the default `popen2` ipc and the `native` ipc:

    >> sympref ipc popen2
    >> x = python_cmd ('return 42');
    >> sympref ipc native
    >> y = python_cmd ('return 42');
    >> x
    x =  42
    >> y
    y = 42
    >> class (x)
    ans = double
    >> class (y)
    ans = int64

Looks like the `popen2` ipc method converts any Python `int` type to an Octave double, regardless of precision overflow, while the `native` ipc method converts Python `int` objects to Octave int64.

This results in a test failure with `test python_cmd`:

    !!!!! test failed
    ASSERT errors for:  assert (q,1)
    
      Location  |  Observed  |  Expected  |  Reason
         ()           O            E         Class int64 != double

I suggest these should be made to agree, we just have to decide what the conversion should be to be consistent.

mtmiller:
As a followup, there are also differences between Python 2 and 3 with the `popen2` ipc driver

_Python 2:_

    >> setenv PYTHON python2
    >> sympref ipc popen2
    Resetting the communication mechanism
    Forcing the popen2() communication mechanism
    >> x = python_cmd ('return 2'); class (x)
    ans = double
    >> x = python_cmd ('return 2**32'); class (x)
    ans = double
    >> x = python_cmd ('return 2**64'); class (x)
    Traceback (most recent call last):
      File "<stdin>", line 2, in <module>
      File "<stdin>", line 12, in octoutput_drv
      File "<stdin>", line 99, in octoutput
    ValueError: octoutput does not know how to export type <type 'long'>
    error: Python exception: ValueError: octoutput does not know how to export type <type 'long'>
        occurred while copying variables from Python.
        Consider filing an issue at https://github.com/cbm755/octsympy/issues
    error: called from
        python_cmd at line 184 column 5

_Python 3:_

    >> setenv PYTHON python3
    >> sympref ipc popen2
    Resetting the communication mechanism
    Forcing the popen2() communication mechanism
    >> x = python_cmd ('return 2'); class (x)
    ans = double
    >> x = python_cmd ('return 2**32'); class (x)
    ans = double
    >> x = python_cmd ('return 2**64'); class (x)
    ans = double
    >> x = python_cmd ('return 2**128'); class (x)
    ans = double

cbm755:
Probably `popen2` should be modified to return `int64` like pytave does.

[...]

PR #774, opened 2017-04-07
Convert int/long to int64

cbm755:
This addresses #772.  @mtmiller, what do you think?

There might be some way to get the int64 directly instead of converting to double (maybe `sscanf`?) but I didn't make it work.  The difference between `flintmax` and `intmax('int64')` is only a few orders of magnitude so maybe its not worth worrying about.

Also worth noting: large mathematical integers *should* be `sympy.Integer` types so this limit should only effect things like indices.

### DIFF
--- a/inst/private/check_and_convert.m
+++ b/inst/private/check_and_convert.m
@@ -1,5 +1,5 @@
 %% Copyright (C) 2016 Abhinav Tripathi
-%% Copyright (C) 2016 Colin B. Macdonald
+%% Copyright (C) 2016, 2017 Colin B. Macdonald
 %%
 %% This file is part of OctSymPy.
 %%
@@ -23,6 +23,7 @@ function obj = check_and_convert(var_pyobj)
   persistent list_or_tuple
   persistent _sym
   persistent string_types
+  persistent integer_types
   if isempty(builtins)
     builtins = pyeval("__builtins__");
     list_or_tuple = py.tuple({builtins.list, builtins.tuple});
@@ -30,6 +31,7 @@ function obj = check_and_convert(var_pyobj)
     sp = py.sympy;
     _sym = py.tuple({sp.Basic, sp.MatrixBase});
     string_types = py.six.string_types;
+    integer_types = py.six.integer_types;
   end
 
 
@@ -56,9 +58,13 @@ function obj = check_and_convert(var_pyobj)
       % make sure values are converted to sym
       s = structfun (@(t) check_and_convert (t){:}, s, 'UniformOutput', false);
       obj{i} = s;
-    elseif (py.isinstance(x, pyeval('int')))
+    elseif (py.isinstance(x, integer_types))
       if (py.isinstance(x, pyeval('bool')))
         error ('unexpected python bool')
+      end
+      if (abs (r) > intmax ('int64'))
+        error ('precision would be lost converting integer larger than %ld', ...
+               intmax ('int64'))
       end
       obj{i} = int64 (x);
     else
--- a/inst/private/check_and_convert.m
+++ b/inst/private/check_and_convert.m
@@ -62,7 +62,7 @@ function obj = check_and_convert(var_pyobj)
       if (py.isinstance(x, pyeval('bool')))
         error ('unexpected python bool')
       end
-      if (abs (r) > intmax ('int64'))
+      if (abs (double (x)) > intmax ('int64'))
         error ('precision would be lost converting integer larger than %ld', ...
                intmax ('int64'))
       end
--- a/inst/private/extractblock.m
+++ b/inst/private/extractblock.m
@@ -131,9 +131,6 @@ function r = process_item(item)
       assert(M == 1)
       % Note: is it possible to avoid converting to double first?
       r = str2double(C{2});
-      if (~ exist ('OCTAVE_VERSION', 'builtin'))
-        flintmax = 9007199254740992;
-      end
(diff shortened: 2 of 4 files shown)


inst/private/extractblock.m
```diff
@@ -129,7 +129,16 @@
   switch wh
     case OCTCODE_INT
       assert(M == 1)
+      % Note: is it possible to avoid converting to double first?
```
mtmiller: Unfortunately no, there are no ways to parse a string directly into a 64-bit integer at the moment, loss of precision seems inevitable at this point

inst/python_cmd.m
```diff
@@ -216,7 +216,15 @@
 
 %!test
 %! % int
-%! assert (python_cmd ('return 123456,') == 123456)
+%! r = python_cmd ('return 123456');
+%! assert (r == 123456)
+%! assert (isinteger (r))
```
mtmiller: You can combine these two into a single `assert (r, int64 (123456))`

inst/python_cmd.m
```diff
@@ -465,9 +473,9 @@
 %! if (exist ('OCTAVE_VERSION', 'builtin'))
 %! s = ['abc'; 'defgh'; '12345'];
 %! q = python_cmd ('return len(_ins)', s);
-%! assert (q, 1)
+%! assert (q == 1)
```
mtmiller: Or `assert (double(q), 1)` or `assert (q, int64(1))`

inst/private/python_header.py
```diff
@@ -177,7 +177,7 @@ def octoutput(x, et):
             c = ET.SubElement(et, "list")
             for y in x:
                 octoutput(y, c)
-        elif isinstance(x, int):
+        elif isinstance(x, int) or (sys.version_info < (3, 0) and isinstance(x, long)):
```
mtmiller: You use `six` above, are you intentionally avoiding it here?

inst/private/check_and_convert.m
```diff
@@ -56,10 +58,14 @@
       % make sure values are converted to sym
       s = structfun (@(t) check_and_convert (t){:}, s, 'UniformOutput', false);
       obj{i} = s;
-    elseif (py.isinstance(x, pyeval('int')))
+    elseif (py.isinstance(x, integer_types))
       if (py.isinstance(x, pyeval('bool')))
         error ('unexpected python bool')
       end
+      if (abs (r) > intmax ('int64'))
```
mtmiller: `r` is undefined here, maybe `if (abs (double (x)) > intmax ('int64'))`?

This is the kind of range test I feared might be too costly.

cbm755: gah, also `abs` is not defined by pyobject... Thanks!

cbm755: Re: timing:
````
octave:20> tic; py.isinstance(x, pyeval('bool')); toc
Elapsed time is 0.00637603 seconds.
octave:21> tic; py.isinstance(x, pyeval('bool')); toc
Elapsed time is 0.00800395 seconds.
octave:22> tic; py.isinstance(x, pyeval('bool')); toc
Elapsed time is 0.00632501 seconds.
octave:23> 
octave:23> 
octave:23> tic; abs(double(x)) > intmax('int64'); toc
Elapsed time is 0.00253701 seconds.
octave:24> tic; abs(double(x)) > intmax('int64'); toc
Elapsed time is 0.00378704 seconds.
octave:25> tic; abs(double(x)) > intmax('int64'); toc
Elapsed time is 0.00371313 seconds.
````
So its no more expensive than the bool sanity check right above it.

File: inst/private/extractblock.m [hunk 6]
```diff
@@ -129,7 +129,16 @@
...
       r = str2double(C{2});
+      if (~ exist ('OCTAVE_VERSION', 'builtin'))
+        flintmax = 9007199254740992;
```
mtmiller: Why? `flintmax` should be in Matlab

inst/python_cmd.m
cbm755: Right but this way I only test it's an integer class in case some backend wants to mix int32 and int64 for some reason.

inst/private/check_and_convert.m
mtmiller: Also I think I overestimated how much this code path is actually used. With `sympref ipc native`, I only hit this error once in the entire test suite.

inst/private/python_header.py
cbm755: somewhat, just following the style of the local code, filed #775 to clean it up.

inst/private/check_and_convert.m
mtmiller: gah, that was with Python 2, nevermind

inst/python_cmd.m
cbm755: yes, but this test doesn't care about any of that, just want to make sure q has unit length.

mtmiller:
LGTM now, all tests pass for me with pytave under both 2 and 3. I ran some timing tests of the overall test suite run and any time added by the range checks is negligible AFAICT. Pytave with Python 2 is just a hair faster than popen2, but pytave with Python 3 is much slower, possibly because of the `int` class.

cbm755:
We really should be able to do much better than popen2: everything is serialized to xml text, then read back and extracted.  But maybe only for specific cases like large matrices.

Anyway, speed improvements for later, merging this for now.

PR #776, opened 2017-04-07
ci: build against Octave daily

mtmiller:
Test to see whether this approach works.

@cbm755 let me know if you think installing Octave from some tarball hosted somewhere is out of the question (I concede it's suspect at best, even though I maintain it).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,17 +5,27 @@ language: generic
 
 matrix:
   include:
-    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no SYMPY_VER=1.0 OCT_PPA=yes DOCTEST=yes COLUMNS=80
-    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT_PPA=yes DOCTEST=yes COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no SYMPY_VER=1.0 OCT=ppa   DOCTEST=yes COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT=ppa   DOCTEST=yes COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=yes COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=yes COLUMNS=80
+  allow_failures:
+    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=yes COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=yes COLUMNS=80
 
 # need octave devel pkgs for doctest (has compiled code as of July 2015)
 before_install:
-  - if [ "x$OCT_PPA" = "xyes" ]; then
+  - if [ "x$OCT" = "xppa" ]; then
         sudo apt-add-repository -y ppa:octave/stable;
     fi
   - sudo apt-get update -qq
   - sudo apt-get install -y liboctave-dev
   - sudo apt-get install -y octave
+  - if [ "x$OCT" = "xdaily" ]; then
+        sudo apt-get install -y libarpack2-dev libgl2ps-dev libgraphicsmagick++-dev libqrupdate-dev libsuitesparse-dev;
+        wget https://s3.amazonaws.com/octave-snapshot/public/octave-ubuntu-trusty-snapshot.tar.xz;
+        sudo tar --extract --directory=/usr/local --strip-components=1 --file=octave-ubuntu-trusty-snapshot.tar.xz;
+    fi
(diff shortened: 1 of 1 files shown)


mtmiller:
Took a few tries, but the point here is to add a CI test against a rolling snapshot of Octave's development branch.

This would also facilitate CI testing against pytave.

Of course the downside is Octave is no longer a stable point of comparison, CI tests may fail because of something Octave broke or changed.

cbm755:
Seems like a great idea to me.  Some quick thoughts:

1.  I haven't looked carefully, but how can this be passing?  Cell arrays of syms render differently with Octave master...  Actually I see, its b/c you don't run doctests.

2.  Sympy manages to set some Travis runs as  "allowed to fail", if that's easy to do, maybe we should do that here?   Or leave it as-is for a bit and see if causes trouble.

.travis.yml
```diff
@@ -7,6 +7,8 @@ matrix:
   include:
     - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no SYMPY_VER=1.0 OCT_PPA=yes DOCTEST=yes COLUMNS=80
     - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT_PPA=yes DOCTEST=yes COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no SYMPY_VER=1.0 OCT_DEV=yes DOCTEST=no  COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT_DEV=yes DOCTEST=no  COLUMNS=80
```
cbm755: Probably bikeshedding but would `OCT=ppa`, `OCT=devsnap`, `OCT=` be more readable instead of `OCT_PPA` and `OCT_DEV`?

mtmiller: Sure, I've got no problem combining these

mtmiller:
> set some Travis runs as "allowed to fail"

I'll look into that, thanks

.travis.yml
```diff
@@ -16,6 +18,13 @@ before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -y liboctave-dev
   - sudo apt-get install -y octave
+  - if [ "x$OCT_DEV" = "xyes" ]; then
+        sudo apt-get install -y libarpack2-dev libgl2ps-dev libgraphicsmagick++-dev libqrupdate-dev libsuitesparse-dev;
+        wget https://s3.amazonaws.com/octave-snapshot/public/octave-ubuntu-trusty-snapshot.tar.xz;
+        sudo tar --extract --directory=/usr/local --strip-components=1 --file=octave-ubuntu-trusty-snapshot.tar.xz;
+        sudo chmod -R a+rX /usr/local/;
```
mtmiller: This was a hack, I've fixed it in my scripts for building the image, shouldn't be needed after the next build

.travis.yml
```diff
@@ -16,6 +18,13 @@ before_install:
...
+        sudo ldconfig;
```
mtmiller: This shouldn't be necessary either, no files are installed directly in /usr/local/lib, no need to update the cache

.travis.yml
```diff
@@ -3,19 +3,30 @@ dist: trusty
 
 language: generic
 
+env:
```
mtmiller: Travis docs claim this line is required for `allow_failures` to work, but it also adds an extra job with no env vars that fails, hmm

cbm755: Sorry to state the obvious but try without that line?  Or can matrix go inside the `env:`?

Otherwise, I guess we set `DOCTEST=no` for these and leave a comment about `allow_failures` for now?

mtmiller:
Ok looks like the obvious worked.

.travis.yml
```diff
@@ -5,17 +5,27 @@ language: generic
...
 
 # need octave devel pkgs for doctest (has compiled code as of July 2015)
 before_install:
-  - if [ "x$OCT_PPA" = "xyes" ]; then
+  - if [ "x$OCT" = "xppa" ]; then
         sudo apt-add-repository -y ppa:octave/stable;
     fi
   - sudo apt-get update -qq
   - sudo apt-get install -y liboctave-dev
   - sudo apt-get install -y octave
```
cbm755: I support if `$OCT` is daily then we don't need to install octave and liboctave-dev here?  (possibly future improvement anyway)

cbm755:
Ok, I guess these commits should be squashed?  Anything else to do before I merge this?

mtmiller:
Yes, I definitely want to squash this. I want to look at adding proper pytave support with `OCT=daily` next, but that can be a new PR.

.travis.yml
mtmiller: I think it's nicer to keep them for now, because otherwise we would have to list all the individual shared libraries and dependencies that Octave needs to work (as I have to do [here](https://github.com/mtmiller/octave-snapshot/blob/master/Dockerfile)).

cbm755:
Sure, let's do it as a new PR.  This is nice to have indep of pytave.

mtmiller:
Hmm, I have to pick branch names better in the future, the PR is rebased and renamed, but there's no easy way to fix the branch name.

cbm755:
As a joke, I cherry-picked the commit---in case you're losing sleep over the branch name ;-)

Although that's probably worse for the history as now its not linked to this PR and discussion: sorry!

cbm755:
Relevant commit is 14d2c98f392a832def33e5615b67d1fc00c10975

mtmiller:
Thanks! 👍

PR #778, opened 2017-04-11
ipc: error on strings that are not row vectors

cbm755:
Previously we silently took the first row when a char array had multiple
rows.  See Issue #664.  Now we are more compatible with the `py` command.
See also: https://bitbucket.org/mtmiller/pytave/issues/75
Note `strcmp` against `''` rather than `isempty` to avoid `2 x 0` arrays.


mtmiller:
Thanks, this looks good to me. I will do the same on my end.

PR #782, opened 2017-05-08
Use sympy.compatibility to test for certain types

### LINKED ISSUE #775: use sympy.compatibility for integer_types and string_types
Here's an easy one.  In various places we use special casing for `str` and `unicode` and `int` and `long`.  In some places we use `six`.  All of that could be unified with `sympy.compatibility.integer_types` and `...string_types`.

mtmiller:
Use the sympy.compatibility integer_types and string_types wrappers as
type checks.  No longer require the "six" module.

Fixes #775.


mtmiller:
I ran into this while testing OctSymPy with pytave and a minimal Python environment, forgetting to install six. So it seemed like a good time to take care of this.

cbm755:
thanks!

PR #794, opened 2017-05-13
sympref: recover from setting an invalid ipc

### LINKED ISSUE #783: after setting `sympref ipc` to some bogus value, it cannot be reset
````
octave:1> sympref ipc meh
Resetting the communication mechanism
warning: Unsupported IPC mechanism 'meh': hope you know what you're doing
warning: called from
    sympref at line 295 column 13

octave:2> sympref ipc popen2
Resetting the communication mechanism
error: invalid ipc mechanism
error: called from
    python_ipc_driver at line 76 column 7
    sympref at line 309 column 9
    sympref at line 275 column 9
octave:2> 
octave:2> sympref ipc default
Resetting the communication mechanism
error: invalid ipc mechanism
error: called from
    python_ipc_driver at line 76 column 7
    sympref at line 309 column 9
    sympref at line 275 column 9
octave:2> 
````

### PR DESCRIPTION
**Pull request by mtmiller**
Allow sympref to restore everything to a working state with a valid ipc
after configuring an invalid ipc setting.

Fixes #783.

### DIFF
--- a/inst/private/python_ipc_driver.m
+++ b/inst/private/python_ipc_driver.m
@@ -73,5 +73,10 @@ function [A, db] = python_ipc_driver(what, cmd, varargin)
       [A, db] = python_ipc_sysoneline(what, cmd, false, varargin{:});
 
     otherwise
-      error('invalid ipc mechanism')
+      if (strcmp(what, 'reset'))
+        A = true;
+        db = [];
+      else
+        error('invalid ipc mechanism')
+      end
   end
--- a/inst/sympref.m
+++ b/inst/sympref.m
@@ -420,6 +420,13 @@ end
 %! delete('tmp_python_cmd.py')
 
 %!test
+%! warning ('off', 'OctSymPy:sympref:invalidarg', 'local');
+%! sympref ('ipc', 'bogus');
+%! assert (strcmp (sympref ('ipc'), 'bogus'))
+
+%!error <invalid ipc mechanism> syms ('x')
+
+%!test
 %! sympref('defaults')
 %! assert(strcmp(sympref('ipc'), 'default'))
 %! sympref('quiet', 'on')

### CODE REVIEW

File: inst/sympref.m [hunk 1]
```diff
@@ -420,6 +420,13 @@
 %! delete('tmp_python_cmd.py')
 
 %!test
+%! warning ('off', 'OctSymPy:sympref:invalidarg', 'local');
```
cbm755: Looks like we the pattern `s = warning ('off', ...);` then `warning (s)` 

IIRC, that's because my hack to run the tests on Matlab doesn't put them in their own scope.

Maybe I'll fix that post-merge which has the nice side-effect of documenting this in the git log.

inst/private/python_ipc_driver.m
```diff
@@ -73,5 +73,10 @@
       [A, db] = python_ipc_sysoneline(what, cmd, false, varargin{:});
 
     otherwise
-      error('invalid ipc mechanism')
+      if (strcmp(what, 'reset'))
+        A = true;
+        db = [];
+      else
+        error('invalid ipc mechanism')
```
cbm755: Its ok to move these things to Octave-formating-style when updating existing code.  (well, it should still be matlab-dialect).

cbm755:
I merged this and addressed the above comments.

cbm755:
FTR 0680876fbbaf57293664fd185676452715cb472f

inst/sympref.m
mtmiller: Ah, ok. I missed that you had a test runner for Matlab, that's a good thing.

We could make it more like Octave, with a separate function scope for each distinct test "paragraph", if you're open to that.

inst/private/python_ipc_driver.m
mtmiller: Yeah, good catch. I copy-pasted the `strcmp` line from one of the other drivers, and just indented the else without looking at the style.

inst/sympref.m
cbm755: Its pretty horrible: https://github.com/cbm755/octsympy/blob/master/misc/extract_tests_for_matlab.m

Probably easy enough to change it to output one file per test instead of one `test_...` file per .m file.  None of the Matlab stuff is very high priority for me or AFAIK anyone else.

Issue #793, opened 2017-05-12
[maint] run CI tests on pytave


cbm755:
@mtmiller do recent changes to pytave make this do-able?  No hurry, just wanted to have an issue number for it.

mtmiller:
Yes I think so. All tests now pass for me locally. It would have to be against `OCT=daily` (at least until we have a 4.2 apt installable).

mtmiller:
Hmm, see also #461

mtmiller:
And to clarify, yes pytave does build and run against the gcc and pythons that are available in Ubuntu 14.04 that Travis CI currently targets. I'm using 14.04 in pytave CI as well.

cbm755:
Fixed by f94f0f10933bc4c9aecf8e0ef8b07375f347546e (#801).

Issue #461, opened 2016-05-31
get PyTave working on Travis CI


cbm755:
The default ipc is set in `sympref`.  But Travis is going to freak out if we do that.

I wonder how hard it would be to build PyTave on Travis (its Ubuntu 14.04 I think).  No point while we have 132 failing tests but eventually we'll need this.

Configuration is in `.travis.yml`: its quite fiddly to debug :(  I suggest holding off for PyTave to mature a bit.

cbm755:
A nice way for this to work would be if the Travis test matrix included running the full test set with popen2 as well as a separate run for pytave.  We could define a `OCTSYMPY_DEFAULT_IPC` env variable: similar to how doctests are dealt with in some runs and not others.

We might not make that env var an official part of the Symbolic API, but just use it for tests.  Or we could just `sed` something in `sympref` within the Travis yml file!

mtmiller:
Wouldn't just putting an inline `sympref ipc $FOO` in the --eval string before running the test suite work? Rather than trying to force sympref to do something different depending on some hidden environment variable or editing the file to do something different.

cbm755:
Fixed by f94f0f10933bc4c9aecf8e0ef8b07375f347546e (#801).

PR #801, opened 2017-05-15
Use pytave on travis with OCT=daily builds

genuinelucifer:
Since, we are trying builds with latest octave, why not run those with pytave? As far as I remember pytave was always meant to be the default IPC, so I think we should use it when possible with new octave. It cannot be build with ppa but can be built with daily.  
Also, I found that `python3-dev` is present on travis but `python-dev` is not. So I added a line for installing `python$PYTHON_VERSION-dev`.  
Also, to properly build pytave with python3 there needs to be `PYTHON_VERSION=3`, I could have made a new variable but since we already had a `PYTHON_VER` so I renamed it.  
And, when path is being fed into octave, I used the variable $TRAVIS_BUILD_DIR so that I dont have to change the username in the path when I am running the build on travis connected to my account.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,17 @@ language: generic
 
 matrix:
   include:
-    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no SYMPY_VER=1.0 OCT=ppa   DOCTEST=yes COLUMNS=80
-    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT=ppa   DOCTEST=yes COLUMNS=80
-    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=no COLUMNS=80
-    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=no COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no  SYMPY_VER=1.0 OCT=ppa   DOCTEST=yes COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no  SYMPY_VER=1.0 OCT=ppa   DOCTEST=yes COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no  SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no  SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=yes SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=yes SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
   allow_failures:
-    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=no COLUMNS=80
-    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=no COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no  SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no  SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=yes SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=yes SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
 
 # need octave devel pkgs for doctest (has compiled code as of July 2015)
 before_install:
@@ -27,8 +31,9 @@ before_install:
         sudo tar --extract --directory=/usr/local --strip-components=1 --file=octave-ubuntu-trusty-snapshot.tar.xz;
     fi
   - sudo apt-get install -y python$PYTHON_VER-pip
+# we need python-dev and python3-dev but we can drop numpy after https://bitbucket.org/mtmiller/pytave/issues/84 is resolved.
   - if [ "x$PYTAVE" = "xyes" ]; then
-        sudo apt-get install -y libboost-python-dev python$PYTHON_VER-numpy;
+        sudo apt-get install -y python$PYTHON_VER-dev python$PYTHON_VER-numpy;
     fi
 
 install:
@@ -39,12 +44,11 @@ install:
         octave -W --no-gui --eval "pkg install doctest-0.5.0.tar.gz";
     fi
   - if [ "x$PYTAVE" = "xyes" ]; then
-        hg clone https://bitbucket.org/genuinelucifer/pytave_main pytave;
+        hg clone https://bitbucket.org/mtmiller/pytave pytave;
         pushd pytave;
-        hg checkout travis;
         pwd;
         autoreconf --install || exit 1;
-        ./configure && make || exit 1;
+        ./configure PYTHON_VERSION=$PYTHON_VER && make || exit 1;
         popd;
     fi
 
@@ -55,10 +59,15 @@ script:
   - stty cols $COLUMNS rows 40
   - tput cols; stty size
   - pushd inst
-  - octave --path=/home/travis/build/cbm755/octsympy/pytave -W --no-gui --eval "r = octsympy_tests; exit(r)"
+  - if [ "x$PYTAVE" = "xyes" ]; then
+        export IPC=native;
+    else
+        export IPC=default;
+    fi
+  - octave --path=$TRAVIS_BUILD_DIR/pytave -W --no-gui --eval "sympref ipc $IPC; r = octsympy_tests; exit(r)";
   - cat octsympy_tests.log
   - if [ "x$DOCTEST" = "xyes" ]; then
-        octave --path=/home/travis/build/cbm755/octsympy/pytave -W --no-gui --eval "pkg load doctest; syms x; r = doctest('.'); exit(~r)";
+        octave --path=$TRAVIS_BUILD_DIR/pytave -W --no-gui --eval "pkg load doctest; sympref ipc $IPC; syms x; r = doctest('.'); exit(~r)";
     fi
   - popd
 


.travis.yml
```diff
@@ -55,10 +54,14 @@ script:
...
-  - octave --path=/home/travis/build/cbm755/octsympy/pytave -W --no-gui --eval "r = octsympy_tests; exit(r)"
+  - if [ "x$PYTAVE" = "xyes" ]; then
+        octave --path=$TRAVIS_BUILD_DIR/pytave -W --no-gui --eval "sympref ipc native; r = octsympy_tests; exit(r)";
+    else
+        octave -W --no-gui --eval "r = octsympy_tests; exit(r)";
+    fi
   - cat octsympy_tests.log
   - if [ "x$DOCTEST" = "xyes" ]; then
-        octave --path=/home/travis/build/cbm755/octsympy/pytave -W --no-gui --eval "pkg load doctest; syms x; r = doctest('.'); exit(~r)";
+        octave --path=$TRAVIS_BUILD_DIR/pytave -W --no-gui --eval "pkg load doctest; syms x; r = doctest('.'); exit(~r)";
```
genuinelucifer: Since doctests are not run on daily builds so passing pytave directory doesn't really do anything but since it was already there so I didn't remove it.

.travis.yml
```diff
@@ -55,10 +54,14 @@ script:
   - stty cols $COLUMNS rows 40
   - tput cols; stty size
   - pushd inst
-  - octave --path=/home/travis/build/cbm755/octsympy/pytave -W --no-gui --eval "r = octsympy_tests; exit(r)"
+  - if [ "x$PYTAVE" = "xyes" ]; then
+        octave --path=$TRAVIS_BUILD_DIR/pytave -W --no-gui --eval "sympref ipc native; r = octsympy_tests; exit(r)";
+    else
+        octave -W --no-gui --eval "r = octsympy_tests; exit(r)";
+    fi
```
genuinelucifer: Since pytave is not selected by default, we need to sympref it, so I added if-else and since pytave is not required in else so no need to pass the path.

.travis.yml
```diff
@@ -26,22 +26,21 @@ before_install:
...
-  - pip$PYTHON_VER install --user sympy==$SYMPY_VER
+  - pip$PYTHON_VERSION install --user sympy==$SYMPY_VER
   - if [ "x$DOCTEST" = "xyes" ]; then
         mkdir ${HOME}/octave;
         wget https://github.com/catch22/octave-doctest/releases/download/v0.5.0/doctest-0.5.0.tar.gz;
         octave -W --no-gui --eval "pkg install doctest-0.5.0.tar.gz";
     fi
   - if [ "x$PYTAVE" = "xyes" ]; then
-        hg clone https://bitbucket.org/genuinelucifer/pytave_main pytave;
+        hg clone https://bitbucket.org/mtmiller/pytave pytave;
```
genuinelucifer: The build can be done with the main repo. No need for a proxy repo and bookmark.

.travis.yml
```diff
@@ -26,22 +26,21 @@ before_install:
         wget https://s3.amazonaws.com/octave-snapshot/public/octave-ubuntu-trusty-snapshot.tar.xz;
         sudo tar --extract --directory=/usr/local --strip-components=1 --file=octave-ubuntu-trusty-snapshot.tar.xz;
     fi
-  - sudo apt-get install -y python$PYTHON_VER-pip
+  - sudo apt-get install -y python$PYTHON_VERSION-pip
   - if [ "x$PYTAVE" = "xyes" ]; then
-        sudo apt-get install -y libboost-python-dev python$PYTHON_VER-numpy;
+        sudo apt-get install -y python$PYTHON_VERSION-dev python$PYTHON_VERSION-numpy;
```
genuinelucifer: No need for boost. But we need python-dev.

cbm755:
Looks great!  I suggest we run both popen2 *and* pytave with daily (that is, add to the matrix instead of replacing)

This closes Issue #793 and Issue #461.

.travis.yml
mtmiller: Please try to keep the parts of the eval that don't change together. Just set a variable to include an extra few commands in the eval if `$PYTAVE` is enabled.

.travis.yml
mtmiller: 👍 does no harm to keep it there for the future

mtmiller:
Same, please include the pytave tests as additions, not changes, to the existing matrix. In the future we may have stable, stable+pytave, daily, and daily+pytave.

.travis.yml
```diff
@@ -5,13 +5,13 @@ language: generic
...
-    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=no COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VERSION=  PYTAVE=no  SYMPY_VER=1.0 OCT=ppa   DOCTEST=yes COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VERSION=3 PYTAVE=no  SYMPY_VER=1.0 OCT=ppa   DOCTEST=yes COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VERSION=  PYTAVE=yes SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VERSION=3 PYTAVE=yes SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
   allow_failures:
-    - env: PYTHON=python2 PYTHON_VER=  PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=no COLUMNS=80
-    - env: PYTHON=python3 PYTHON_VER=3 PYTAVE=no SYMPY_VER=1.0 OCT=daily DOCTEST=no COLUMNS=80
+    - env: PYTHON=python2 PYTHON_VERSION=  PYTAVE=yes SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
+    - env: PYTHON=python3 PYTHON_VERSION=3 PYTAVE=yes SYMPY_VER=1.0 OCT=daily DOCTEST=no  COLUMNS=80
```
mtmiller: The reason for changing these variable names? If it's to automatically pass this value to pytave's configure, I think that is a little bit too magical, it should be done explicitly instead.

.travis.yml
cbm755: I have a slight preference for the if-else form, but not a big deal for me either way.

Doctest runner will need this sympref bit too, eventually

.travis.yml
genuinelucifer: Yes it is for pytave configure. Well we already used a PYTHON_VER so I thought it's better not to include new variables. Moreover defining a python version for the whole environment should signify that it will be used for everything, doesn't it?

genuinelucifer:
I will change it to add pytave builds and not replace the existing ones by tomorrow.

.travis.yml
mtmiller: Defining a Python version for the entire build environment would be great, yes. That's not what this does though 😉 Are you going to set `PYTHON_VERSION=2`? I would be happy to see that here.

.travis.yml
mtmiller: And NumPy is only for now, can be dropped when a pytave bug is fixed.

genuinelucifer:
> Are you going to set `PYTHON_VERSION=2` ?

I would like to but doing that breaks the build of pytave. Apparently there is no binary named `python2` but there is `python` for python2 and `python3` for itself. How would you like to pass the python version to configure then? Should add an `export PYTHON_VERSION=$PYTHON_VER` before calling configure?  
  
> Please try to keep the parts of the eval that don't change together. Just set a variable to include an extra few commands in the eval if $PYTAVE is enabled.
  
Do you mean: add an `if` in eval block to check for the environment variable PYTAVE and call sympref if that is 'yes'?

.travis.yml
```diff
@@ -26,22 +30,21 @@ before_install:
...
         octave -W --no-gui --eval "pkg install doctest-0.5.0.tar.gz";
     fi
   - if [ "x$PYTAVE" = "xyes" ]; then
-        hg clone https://bitbucket.org/genuinelucifer/pytave_main pytave;
+        hg clone https://bitbucket.org/mtmiller/pytave pytave;
         pushd pytave;
-        hg checkout travis;
         pwd;
         autoreconf --install || exit 1;
         ./configure && make || exit 1;
```
cbm755: To answer one of your questions, I would expect this to work:
````
PYTHON_VERSION=$PYTHON_VER ./configure && ...
````

.travis.yml
cbm755: maybe we should leave a comment in the yml file then to remember to remove it later

.travis.yml
mtmiller: `./configure PYTHON_VERSION=$PYTHON_VER` also works

mtmiller:
> How would you like to pass the python version to configure then?

Either setting it in the environment via the `env` settings, setting it in the environment in the `install` script, or passing it as an argument to configure.

My point is just that we should explicitly set both `PYTHON_VERSION=2` and `PYTHON_VERSION=3`, not let `PYTHON_VERSION` be empty for one of them and trust the autodetection to work.

Changing the `env` block to set `PYTHON_VERSION` would be fine, as long as it's set to both 2 and 3, not left as an empty string.

mtmiller:
Take a look at [pytave's CI configuration](https://bitbucket.org/mtmiller/pytave/src/14b134ffdc242a64eb342fe434ed414278f3dd57/bitbucket-pipelines.yml) for how I call configure for both Python 2 and 3. Bitbucket CI doesn't have a configuration matrix, so different configurations are done in series instead of in parallel, but you can see how it could be adapted to a Travis style.

mtmiller:
> Do you mean: add an if in eval block to check for the environment variable PYTAVE and call sympref if that is 'yes'?

I was thinking something like

    if [ "x$PYTAVE" = xyes ]; then ipc=native; else ipc=default; fi
    …
    octave … --eval "sympref ipc $ipc; …"

for my preference. But @cbm755 prefers listing the two eval strings out in full.

cbm755:
I'm fine with @mtmiller's suggested approach too

genuinelucifer:
I have incorporated most of your suggestions. I could not set PYTHON_VERSION to 2 for python2 as that will require an extra if for configure which seems much more work. And we cannot set PYTHON_VER to 2 as that will break installation of packages.  
Hopefully, all the tests will pass too.

genuinelucifer:
Also, I didn't comment about removing numpy in future as I couldn't find any issue on pytave repo that could be referenced.

mtmiller:
> couldn't find any issue on pytave repo

[pytave #84](https://bitbucket.org/mtmiller/pytave/issues/84), sorry it's not clear that this is causing errors when NumPy is not installed 😄 will fix immediately

mtmiller:
> I could not set PYTHON_VERSION to 2 for python2 as that will require an extra if for configure which seems much more work

Ok, I think it's important to avoid autodetection, but I can look at that in a followup PR

genuinelucifer:
I have added the comment and linked to pytave issue.

genuinelucifer:
> Doctest needs sympref IPC too

Done.

cbm755:
I'll merge this b/c I want to test it with my next PR :-)

But of course further refinement most welcome: in particular I agree with @mtmiller's comments about defining `PYTHON_VERSION` in the `env`, and working around this for the package install bits (``python-dev`` versus `python3-dev`), instead of vice versa as we have it now.

Issue #795, opened 2017-05-13
sympref tests failing on windows


genuinelucifer:
8 tests in sympref.m fail on windows.  
Output of `test sympref verbose`:  
```
octave:4> test sympref verbose
>>>>> D:\repos\octsympy\inst\sympref.m
***** shared sympref_orig
 sympref_orig = sympref ();
***** test
 % test quiet, side effect of making following tests a bit less noisy!
 sympref quiet on
 a = sympref('quiet');
 assert(a == 1)
***** error <line 1> python_cmd('raise ValueError');
Traceback (most recent call last):
  File "<string>", line 1, in <module>
OSError: [Errno 22] Invalid argument: 'D:\repos\\octsympy\\inst\\private\\python_header.py'
status =  1
out =
ind = [](0x0)
!!!!! error failed.
Expected <line 1>, but got <sysoneline ipc: system() call failed!>
```
  
Also, the IPC is not restored when the tests fail!

cbm755:
Can you try the following?
````
sympref ipc system
python_cmd('raise ValueError')
````

I thought I tested this on a windows box before or soon after the last release (but I only have sporadic access so I might not have).

Does it happen with 2.5.0 and stable Octave?  I probably only tested the "bundle" package.

**Comment by genuinelucifer** [comment]
on using system ipc, the system call is failing on windows:
```
octave:3> python_cmd('raise ValueError')
Traceback (most recent call last):
  File "<string>", line 1, in <module>
OSError: [Errno 22] Invalid argument: 'D:\repos\\octsympy\\inst\\private\\python_header.py'
OctSymPy v2.5.1-dev: this is free software without warranty, see source.
Using system()-based communication with Python [sysoneline].
Warning: this will be *SLOW*.  Every round-trip involves executing a
new Python process and many operations involve several round-trips.
Warning: "sysoneline" will fail when using very long expressions.
status =  1
out =
ind = [](0x0)
error: sysoneline ipc: system() call failed!
error: called from
    python_ipc_sysoneline at line 141 column 5
    python_ipc_driver at line 62 column 15
    python_cmd at line 164 column 9
```

genuinelucifer:
I tried the same with release 2.5 of symbolic and octave 4.2.0 as well as 4.2.1 (both stable releases) and the error occurs on both octave versions.

mtmiller:
[...]

PR #803, opened 2017-05-17
Fix escaping of filename which prevents system IPC to work on windows.

genuinelucifer:
Fixes issue #795


cbm755:
This was already +1'd by @mtmiller, merging.

PR #804, opened 2017-05-18
sym() and vpa() refactor

### LINKED ISSUE #604: Treatment of unit imaginary
Octave has `I` and `i` and `j`.  Matlab has `i` and `j`.
- [x] Currently: `sym(I)`, `sym(i)` and `sym(j)` gives the symbolic imaginary unit `sym(i)`.  This seems good to me.
- [ ] `sym('I')` gives `sym(i)`.  SMT gives a symbolic variable named `I`.
- [ ] `sym('i')` gives `sym(i)`.  SMT gives a symbolic variable named `i`.
- [x] ~~`sym('j')` gives `sym(i)`.~~
- [x] `sym('j')` gives symbolic var named `j`.  SMT gives a symbolic variable named `j`.

I tend to think the SMT behaviour is what a user would expect.  @latot do you agree?

I guess `sym('inf')` and `sym('oo')` should not have special treatment either.  Looks like strings should generally just make symbols.  But this does move us further from SymPy...  In particularly `sym(char(x))` is (almost?) an identity right now.

cbm755:
I feel like I could tweak this forever, so here's some progress anyway.  I've taken apart from of @latot's work to centralize `const_to_python_str` (which dealt with doubles *and* strings).  That logic is a little more spread out now, but IMHO its clearly to understand `@sym/sym` which is the code that matters.

There were also sublte differences between what should happen with e.g., `pi` in `sym(pi, 'f')`, `sym(pi, 'r')` and `vpa(pi)`.

This also fixes a bug I introduces where `sym(inf, 'f')` was zero.`

Finally, it fixes #604 by allowing `sym('i')` to make a variable named `i` rather than imaginary unit, which is for SMT compatibility.


cbm755:
I guess this is mostly @latot's stuff: what do you think?  Its probably hard to review overall but if you go commit-by-commit, I hope each one is sensible.

inst/@sym/private/double_to_sym_heuristic.m
```diff
@@ -0,0 +1,64 @@
...
+  elseif (isinf (x))
+    y = python_cmd ('return S.Infinity');
+  elseif (isequal (x, pi))
+    %% Special case for pi
+    y = python_cmd ('return S.Pi');
+  elseif (isequal (x, -pi))
+    y = python_cmd ('return -S.Pi');
+  elseif ((abs (x) < flintmax) && (mod (x, 1) == 0))
+    y = python_cmd ('return Integer(_ins[0])', int64 (x));
+  else
```
latot: i would like keep here some loop instead of check every var line by line....

inst/@sym/sym.m
```diff
@@ -277,7 +277,6 @@
   end
 
   asm = {};
-  check = true;
```
latot: I was trying to avoid the most possible call python_cmd every time, the main reason is https://github.com/cbm755/octsympy/pull/561, to can use classdef we will run an extra lines script every time we call that function for the limitations, so we can avoid write a lot of code and reduce it....

If you want remove it maybe instead keep it as comments? Just to don't loose the idea.

inst/vpa.m
```diff
@@ -62,8 +62,15 @@
         'x, n = _ins'
         'return sympy.N(x, n),' };
     r = python_cmd (cmd, x, n);
-  elseif (ischar(x))
-    x = magic_str_str(x);
+  elseif (ischar (x))
+    if (strcmp (x, 'inf') || strcmp (x, 'Inf') || strcmp (x, '+inf') || ...
```
latot: Hi, i maybe instead of this vpa we can simplify to:
```
function r = vpa(x, n)

  if (nargin == 1)
    n = digits();
  elseif (nargin ~= 2)
    print_usage ();
  end

  n = sym(n);

  cmd = {
      'x, n = _ins'
      'return sympy.N(x, n),' };
  r = python_cmd (cmd, x, n);
end
```
Actually pass all the tests, and seems works.

All the check we do here are basically what sym function does.

inst/@sym/private/double_to_sym_heuristic.m
```diff
@@ -0,0 +1,64 @@
...
+    err3 = abs (sqrt (N3) - x);
+    if (err1 <= err3)
+      if (err1 <= err2)
+        y = python_cmd ('return Rational(*_ins)', int64 (N1), int64 (D1));
+      else
+        y = python_cmd ('return Rational(*_ins)*S.Pi', int64 (N2), int64 (D2));
+      end
+    else
+      y = python_cmd ('return sqrt(Integer(*_ins))', int64 (N3));
+    end
```
latot: Thinking in the future, this is a great idea, to complement it i would like.. "split"? the differents methods to get the aproximation, so then we can get this file to be more easy to edit, i'm thinking to can add some "functions" and then the script run all in some loop and choose the best.

inst/@sym/private/double_to_sym_exact.m
```diff
@@ -17,29 +17,22 @@
...
+    y = python_cmd ('return S.NaN');
+  elseif (isinf (x) && x < 0)
+    y = python_cmd ('return -S.Infinity');
+  elseif (isinf (x))
+    y = python_cmd ('return S.Infinity');
   else
-    s = x;
+    %% Rational will exactly convert from a float
+    y = python_cmd ('return Rational(_ins[0])', x);
   end
```
latot: same as below

inst/@sym/private/double_to_sym_heuristic.m
latot: i think is more easy to handle, or maybe with less code?

inst/@sym/sym.m
```diff
@@ -320,62 +320,80 @@
...
-          s = python_cmd ('return Rational(_ins[0])', x);
-        else
-          s = python_cmd ('return Rational(_ins[0]) + I*Rational(_ins[1])', ...
-                          real (x), imag (x));
-        end
-        return
+    iscmplx = ~isreal (x);
+    if (iscmplx && isequal (x, 1i))
+      s = python_cmd ('return S.ImaginaryUnit');
+      return
```
latot: this part is related to the "check" comment.

inst/@sym/private/double_to_sym_exact.m
latot: https://github.com/cbm755/octsympy/pull/804#discussion_r117492313

latot:
Ready!, there is greats changes, there is the comments.

inst/@sym/private/double_to_sym_heuristic.m
cbm755: I figured you might say this ;-)  I respect that point of view, but I disagree.   To me a bunch if `elseif` is readable and clear.  Ultimately just a matter of style.

But if that list gets much longer in the future, then maybe I prefer your approach.

inst/vpa.m
cbm755: Wow, cool, I will think about this!  Did you mean `x = sym(x)`?  Otherwise I don't see how that could work.

inst/@sym/private/double_to_sym_heuristic.m
cbm755: I suppose that could be done in the future.  Although I'm not sure there will be more heuristics: I read through the SMT docs and it sounds like these are the three cases they do.

inst/@sym/sym.m
cbm755: I'm not sure I understand.  Is this about `sym` explicitly calling itself?  I thought we were not supposed to have the sym ctor *explicitly* calling `sym(...)`, but it was ok if sym called some other function which then called `sym(...)`.

What do you mean by "extra lines script every time we call that function for the limitations"?  Can you give me an explicit bit of code that will need to change?

Sorry, but I've kind of lost track of the various classdef issues we face.

Does this comment need to be addressed before I merge this?

cbm755: Would adding this comment address your concern?
````
% Note: a previous version of this code deferred most calls to `python_cmd
% until near the end of the function.  Now, it is called in many places.  Its
% possible this has some negative consequences for a future port to classdef.
% If you're reading this in the future and have successfully done that port,
% then please delete this comment ;-)
````

Otherwise, what should it say?  (I don't understand what the issue is here so harder for me to make a precise statement).

latot: if i remember right this was the reason https://savannah.gnu.org/bugs/?49421, in classdef we can't set properties, so every time when we try to set it we need to run the "Copy" function, other way maybe is store the result in a new var and set it in the end of the script, then we can call the function only one time.

inst/vpa.m
latot: i do something really bad when i test this, is necessary check this.

inst/@sym/sym.m
cbm755: Ok, I think I see, thank you.

There are paths in #590 that end early with `s = meh...; return`, so I don't see this PR's code as being much worse.  I will merge for now, fight with classdef when we get there ;-)

PR #805, opened 2017-05-18
Pass strings instead of generating code

cbm755:
@latot, do these two commits look ok?

### DIFF
--- a/inst/@sym/sym.m
+++ b/inst/@sym/sym.m
@@ -419,7 +419,6 @@ function s = sym(x, varargin)
       if (flag)
         check = false;
       end
-      x = strrep (x, '"', '\"');   % Avoid collision with S("x") and Symbol("x")
 
       isnum = ~isempty (regexp (x, '^[-+]*?\d*\.?\d*(e-?\d+)?$'));  % Is Number
     end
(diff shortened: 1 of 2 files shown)


latot:
tomorrow i'll have time to check the 2 pr, now i'll go to sleep :D 

Cya.

latot:
Hi, this is great, now is possible this because some behavior change, before when we get the python string we get ```Symbol("x")```, so with the syntax we was getting the error, but now we get ```Symbol('x')``` and works fine, only as a note, some possible conflict is if we write some code with ```'``` instead of ```"``` we cat get that error again, so i think we can keep this change and wait that possibility or  do something like this:

```x = strrep (x, "'", "\'");```

In any case, while we use only double quotes inside the python expressions we shouldn't get this problem.

Just in case we add a test to check this:

```
x=sym("x");
assert(x, sym(sympy(x)))
```

(I don't know how we miss this test xD)

I think this would can catch the most possible problems, because this in one of the few files where we really manipulate the python string like that.

Cya.

cbm755:
Thanks @latot, I've added that test to `@sym/sympy.m` (I think it was passing on master but agree that's an important case to test.)

I'll assume your comments are a `+1` so I will merge when tests pass, unless I hear otherwise.

The nice thing about this approach is strings don't need to be escaped because we are passing them as arguments.  (For some IPC mechanism, the string will need to be escaped to pass it as an argument, but that is Someone Else's Problem as far as  `@sym/sym.m` is concerned.)

PR #806, opened 2017-06-01
sym: accept function_handle inputs

cbm755:
This is for compatibility with SMT.


inst/@sym/sym.m
```diff
@@ -275,6 +292,24 @@
     return
   end
 
+  if (isa (x, 'function_handle'))
+    assert (nargin == 1)
+    %% Need argnames of the handle.  TODO: can do better than regex?
+    T = regexp (func2str (x), '^\@\((?:(\w+)(?:, ))*(\w+)?\).*', 'tokens');
```
cbm755: @mtmiller do you know of a better way to get the argnames of a `function_handle`?

mtmiller: LGTM, I don't know of a better way without going to C++

cbm755:
hmmm, mucked this up, will merge from command line

PR #811, opened 2017-06-21
Update @sym/mpower test for SymPy dev

### LINKED ISSUE #787: `test @sym/mpower` fails with SymPy 1.0.1.dev
The test for a scalar raised to a matrix exponent fails with a different than expected error with SymPy 1.0.1.dev.

In Python, the new behavior is

    >>> x = S('x')
    >>> A = Matrix([[1, 2], [3, 4]])
    >>> x**A
      File "<console>", line 1, in <module>
      File "/home/mike/src/sympy/sympy/core/decorators.py", line 78, in __sympifyit_wrapper
        return func(a, b)
      File "/home/mike/src/sympy/sympy/core/decorators.py", line 119, in binary_op_wrapper
        return func(self, other)
      File "/home/mike/src/sympy/sympy/core/expr.py", line 150, in __pow__
        return Pow(self, other)
      File "/home/mike/src/sympy/sympy/core/cache.py", line 93, in wrapper
        retval = cfunc(*args, **kwargs)
      File "/home/mike/src/sympy/sympy/core/compatibility.py", line 792, in wrapper
        return user_function(*args, **kwds)
      File "/home/mike/src/sympy/sympy/core/power.py", line 217, in __new__
        c, ex = factor_terms(e, sign=False).as_coeff_Mul()
      File "/home/mike/src/sympy/sympy/matrices/matrices.py", line 2217, in __getattr__
        "%s has no attribute %s." % (self.__class__.__name__, attr))
    AttributeError: ImmutableDenseMatrix has no attribute as_coeff_Mul.

In previous versions we were expecting the `NotImplementedError`, now raises an `AttributeError`. If this is the correct new behavior, could fix by simply updating the error regex to match both possible errors.

mtmiller:
Fixes #787.

Kind of awkward, but the cleanest way I could come up with to test something that may work or may throw a specific error. Also open to other ways to test the value of B if this is too ugly.

### DIFF
--- a/inst/@sym/mpower.m
+++ b/inst/@sym/mpower.m
@@ -130,11 +130,18 @@ end
 %! C = subs(B, n, 0);
 %! assert (isequal (C, sym(eye(2))))
 
-%!error <NotImplementedError>
-%! % scalar^array not implemented
+%!test
+%! % scalar^array works in SymPy > 1.0.0, otherwise not implemented
 %! syms x
 %! A = [1 2; 3 4];
-%! B = x^A;
+%! try
+%!   B = x^A;
+%!   waserr = false;
+%! catch
+%!   waserr = true;
+%!   notimpl = any (strfind (lasterr (), 'NotImplementedError'));
+%! end
+%! assert ((~ waserr && strcmp (regexprep (disp (B, 'flat'), '\s+', ''), 'x**Matrix([[1,2],[3,4]])')) || (waserr && notimpl))
 
 %!error
 %! A = sym([1 2; 3 4]);


cbm755:
Minor nitpick: don't use octave specific strings (with ")

I would probably have just skipped the test with sympy ver <= 1.0.0 but this is even better.

cbm755:
I'm concerned that this matches `2.^A`... it should not.  That might be an upstream problem.

(It should be defined via the eigendecomposition of A)

mtmiller:
> Minor nitpick: don't use octave specific strings (with ")

Thanks, force of habit.

> I'm concerned that this matches 2.^A... it should not. That might be an upstream problem.

Yeah good point, I spent time figuring out how to test the result that I missed the fact that it's turning the matrix exponential into an elementwise operation. May be a codegen issue, since it's turning the exponentiation into a ".^".

mtmiller:
I think upstream hasn't really got the codegen part figured out for this operation yet.

    expr = S('x') ** Matrix([[1, 2], [3, 4]])

Turning this into C code returns

    f_result = pow(x, Matrix([[1, 2], [3, 4]]))

which is clearly useless.

So back to my original worry, the way I'm testing the validity of the result is very hacky, any other ideas besides trying to turn it into a function and evaluate it? I could try to test `B.flat` with whitespace removed.

mtmiller:
assert (regexprep (B.flat, '\s', ''), 'x**Matrix([[1,2],[3,4]])')

mtmiller:
This works for me. I can also use `char(B)` or `disp(B, 'flat')`, whichever form you prefer. But it needs the regexprep, not just strtrim, because there are embedded newlines in there.

I do plan on squashing this once we have something you think is the most maintainable.

cbm755:
Thanks, I like `disp(B, 'flat')`.

Lots of stuff doesn't quite work with codegen, but I should file a bug for the matrix exponential to `.^` thing.

cbm755:
Upstream issue: https://github.com/sympy/sympy/issues/12789

cbm755:
> because there are embedded newlines in there.

Hmm, maybe that was a design mistake when I wrote octave_code.  I think Sympy  makes a distinction between generating a multiline .m file (`codegen`) and generating code from an expression (`octave_code()`).  Presumably it could be changed to embed newlines for the former but not the latter.

mtmiller:
Final (?) update. Using `disp(…, 'flat')` now, and caught another Matlab compatibility thing.

mtmiller:
Fixed up and rebased, ready to be merged now

PR #815, opened 2017-07-01
Update decomposition methods for SymPy 1.1

### LINKED ISSUE #812: matrix factorizations may change names in dev sympy
Old names are deprecated so we'll probably get lots of warnings.

mtmiller:
Prefer the new SymPy method names 'cholesky_decomposition',
'LU_decomposition', and 'QR_decomposition' to the now-deprecated names.

Fixes #812.

Tested locally against Python 3.6.1 with SymPy 1.1 dev and against Python 3.5.3 with SymPy 1.0.


inst/@sym/chol.m
```diff
@@ -66,7 +66,15 @@
   elseif (nargin > 2)
     print_usage ();
   end
-  y = python_cmd('return _ins[0].cholesky(),', x);
+
+  cmd = { 'A = _ins[0]' ...
+          'try:' ...
+          '    y = A.cholesky_decomposition()' ...
```
mtmiller: This new variable could be named `L` if you like

inst/@sym/chol.m
```diff
@@ -66,7 +66,15 @@
...
+          'except AttributeError:' ...
+          '    y = A.cholesky()' ...
+          'return y' };
```
mtmiller: I see `return foo,` in some functions that return a single value, is there a preference for with or without the comma?

cbm755: Preference for without.

(Comma used to be required, someone convinced me to drop it, maybe @latot).

inst/@sym/chol.m
cbm755: I probably would, but I'm not bothered either way

cbm755:
LGTM, ok to merge?  (its marked WIP)/

mtmiller:
Ok

cbm755:
Some churn upstream, will wait a bit and see what happens before merging.

Thanks for this.

mtmiller:
The upstream name changes were reverted for SymPy 1.1, no longer needed.

Issue #771, opened 2017-04-02
Automatically test python to works internally


latot:
Hi, is related to this https://github.com/cbm755/octsympy/issues/770, actually to start using python we just try python, but in some systems like in the issue sympy is only available in 2 or 3, so the idea is can detect which version works.

An extension of this would be improve python in sympref but different, when we check the available version of python list them there to can select it, there can be some ideas, list all python versions we can found, or just list python versions with sympy.

Thx. Cya.

cbm755:
Sounds helpful to certain users but maybe a lot of out-of-scope effort for us (especially to make it work Windows/Mac).  Also, I personally dislike OS-level features ending up inside every little tool.  I mean, surely it should be Someone Else's Problem to have a working Python and Sympy install.  Does our little tool really need extra knobs and whistles to workaround the (apparently) hard problem of software installation?

I suppose I might accept a simple patch for the first idea (automatically trying "python" and "python3" to find a sympy installation).

We should also be careful we don't end up with settings that user then expect but are difficult to make work with pytave.

mtmiller:
Yeah my understanding of this was that it would involve just testing "python" and "python3" to find the first one that has a usable SymPy.

And yes, pytave currently only works with a single instance of Python, whichever was selected when the user configured and built it. It may be able to support 2 and 3 at runtime in the future, but not right now.

mtmiller:
Or maybe test all three of "python", "python2", and "python3", in some order, since some distros have already made python be python3.

latot:
I'm agree with that, i'm thinking only as an option, we can automatically select the right python version, but we should be do some way to the user can select what python works?

I mean, lets suppose sympy and all the complements works fine over python2 and 3, so we can do some option to can select what version we want to use, that i think can works in all platforms, and to see what python is we can show it like:

python (3.7)
python2 (2.4)
python3 (3.4)

What do you think?

mtmiller:
That seems out of scope to me. If the user wants to know what `python3` is, they can run `system('python3 -V')`.

If the user wants to use a particular Python, they can `export PYTHON=python3` or `setenv PYTHON python3`.

If the user wants a suitable Python to be detected and used, then they would make sure `PYTHON` is **not** in the environment and let the automatic method proposed here do its job.

latot:
Okis, so then we can send a warning if the default python version don't support some component and continue.

PR #830, opened 2017-08-03
diagnostics for debugging Python and SymPy installs

cbm755:
@mtmiller want to collaborate on exactly what this might look like?

Here's my draft.  Its more verbose but not a lot of new info yet.

Having done this, I'm thinking it could just be `assert_have_python_and_sympy` with a `verbose` flag.


mtmiller:
Great, looks good to me so far.

Agree with combining with `assert_have_python_and_sympy`. This would serve a few useful purposes: reducing duplication of code, no risk of having the tests in this function not matching the actual requirements used when setting up the interface, and it would be a sort of self-documenting annotation showing what the point of each of the tests is.

inst/private/assert_have_python_and_sympy.m
```diff
@@ -16,35 +16,117 @@
...
-function assert_have_python_and_sympy(pyexec)
-%private function
+%% -*- texinfo -*-
+%% @defun assert_have_python_and_sympy (pyexec, verbose)
+%% Check that Python and SymPy are installed and working.
+%%
+%% @seealso{sympref}
+%% @end defun
+
+function assert_have_python_and_sympy (pyexec, verbose)
```
mtmiller: `verbose = false`? Does Matlab allow default arg values?

Or `if (nargin < 2) verbose = false; end`

cbm755:
AFAIK, Matlab does not (did not?) allow default arg values.

1.  We could list `PYTHONPATH` env, although surely anyone messing with that is not the target audience...

2.  In the case when sympy is found but is too old, maybe we could give some info about the python we found (https://stackoverflow.com/questions/247770/retrieving-python-module-path/12154601#12154601)

3.  OS information?  Ubuntu version?  "uname -a" output if on unix?

Or maybe its good enough for now?

mtmiller:
Yeah I think the tests performed are good enough for now.

The only change I would make now is maybe format the "suggestions" part of the output better so they stand out. The most important feature of this is giving the user clear actions to try next.

Instead of this block of text

    Unfortunately status was non-zero: probably Python cannot import sympy.
    Have you installed SymPy?  Please try to install it and try again.
    If you have installed SymPy, perhaps for a different Python interpreter?
    Please try "setenv" as described above to change your python interpreter.

maybe space it out and itemize it?

    Unfortunately status was non-zero: probably Python cannot import sympy.

    * Do you have SymPy installed?  If not, please try to install it and try again.

    * If you do have SymPy installed, maybe it's installed for a different Python interpreter than the one we found?
      Please try "setenv" as described above to change your python interpreter.

mtmiller:
Other possible future enhancements:
* bold / color emphasis
* machine-readable output with `result = sympref ('diagnose')`
* a place to try multiple pythons, as in #771

inst/private/assert_have_python_and_sympy.m
```diff
@@ -16,35 +16,126 @@
...
+    disp ('Symbolic package diagnostics')
+    disp ('============================')
+    disp ('')
+    disp ('Python and SymPy are needed for most features of the Symbolic package.')
+    disp ('')
+    fprintf ('The Python interpreter is currently: "%s".\n', pyexec)
+    disp ('')
+    disp ('Computers may have more than one Python interpreter installed.  Depending')
+    disp ('on how your system is configured, you may need to select a different one.')
+    disp ('See "help sympref" for details, but for example, to use Python 3, try')
```
cbm755: I will try to rephrase this: sounds like something is wrong and action is required.

cbm755:
I like your suggestions.   Will do those and merge (leaving future enhancement for the future!)

PR #832, opened 2017-08-04
use Octave's `__run_test_suite__` for our tests

cbm755:
- [x] make our `octsympy_test` run `__run_test_suite__`.
- [x] Determine what version of Octave is recent enough for this.
    - [x] ~~if its 4.0, we can drop the forked code.~~
    - [x] seems to be 4.2.1, b/c of https://savannah.gnu.org/patch/?9251
- [x] drop tests from private dirs/
- [x] fix ugly output (because long full path is shown)
- [x] look at other packages and see how they handle the "tests-after-installation" problem.
- [x] review https://savannah.gnu.org/bugs/?41215
- [x] why do I need to delete tests from `private/`?  Some parts of `__run_test_suite__` explicitly skip private dirs and others do not (see `run_test_script` versus `run_test_dir`).  *update 2019:* meh, not losing sleep over this, although maybe it should be investigated upstream?
- [x] empirically, when using `__run_test_suite__`, using ctrl-c to interrupt the tests causes octave to lockup much more often (maybe 1 out of 3 tries) than before.  I'm inclined to blame the unwind with nested try-catch.  *update 2019*: tried a few times, cannot reproduce...


cbm755:
Rebased: had a hard time debugging a CI hang: seems like it was related to calling `sympref reset` (and maybe changing ipc) while figures were open.

inst/octsympy_tests.m
```diff
@@ -18,22 +18,37 @@
...
+%% until @code{pkg test} works upstream: @url{https://savannah.gnu.org/bugs/?41215}
+%%
+%% @seealso{test, runtests, doctest}
 %% @end defun
 
-%% Author: Colin B. Macdonald, David Bateman
-%% Keywords: tests
-
 function anyfail = octsympy_tests ()
+  if (compare_versions (OCTAVE_VERSION (), '4.2.1', '>='))
```
mtmiller: This should be '4.4.0' now (and corresponding comments) due to https://savannah.gnu.org/bugs/?51660

mtmiller:
This might be ready now (after rebasing and updating the min version to 4.4.0)?

inst/octsympy_tests.m
cbm755: thanks!  I'll look into merging this...

cbm755:
Hmmm, CI showing:
```
  ./@symfun/times.m .............................................. PASS    2/2   
  ./@symfun/uminus.m ............................................. PASS    2/2   
fnm(1): out of bound 0

The command "octave --path=$TRAVIS_BUILD_DIR/pytave -W --no-gui --eval "sympref ipc $IPC; r = octsympy_tests; exit(r)";" exited with 0.

$ cat fntests.log

----- 8< -----

>>>>> processing ./@symfun/symfun.m
>>>>> processing ./@symfun/symvar.m
>>>>> processing ./@symfun/times.m
>>>>> processing ./@symfun/uminus.m
The command "cat fntests.log" exited with 0.

```

Not thrilled by that `fnm(1): out of bound 0`...

cbm755:
Minimal working example posted upstream, its a regression so probably 4.4 is ok... https://savannah.gnu.org/bugs/index.php?55841

PR #833, opened 2017-08-04
ci: Python env changes, install NumPy with pip

### LINKED ISSUE #823: CI: pytave python3 run is failing early
````
Using experimental native Python/C communications with SymPy.

error: SystemError: <class 'tuple'> returned a result with an error set
````

mtmiller:
Travis now includes updated Python 2 and 3 development environments
outside of apt, so installing Python packages with apt is pointless.
Fixes #823.

Assuming the python3-with-pytave build passes (and none of the others break), this is ready to merge.


cbm755:
I thought building numpy from scratch would be  s l o w .  But it seems to use a cached binary!  

Nice.  I will merge.

mtmiller:
Thanks! I thought the same, but apparently no extra runtime.

Now that we don't have to deal with the `python-pip` vs `python3-pip` naming scheme, I will try following up with a small cleanup commit.

PR #834, opened 2017-08-04
ci: refactor Python version selection

mtmiller:
This simplifies the environment matrix a bit now that we can depend on Travis just having Python 2 and 3 installed and ready to use. We declare the version as either 2 or 3 and derive everything else from that.


mtmiller:
Passes and OK with me to merge.

PR #836, opened 2017-08-07
reduce CI dependencies

mtmiller:
See if we can build against the Octave daily snapshot now that upstream has been fixed to eliminate unnecessary dependencies on Ubuntu 14.04.


mtmiller:
CI failure, but unrelated to this change I think.

mtmiller:
This looks good now, should be ok to merge.

Issue #847, opened 2017-10-10
should vpa warn when string input contains a decimal point?


cbm755:
From a discussion at https://savannah.gnu.org/bugs/?func=detailitem&item_id=51748

Things like `vpa('0.1')` and `vpa('sqrt(2)')` are ok and recommented.

But these are likely to lose precision: `vpa('sin(0.1)')`, `vpa('2**0.1')`, `vpa('2*0.1')`: here we could consider issuing a warning.

Also
 * the docs should be expanded to warn about this
 * the docs should mention that `vpa('0.1')` is correct usage.

mtmiller:
Should we try to regex match and warn against complex expressions that may not work as intended?

Another interesting side-effect happens with an unrecognized function name

    >> vpa('abs(1)')
    ans = (sym) abs(1)
    >> vpa('sqrtt(2)')
    ans = (sym) sqrtt(2)

cbm755:
Minor correction to something I said above: `vpa('sqrt(2)')` is not so much "recommended" as "not yet deprecated", which means I'm not as worried about `vpa('abs(1)')` as I otherwise would be.

Basically, things you put in strings to `vpa` and `sym` are interpreted as Python (unless we do some parsing ourselves, which we do in some cases).

SMT has deprecated non-number strings in vpa/sym calls.

PR #848, opened 2017-10-16
vpa updates: docs and warnings

cbm755:
This adds warnings and documentation as discussed in #847

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,9 @@ octsympy 2.6.1-dev
 
           euler
 
+  * `vpa` now warns of low-precision on non-numbers strings with decimals
+    such as `vpa('cos(0.1)')`.  See "help vpa" for details.
+
   * `subs` allows multiple scalar variables to be replaced with symbolic
     matrix expressions.
 
--- a/inst/@sym/sym.m
+++ b/inst/@sym/sym.m
@@ -90,7 +90,8 @@
 %% @example
 %% @group
 %% sym(0.1)
-%%   @print{} warning: passing floating-point values to sym is dangerous, see "help sym"
+%%   @print{} warning: passing floating-point values to sym is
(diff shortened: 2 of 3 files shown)


mtmiller:
A couple minor cases that trigger a false alarm with this change

    >> vpa('2.5e+1')
    warning: string expression involving decimals is dangerous, see "help vpa"
    warning: called from
        vpa at line 118 column 7
    ans = (sym) 25.000000000000000000000000000000
    >> vpa('  0.1  ')
    warning: string expression involving decimals is dangerous, see "help vpa"
    warning: called from
        vpa at line 118 column 7
    ans = (sym) 0.10000000000000000000000000000000

These are both strings that sympy.Float can handle.

In fact any number of spaces anywhere in the number are ignored by Float, as well as the underscore that Python also allows in numeric literals

    >> vpa('1 234.0')
    warning: string expression involving decimals is dangerous, see "help vpa"
    warning: called from
        vpa at line 118 column 7
    ans = (sym) 1234.0000000000000000000000000000
    >> vpa('1_234.0')
    warning: string expression involving decimals is dangerous, see "help vpa"
    warning: called from
        vpa at line 118 column 7
    ans = (sym) 1234.0000000000000000000000000000

cbm755:
Thanks!

I fixed underscores in `vpa` and filed [an upstream bug](https://github.com/sympy/sympy/issues/13638).  The underscore thing is not universal (added in Python 3.6? https://www.python.org/dev/peps/pep-0515/) so I didn't add tests for it.

Underscores do not work properly in `sym('123_456')`: this is because sym eventually calls Sympy's `S` operator which does not support this.  I didn't yet file a bug about this.

Spaces between digits: I didn't do anything about this.  I figure Octave users are unlikely to do this ( `[1 2 3]`---is it a 3 vector or a scaler 123? ;-)

PR #849, opened 2017-10-17
check for incompatible xml.dom in older Pythons

### LINKED ISSUE #843: Check for minimum python version 2.7.3
On help@octave mailing list, Anoop has reported failures that seem to suggest we don't play nice with Windows and Python 2.  This is strange because the bundle package uses Python 2 and that has been tested...

- - - - - - - -

Based on your detailed output (thanks!), python appears to be working correctly.  It looks like some kind of newline problem...  Considering the "A" output before the error message: it looks like each string has a extraneous newline appended.  The "out" variable looks wrong too: on my system it has things like "<f>Communication established.</f>" but on Anoop's machine, there are extra newlines.

My first reaction was no one has tested the combination of Windows and Python2.  But that seems unlikely; the wiki [1] explicitly mentions Python 2.

https://github.com/cbm755/octsympy/wiki/Notes-on-Windows-installation

- - - - - - - -

Here is the relevant output:
````
GNU Octave, version 4.2.1
Copyright (C) 2017 John W. Eaton and others.
This is free software; see the source code for copying conditions.
There is ABSOLUTELY NO WARRANTY; not even for MERCHANTABILITY or
FITNESS FOR A PARTICULAR PURPOSE.  For details, type 'warranty'.

Octave was configured for "x86_64-w64-mingw32".

Additional information about Octave is available at http://www.octave.org.

Please contribute if you find this software useful.
For more information, visit http://www.octave.org/get-involved.html

Read http://www.octave.org/bugs.html to learn how to submit bug reports.
For information about changes from previous versions, type 'news'.

>> pkg load symbolic
>> syms x
OctSymPy v2.6.0: this is free software without warranty, see source.
Some output from the Python subprocess (pid 9436) might appear next.
Python 2.7 (r27:82525, Jul  4 2010, 07:43:08) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> >>>
A =
{
  [1,1] =
Communication established.

  [1,2] =
1.1.1

  [1,3] =
2.7 (r27:82525, Jul  4 2010, 07:43:08) [MSC v.1500 64 bit (AMD64)]

}
out = <output_block>
<list>
<item>
<f>
1003
</f>
<f>
Communication established.
</f>
</item>
<item>
<f>
1003
</f>
<f>
1.1.1
</f>
</item>
<item>
<f>
1003
</f>
<f>
2.7 (r27:82525, Jul  4 2010, 07:43:08) [MSC v.1500 64 bit (AMD64)]
</f>
</item>
</list>
</output_block>
Closing the Python pipe...
error: ipc_popen2: something unexpected has gone wrong in starting python
error: called from
    python_ipc_popen2 at line 138 column 7
    python_ipc_driver at line 59 column 13
    python_cmd at line 164 column 9
    valid_sym_assumptions at line 38 column 10
    assumptions at line 82 column 7
    syms at line 97 column 13
>>
````

### PR DESCRIPTION
**Pull request by mtmiller**
This fixes #843 by adding another test case to the ipc init and to `sympref diagnose` for the xml.dom library.

I need to do some more testing on this with a few Python installations and maybe clean up some of the wording, but the basic approach is what I want to do. Testing and feedback on that welcome, especially anyone with Windows.

### DIFF
--- a/inst/private/assert_have_python_and_sympy.m
+++ b/inst/private/assert_have_python_and_sympy.m
@@ -199,10 +199,10 @@ function assert_have_python_and_sympy (pyexec, verbose)
     end
     disp ('**** Your Python is too old! ****')
     disp ('')
-    disp ('The XML output shown above must have "<item>value</item>" on the same line.')
-    disp ('If your Python interpreter is an old version of Python 2, or Python 3 older')
-    disp ('than 3.2.3, please try upgrading to a more recent Python to use the Symbolic')
-    disp ('package.')
+    disp ('The XML output shown above must have "<item>value</item>" all on one line.')
+    disp ('If your Python interpreter is an older release, please try upgrading to a more')
+    disp ('recent version.  Symbolic should work with Python 2 (version 2.7.3 or newer)')
+    disp ('and Python 3 (version 3.2.3 or newer).')
     return
   end
 
--- a/inst/private/assert_have_python_and_sympy.m
+++ b/inst/private/assert_have_python_and_sympy.m
@@ -159,9 +159,9 @@ function assert_have_python_and_sympy (pyexec, verbose)
     disp ('')
   end
 
-  script = 'import xml.dom.minidom as minidom;';
-  script = [script ' doc = minidom.parseString(\"<item>value</item>\");'];
-  script = [script ' print(doc.toprettyxml(indent=\"\"))'];
+  script = ['import xml.dom.minidom as minidom; ' ...
+            'doc = minidom.parseString(\"<item>value</item>\"); ' ...
+            'print(doc.toprettyxml(indent=\"\"))'];
 
   if (verbose)
     fprintf ('Attempting to run %s -c "%s"\n\n', pyexec, script);
--- a/inst/private/assert_have_python_and_sympy.m
+++ b/inst/private/assert_have_python_and_sympy.m
@@ -146,6 +146,7 @@ function assert_have_python_and_sympy (pyexec, verbose)
   end
 
   if (verbose)
+    disp ('')
     disp ('Good, a working version of SymPy is installed.')
     disp ('')
     disp ('')
@@ -153,8 +154,8 @@ function assert_have_python_and_sympy (pyexec, verbose)
     disp ('----------------------------------')
     disp ('')
     disp ('The XML DOM library is used by Symbolic for passing values to and from Python.')
-    disp ('Some older versions of Python formatted XML output differently.  If you have')
-    disp ('a Python that is older than 2.7.3 or 3.2.3, this will probably fail.')
+    disp ('Some older versions of Python formatted XML output differently.  As long as you')
+    disp ('have any reasonably recent version of Python, this should pass.')
     disp ('')
   end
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: inst/private/assert_have_python_and_sympy.m [hunk 1]
```diff
@@ -146,6 +146,67 @@ function assert_have_python_and_sympy (pyexec, verbose)
...
+    disp ('')
+    disp ('The XML DOM library is used by Symbolic for passing values to and from Python.')
+    disp ('Some older versions of Python formatted XML output differently.  As long as you')
+    disp ('have any reasonably recent version of Python, this should pass.')
+    disp ('')
+  end
+
+  script = 'import xml.dom.minidom as minidom;';
+  script = [script ' doc = minidom.parseString(\"<item>value</item>\");'];
+  script = [script ' print(doc.toprettyxml(indent=\"\"))'];
```
cbm755: can be one assignment across multiple lines?

inst/private/assert_have_python_and_sympy.m
```diff
@@ -146,6 +146,67 @@ function assert_have_python_and_sympy (pyexec, verbose)
...
+
+  if (isempty (strfind (xmlout, '<item>value</item>')))
+    if (~ verbose)
+      error ('OctSymPy:oldxmldom', ...
+             ['Python XML output is not compatible, probably an old version of Python?\n' ...
+              '    Try "sympref diagnose" for more information.'])
+    end
+    disp ('**** Your Python is too old! ****')
+    disp ('')
+    disp ('The XML output shown above must have "<item>value</item>" on the same line.')
```
cbm755: "all on one line"?

inst/private/assert_have_python_and_sympy.m
```diff
@@ -146,6 +146,67 @@ function assert_have_python_and_sympy (pyexec, verbose)
...
+    disp ('If your Python interpreter is an old version of Python 2, or Python 3 older')
+    disp ('than 3.2.3, please try upgrading to a more recent Python to use the Symbolic')
+    disp ('package.')
```
cbm755: Maybe its just hard to phrase this briefly but looks a little odd to explicitly mention py3 version but not not py2 version.

cbm755:
This seems a bit heavy compared to just a version check, but I guess it self-documents against future xml.dom change too.  Let me know when you're happy with the phrasing and I'll merge.

inst/private/assert_have_python_and_sympy.m
mtmiller: Yeah this is the awkward part to write, especially for people who don't care what the difference is between Python 2 and 3. Maybe just simplify it like

> If your Python interpreter is an older release, please try upgrading to a more recent Python to use the Symbolic package.  Symbolic works with any Python 2.7 version 2.7.3 or newer and any Python 3 version 3.2.3 or newer.

mtmiller:
> This seems a bit heavy compared to just a version check

But it also documents the justification for the minimum version, and it allows the possibility of some third-party Python implementation working even if they have a totally different versioning scheme.

inst/private/assert_have_python_and_sympy.m
cbm755: #bikeshed ;-)

> If your Python interpreter is an older release, please try upgrading to a
> more recent version.  Symbolic should work with Python 2 version 2.7.3
> or newer and Python 3 version 3.2.3 or newer.

or

> If your Python interpreter is an older release, please try upgrading to a
> more recent version.  Symbolic should work with Python 2 (version 2.7.3
> or newer) and Python 3 (version 3.2.3 or newer).

mtmiller:
I'm happy with this now, thanks for the feedback.

cbm755:
Looks like no one tested on Windows; it fails #871 :(

PR #851, opened 2017-11-19
fortran: Fix typo in documentation.

alexvong1995:
Hello,

The documentation for the `fortran` function mentions that it generates C code,
which I believe is a typo.


cbm755:
Thanks!

Issue #863, opened 2018-02-14
Improve suppression of output when first creating symbolic variables


Alexander-Wilms:
Since other packages don't show any output when one of their functions is executed, I was wondering how much of the output is deliberate?
Could it be suppressed or limited to just `OctSymPy v2.6.0: this is free software without warranty, see source.`?

> OctSymPy v2.6.0: this is free software without warranty, see source.
> Initializing communication with SymPy using apopen2() pipe.
> Some output from the Python subprocess (pid 15) might appear next.
> Python 2.7.12 (default, Jan 10 2018, 06:15:19)
> [GCC 6.2.0] on linux2
> Type "help", "copyright", "credits" or "license" for more information.
> >>> >>>
> OctSymPy: Communication established.  SymPy v1.1.1.

cbm755:
`sympref quiet on` helps a bit, but `Python 2.7...` down to `>>> >>>` is a technical limitation.  The long-term fix is to use [https://bitbucket.org/mtmiller/pytave](pytave).

See also #273.

mtmiller:
In case anyone's interested, I played with a local change a while ago that quiets this a lot:

    $ octave
    >> sympref quiet on
    >> syms x
    >>> >>> 
    >> x
    x = (sym) x
    >>

Alexander-Wilms:
`sympref quiet on` doesn't seem to work for me either:

```
>> sympref quiet on
>> syms x
Python 2.7.12 (default, Jan 10 2018, 06:15:19)
[GCC 6.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> >>>
>> x
x = (sym) x
```

But it's good to know that in the long-term it'll be fixed.

cbm755:
@Alexander-Wilms That is---unfortunately---the expected output with `sympref quiet on`.

@mtmiller how do you do that?  If we could make it that quiet, we could consider making `sympref quiet on` the default.

mtmiller:
I'll open a PR in a few minutes to share what I have. I did this several months ago just to see what could be done, but I apparently didn't think it merge-worthy at the time.

mtmiller:
As I mentioned on #864, my proposed change only works with Python 3.2 or later. With Python 2 there's no way to silence the banner given that we have to run python in interactive mode.

PR #864, opened 2018-02-14
[wip] Make popen2 ipc as quiet as possible

mtmiller:
This is a proof of concept to attempt to run python without the normal interactive startup banner. When the `quiet` sympref is on, try to test whether the version of python being used accepts the `-q` option, and use it.

Re #863.

### DIFF
--- a/inst/private/python_have_quiet_flag.m
+++ b/inst/private/python_have_quiet_flag.m
@@ -1,0 +1,30 @@
+%% Copyright (C) 2018 Mike Miller
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+function r = python_have_quiet_flag (pyexec)
+%private function
+
+  if (ispc () && ~ isunix ())
+    [st, out] = system ([pyexec ' -q -c True 2> NUL']);
+  else
+    [st, out] = system ([pyexec ' -q -c True 2> /dev/null']);
+  end
+
+  r = (st == 0);
+
+end
--- a/inst/private/python_ipc_popen2.m
+++ b/inst/private/python_ipc_popen2.m
@@ -76,12 +76,15 @@ function [A, info] = python_ipc_popen2(what, cmd, varargin)
     end
     pyexec = sympref('python');
     assert_have_python_and_sympy (pyexec)
+    quiet_flag = ~verbose && python_have_quiet_flag (pyexec);
 
     if (ispc() && ~isunix() && compare_versions (OCTAVE_VERSION (), '4.0.2', '<='))
       if (verbose)
         disp('Using winwrapy.bat workaround for bug #43036 (Octave <= 4.0.2, on Windows)')
       end
       [fin, fout, pid] = popen2 ('winwrapy.bat', pyexec);
+    elseif (quiet_flag)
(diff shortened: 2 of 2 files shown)


mtmiller:
I do almost all my Python work in Python 3 now, and now that I look at this again I guess Python 2.7 doesn't support `-q` at all. It does work with Python 3.2 or later.

inst/private/python_have_quiet_flag.m
```diff
@@ -0,0 +1,30 @@
...
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <http://www.gnu.org/licenses/>.
+
+function r = python_have_quiet_flag (pyexec)
+%private function
+
+  if (ispc () && ~ isunix ())
+    [st, out] = system ([pyexec ' -q -c True 2> NUL']);
```
cbm755: has this actually been tested on Windows?

cbm755:
This looks nice to me.  Things I want to think about before I merge:

- [x] It has no effect with winwrapy (I think).
- [ ] tested on Windows
- [ ] Does it play nice with the `py.exe` shipped with in the "batteries included" Windows bundle?  Probably, because `py.exe` is Python 2.

We might also think about making `quiet on` the default on Python 3.  Or maybe just using `-q` even when `quiet off` (and removing the "Some output from the Python subprocess might appear next.").

Maybe `sympref quiet` should be deprecated and become `sympref verbose`.

inst/private/python_have_quiet_flag.m
mtmiller: No, testing very welcome

mtmiller:
How about making `py.exe` be Python 3?

cbm755:
Maybe, but I just use [this](http://www.orbitals.com/programs/pyexe.html) which is Python 2.  I don't really want to get involved in building such.

cbm755:
"help wanted" label is for windows testing: occasionally I test on an old windows 10 tablet, I'll do that when I get a chance but if anyone else can, it would be much appreciated.

cbm755:
At a bash prompt, `python3 -q -i -c "import sys; sys.ps1=''"` seems to silence the initial `>>>`.

So maybe we can do that instead of waiting for `python_header.py` to set `ps1`.  And have a totally silent startup on Python 3.

cbm755:
Similarly, putting `python_header.py` as `$PYTHONSTARTUP` could be done.  I originally avoided this sort of thing because I'm always worried about paths on non-unix systems.

cbm755:
Even better: on both Python 2 and 3,  `python -i -c ...` behaves like `-q` (no banner).  Tested with 2.5.2, 2.7.14 and 3.6.4.

cbm755:
I had a chance to test this on Windows: it sort of works (after hacking around #871).

I get:
````
Symbolic pkg v2.6.1-dev: ;sys.ps2=.... ... ;sys.ps2=... ... ;sys.ps2=;sys.ps2=;sys.ps2=;sys.ps2=Python communication link active, SymPy v1.1.1.
````
So not quite right yet.

**Comment by cbm755** [comment]
Also: the self-contained `py.exe` has `-i -c` do the same as `-c`.  However, it does support setting the environment var PYTHONINSPECT (as does the official python client).

**Comment by mtmiller** [comment]
Unfortunately, CPython exits when using `PYTHONINSPECT=Yes` and stdin is not a terminal. They are not quite the same, despite what `python -h` implies.

Are you able to test that `PYTHONINSPECT=Yes` does work for `py.exe` when it is popen'd? Based on what I've read this will work the same as CPython, stdin must be a terminal.

I used `cat` and `socat` to test running CPython with stdin and stdout assigned to pipes:

```
$ cat | PYTHONINSPECT=x python -c True | cat
# this actually waits for one newline to be entered, then exits with success
$ PYTHONINSPECT=x socat STDIN SYSTEM:'python -c True',pipes | cat
# exits spontaneously after about half a second
```

cbm755:
No I cannot make PYTHONINSPECT env variable work with `py.exe` and popen :(

I think it works with `system` but not with `popen2`.

Maybe easiest to fix `py.exe` so that `-i -c` works as in CPython.  The source is easy to understand but I don't know how to build it.  Could just ask the author...

cbm755:
I read the py.exe code and interactive mode uses the `code` module.  So I tried this:
````
python2 -c "import sys; sys.ps1=''; import code; cons = code.InteractiveConsole(); cons.interact(banner='')"
````
This works nicely on a terminal but not in a pipe.

PR #882, opened 2018-04-06
Make it easier to customize the default python interpreter

### LINKED ISSUE #879: Fedora package should not call unversioned `python`
@mtmiller do you maintain the Debian package?  Is there is a similar issue to this?

[Fedora packaging guidelines](https://fedoraproject.org/wiki/Packaging:Python#Multiple_Python_Runtimes) say:

> However, packages in Fedora MUST NOT depend on
> where /usr/bin/python happens to point but instead
> MUST call the proper executable for the needed python
> major version directly, either /usr/bin/python2 or
> /usr/bin/python3 as appropriate.

We do call `python` by default.  The Fedora package also depends explicitly on `python2-sympy`.  So we will break when Fedora switches the default Python to 3.

Some options:

  * Make the Fedora package depend on both `python2-sympy`
    and `python3-sympy`.  Then it will work whether `python` is 2
    or 3.  But its useless dependency bloat.

  * We could `sed` the `symbolic` pkg code (during Fedora package
    build) to call `python3` by default (when `$PYTHON` is unset).
    Then the package should depend on `python3-sympy`.

I'm leaning towards the second option.

cbm755:
Provide a `private/defaultpython.m` that distributors/vendors can
tweak.  Fixes #879.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -25,6 +25,9 @@ octsympy 2.6.1-dev
 
   * Fixes for several bugs mentioning "immutable matrices".
 
+  * Added a `private/defaultpython.m` to make it easier for distributors
+    to control which python interpreter is used by default.
+
   * Various bug fixes and documentation updates.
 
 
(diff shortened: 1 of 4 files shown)


mtmiller:
This looks great, thank you.

I will try to get the daily build working within the next day or two so Travis will pass again.

Issue #913, opened 2018-10-30
equation solve issue in Matlab


OsellaG:
Hi,
I'm trying to use octsympy in matlab  (Octave is OK but I need some built-in matlab functions)  but I get some problem.

Some parts of the tool work fine but I cannot put at work the function solve.
syms x
solve(x^2,x)
give me the following error message:

Index exceeds matrix dimensions.

Error in python_ipc_sysoneline (line 146)
  A = extractblock(out(ind(1):end));

Error in python_ipc_driver (line 63)
        [A, db] = python_ipc_sysoneline(what, cmd, false, varargin{:});

Error in python_cmd (line 140)
  [A, db] = python_ipc_driver('run', cmd, varargin{:});

Error in sym/solve (line 152)
    out = python_cmd (cmd, varargin{:});

Anyone can help me?
Thanks in advance for any help.

OsellaG:
The problem is in the low level Python command management

the tests__python_cmd.m stop with sintax error at ther following line

 r = python_cmd ('return 42 if sys.version_info >= (3,0) else long(42)');

cbm755:
That's strange, that's an error from Python not Matlab.

What does `sympref diagnose` output?

My best guess here is that the PATH is different on Matlab and its running a different Python than is run on Octave.

mtmiller:
If the syntax error is referring to the `return 42 if …` syntax, then yeah, it seems like your environment is using Python 2.4 or something older, which is not supported by octsympy.

You can check what Python is being used with

    system([sympref('python') ' --version']);

@cbm755 is that the best way to check this? I don't see `sys.version` reported in any verbose output or diagnostics anymore.

cbm755:
> is that the best way to check this?

Can also do:
```
>> s = python_cmd('return sys.version')
s = 2.7.15 (default, Oct 15 2018, 15:24:06) 
[GCC 8.1.1 20180712 (Red Hat 8.1.1-5)]
```

I agree that `sympref diagnose` should be reporting the Python version.

cbm755:
Are there two different things going on here?  I cannot reproduce either one.  It looks in Comment 2 like Python 2.4 is being run.  But then later comments (https://github.com/cbm755/octsympy/issues/913#issuecomment-434587134) show Python 3.7.1 and different errors...

@OsellaG, can you report the output of `s = python_cmd('return sys.version')`?

cbm755:
[...]

PR #914, opened 2018-11-02
diagnostics should report Python version

cbm755:
Don't bother checking unless we're in verbose (diagnostics) mode.


cbm755:
@mtmiller like this?  See discussion in #913.

PR #928, opened 2019-01-24
implement "find"

cbm755:
Probably these could share a common helper function without too much effort, but haven't done that here.


cbm755:
logical refactor for later, just `find` for now.

PR #939, opened 2019-03-06
Remove evalpy

cbm755:
Maybe not now, but soon should merge this.  Originally I thought I'd keep it around in pytave matures...


cbm755:
I suppose SemVer says this should wait for 3.0.0.

But I'm quite worried about practical *real-life confusion*  b/w this old `evalpy` and the `pyeval` in Pythonic.

Its been deprecated since 2017-02 (version 2.5.0).  And I don't think this project has been following SemVer (more like one or  two cycles of minor version deprecation followed by dropping).  We've had plenty of other changes which are really "pre 1.0.0" by SemVer standards.

@mtmiller: do you feel strongly that this should wait for 3.0.0?

mtmiller:
I'm not very worried about confusion, especially because I plan to always steer users away from `pyeval` anyway. I think it's a noble goal to clean up the interface of OctSymPy to get rid of these generic looking functions `evalpy` and `python_cmd`. I think that this could wait until you are ready for 3.0.0. Now is as good a time as any to try to aim for SemVer, right?

PR #946, opened 2019-03-22
Add sinc function

cbm755:


mtmiller:
Given that it is very confusing, I think the help text is clear as you have it now. Unfortunately users will probably still be confused because the `disp` appearance of the expression will always be different.

inst/@sym/sinc.m
```diff
@@ -56,7 +56,7 @@
 %% whereas the output @code{sinc(π⋅x)} is the equivalent SymPy expression
 %% written in terms of the unnormalized sinc function.
 %%
-%% Note if you conversion back into an Octave function works correctly as
```
zmarantz: %% Note conversion back into an Octave function works correctly as

PR #947, opened 2019-03-23
rewrite: add example of "sincos"

### LINKED ISSUE #945: rewrite as "sincos"
Following up on some comments in https://savannah.gnu.org/bugs/?55968#comment6 which are being conflated with an unrelated issue of normalized versus unnormalized sinc function.

@mtmiller: on Matlab R2015b I had someone run:
```
>> rewrite(tan(x)/x, 'sinco') 
Error using sym/rewrite (line 62)
The value of 'T' is invalid. It must satisfy the function:
@(x)any(strcmp(x,{'sin','cos','tan','sincos','exp','log','heaviside','sqrt'})).
 
>> rewrite(tan(x)/x, 'sinc') 
Error using sym/rewrite (line 62)
The value of 'T' is invalid. It must satisfy the function:
@(x)any(strcmp(x,{'sin','cos','tan','sincos','exp','log','heaviside','sqrt'})).
 
>> rewrite(tan(x)/x, 'sin') 
 
ans =
 
-sin(x)/(x*(2*sin(x/2)^2 - 1))
 
>> rewrite(tan(x)/x, 'sincos')
 
ans =
 
sin(x)/(x*cos(x)) 
```
These are reasonable results.  Note difference b/w `sin` and `sincos`.

Maybe other versions of Matlab's SMT give something different...

We should:

  * support `sin` (we do)
  * ideally for SMT compat, we could support `sincos` (unlikely that we do).
  * `sinco` should be either an error or silently do nothing, same as `zzzzZZzz`

Finally, I think `sinc` should rewrite in terms of sinc function.  It certainly should not be some abbreviation for `sincos`.

cbm755:
Fixes #945.

### DIFF
--- a/inst/@sym/rewrite.m
+++ b/inst/@sym/rewrite.m
@@ -1,4 +1,4 @@
-%% Copyright (C) 2016 Colin Macdonald
+%% Copyright (C) 2016, 2019 Colin Macdonald
 %%
 %% This file is part of OctSymPy.
 %%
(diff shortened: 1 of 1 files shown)


cbm755:
@mtmiller ok to close #945 with this?

PR #954, opened 2019-04-02
monkey patch sympy for sinc disp and swap lambertw args

cbm755:
We patch the pretty printer to do this.  The `srepr` ("pickle") remains
unchanged so any actual mathematics are uneffected.

@mtmiller what do you think of this?

One downside is `sym('LambertW(5,6)')` will now be "wrong", at least for someone expecting it to match `lambertw(6,5)`.  But passing raw SymPy strings to `sym` isn't recommended anyway...

So this would fix things for normal users and potentially frustrate someone who knows a lot about SymPy (e.g., is consulting SymPy docs).

So should we:

1.  Do both (this PR).
2.  Do `sinc_un` but don't do `LambertW`.  The logic is `sinc` is worse b/c at least `lambertw` -> `LambertW` capitalization hints that something has happened...
3.  Do neither (e.g., acrument of technical debt for minimal benefit).

### DIFF
--- a/inst/@sym/lambertw.m
+++ b/inst/@sym/lambertw.m
@@ -1,4 +1,4 @@
-%% Copyright (C) 2015, 2016, 2018 Colin B. Macdonald
+%% Copyright (C) 2015, 2016, 2018-2019 Colin B. Macdonald
 %%
 %% This file is part of OctSymPy.
 %%
@@ -32,25 +32,21 @@
 %% lambertw(x)
 %%   @result{} (sym) LambertW(x)
 %% lambertw(2, x)
-%%   @result{} (sym) LambertW(x, 2)
+%%   @result{} (sym) LambertW(2, x)
 %% @end group
 %% @end example
-%% (@strong{Note} that the branch @var{k} must come first in the
-%% input but it comes last in the output.)
 %%
 %% Also supports vector/matrix input:
 %% @example
 %% @group
 %% syms x y
 %% lambertw([0 1], [x y])
-%%   @result{} (sym) [LambertW(x)  LambertW(y, 1)]  (1×2 matrix)
+%%   @result{} (sym) [LambertW(x)  LambertW(1, y)]  (1×2 matrix)
 %% @end group
 %% @end example
 %% @seealso{lambertw}
 %% @end defmethod
 
-%% Author: Colin B. Macdonald
-%% Keywords: symbolic
 
 function W = lambertw(k, x)
   if (nargin == 1)
--- a/inst/@sym/lambertw.m
+++ b/inst/@sym/lambertw.m
@@ -30,9 +30,9 @@
 %% @group
 %% syms x
 %% lambertw(x)
-%%   @result{} (sym) LambertW(x)
+%%   @result{} (sym) lambertw(x)
 %% lambertw(2, x)
-%%   @result{} (sym) LambertW(2, x)
+%%   @result{} (sym) lambertw(2, x)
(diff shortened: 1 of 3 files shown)


cbm755:
Hmmm, this may require bumping the minimum sympy version.  To be honest, running this with anything other than the latest sympy is asking for trouble: we shouldn't be afraid to bump the requirements!

mtmiller:
> One downside is `sym('LambertW(5,6)')` will now be "wrong", at least for someone expecting it to match `lambertw(6,5)`. But passing raw SymPy strings to `sym` isn't recommended anyway...

I don't see that as a problem. But one more change could be to rewrite the name of the function also, to make it clear it's not the same as SymPy `LambertW`.

mtmiller:
Like this aba0a0e77ed2

cbm755:
Yes, that seems like a great idea!  Sure push to this branch

mtmiller:
So one more idea that might help this whole `sinc` confusion would be to just rewrite it as `sin(x)/x` to avoid all doubt about the definition. But using rewrite ends up showing a piecewise defintion which looks much worse

    >> syms x; sinc (x)
    ans = (sym)
    
               ⎛⎛sin(π⋅x)         ⎞           ⎞
      Piecewise⎜⎜────────, π⋅x ≠ 0⎟, (1, True)⎟
               ⎝⎝  π⋅x            ⎠           ⎠

cbm755:
Well it pretty prints better than that, see `help @sym/sinc`...  I'd rather not do this by default.  If people want sinc give them sinc---well some kind of sinc

We can also add `sincpi` to upstream SymPy and then `sinc(x)` would become `sincpi(x)`. `mpmath` (the numerical part of sympy) already does this.

cbm755:
We could also implement `@sym/sincpi` (synonym for `@sym/sinc`) and `@sym/sinc_un` for completeness (after sincpi was done upstream)

mtmiller:
> Hmmm, this may require bumping the minimum sympy version.

The only consideration is Ubuntu stable users get SymPy 1.1.1 if they install it with apt. So some pool of users would have to switch to installing SymPy with pip.

Other than that, yes let's bump the minimum when it makes sense.

cbm755:
This is true, but they get a rather broken SymPy by doing that...

Or we can write these _print functions without using `func_name=` (copy-paste `_print_Function` instead of calling it).  That's probably better for 2.8.0.

cbm755:
TODO: any python_header.py changes needs to be duplicated somewhere in `ipc_native`.

mtmiller:
One more suggestion: f0cd67341829

    >> syms x; sinc (x)
    ans = (sym) sinc(x)
    >> syms x; rewrite (sin (x)/x, 'sinc')
    ans = (sym)
    
          ⎛x⎞
      sinc⎜─⎟
          ⎝π⎠
    
    >> syms x; rewrite (sin (pi*x)/(pi*x), 'sinc')
    ans = (sym) sinc(x)

cbm755:
Wow!  ok I guess we should do that!

Only downside I can think of is the one I mentioned earlier for LambertW: `sym('sinc(x)')` would not work properly.  But `sym('sinc_un(x)')` doesn't work either...

cbm755:
> (copy-paste _print_Function instead of calling it). That's probably better for 2.8.0.

meh, that function is rather big...  I think bumping sympy version is easier.

mtmiller:
Thanks for pulling that in and cleaning this up, this is looking really good now.

PR #960, opened 2019-04-06
diagnose: refer to wiki if Python fails on Windows

### LINKED ISSUE #959: make `sympref diagnose` point Windows users to the wiki when python missing?
Should `sympref diagnose` include a message specifically for Windows users when the basic Python interpreter test fails? Something like this

```
Unfortunately, that command failed!
We expected to see "status = 0" and "output = Python says hello".

  * Is there an error message above?

  * Do you have Python installed?

  * Please try using "setenv" as described above.

  * Most systems search the PATH environment when looking for commands.
    Your path seems to be:

…

  * Most Windows users should install symbolic package bundle that includes Python.
    Please refer to the installation notes at
    <https://github.com/cbm755/octsympy/wiki/Notes-on-Windows-installation>

```

_Inspired by https://lists.gnu.org/archive/html/help-octave/2019-04/msg00027.html_

mtmiller:
Fixes #959.

### DIFF
--- a/inst/private/assert_have_python_and_sympy.m
+++ b/inst/private/assert_have_python_and_sympy.m
@@ -77,6 +77,13 @@ function assert_have_python_and_sympy (pyexec, verbose)
     disp ('    Your path seems to be:')
     disp ('')
     disp (getenv ('PATH'))
+
+    if (ispc () && (~isunix ()))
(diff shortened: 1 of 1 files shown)


cbm755:
LGTM!

Issue #963, opened 2019-04-14
Symbolic matrices do not accept sub-index assignment accepted by regular matrices


jose-matos:
The following code works for regular matrices:

```octave
# Double precision (works)

N = zeros(3);
# here there is a mismatch the left hand side selection is a column but the right hand side is a row:
N (1:2:3, 1) = 1:2:3
# the index selection is empty
N(1:-1:2,1)=1


# vpa version:
N = sym(zeros(3))

# this fails
N(1:2:3,1) = sym(1:2:3)
# this fails as well
N(1:-1:2,1)=1
```

It would be nice to maintain the code compatible between both versions.

cbm755:
The later case has uncovered a bunch of other corner cases, #964 and #965.  Thanks for this.

jose-matos:
Thank you for the fixes. :-)

PR #968, opened 2019-04-16
Various indexing fixes for corner cases

cbm755:


cbm755:
I don't really expect anyone to review this messy part of the code but I'll CC @jose-matos just in case, as the reporter of #963 which triggered all this...

PR #970, opened 2019-04-22
[maint] CI: use docker

cbm755:
- [x] replace ppa etc with docker
- [x] don't test python2 in every possible combination
- [x] put `-W` ~~`--no-gui`~~ back
- [x] re-enable pythonic, ~~or wait for its pkg form?~~
- [x] check if the column width stuff is still needed.  update: I didn't check, just put it back.  IIRC, we had non-deterministic failures in doctests without this, not sure if this is still true or not.
- [x] double check that `cat fntests.log` is unnecessary with `make test`.  Update: needed, in principle only when tests fail, but if tests don't fail, no one will look ;-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,16 @@
+## SPDX-License-Identifier: FSFAP
+## Copyright (C) 2014-2019 Colin B. Macdonald
+## Copyright (C) 2016-2017 Abhinav Tripathi
+## Copyright (C) 2016 Marco Falke
+## Copyright (C) 2016-2019 Mike Miller
+##
+## Copying and distribution of this file, with or without modification,
+## are permitted in any medium without royalty provided the copyright
+## notice and this notice are preserved.  This file is offered as-is,
+## without any warranty.
+##
+## .travis.yml - Travis CI configuration for Octave's Symbolic package
+
 dist: xenial
 language: generic
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ before_install:
 - docker pull mtmiller/octave:${OCT}
 
 script:
-- docker run --name=oc --detach --init --volume=$PWD:/octsympy:z mtmiller/octave:${OCT} sleep inf
+- docker run --name=oc --detach --init --env=LC_ALL=C.UTF-8 --volume=$PWD:/octsympy:z mtmiller/octave:${OCT} sleep inf
 - docker exec oc pip${PYTHON_VERSION} install sympy==${SYMPY_VER}
 - docker exec oc make -C octsympy PYTHON=python${PYTHON_VERSION} test
 - if [ "x$DOCTEST" = "xyes" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,18 @@ matrix:
   - env: PYTHON_VERSION=3 PYTHONIC=no  SYMPY_VER=1.4 OCT=4.4   DOCTEST=yes COLUMNS=80
   - env: PYTHON_VERSION=3 PYTHONIC=no  SYMPY_VER=1.4 OCT=5     DOCTEST=yes COLUMNS=80
   - env: PYTHON_VERSION=3 PYTHONIC=no  SYMPY_VER=1.4 OCT=devel DOCTEST=no  COLUMNS=80
+  - env: PYTHON_VERSION=3 PYTHONIC=yes SYMPY_VER=1.4 OCT=devel DOCTEST=no  COLUMNS=80
+  - env: PYTHON_VERSION=3 PYTHONIC=yes SYMPY_VER=1.4 OCT=5     DOCTEST=no  COLUMNS=80
+  - env: PYTHON_VERSION=2 PYTHONIC=yes SYMPY_VER=1.4 OCT=devel DOCTEST=no  COLUMNS=80
+  - env: PYTHON_VERSION=2 PYTHONIC=yes SYMPY_VER=1.4 OCT=5     DOCTEST=no  COLUMNS=80
   allow_failures:
   - env: PYTHON_VERSION=3 PYTHONIC=no  SYMPY_VER=1.4 OCT=devel DOCTEST=no  COLUMNS=80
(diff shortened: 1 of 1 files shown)


mtmiller:
> put `-W --no-gui` back

Isn't that redundant, `-W` implies `--no-gui`?

.travis.yml
```diff
@@ -1,83 +1,76 @@
...
-    else
-        export IPC=default;
-    fi
-  - octave --path=$TRAVIS_BUILD_DIR/pythonic -W --no-gui --eval "sympref ipc $IPC; r = octsympy_tests; exit(r)";
-  - cat fntests.log
-  - if [ "x$DOCTEST" = "xyes" ]; then
-        octave --path=$TRAVIS_BUILD_DIR/pythonic -W --no-gui --eval "pkg load doctest; sympref ipc $IPC; syms x; r = doctest('.'); exit(~r)";
-    fi
-  - popd
+- docker run --cidfile=/tmp/container --detach --init --volume=$PWD:/octsympy:z mtmiller/octave:${OCT} sleep inf
```
mtmiller: Another option, maybe less verbose, is to give the container a name, any arbitrary label works.

```shell
docker run --name=myname --detach ...
docker exec myname ...
...
docker stop myname
```

.travis.yml
```diff
@@ -1,83 +1,76 @@
...
+- docker exec $(cat /tmp/container) pip${PYTHON_VER} install sympy-${SYMPY_VER}
```
mtmiller: Should we install things in `before_script`?

.travis.yml
```diff
@@ -1,83 +1,76 @@
...
+- docker exec $(cat /tmp/container) make PYTHON=python${PYVER} test
+- if [ "x$DOCTEST" = "xyes" ]; then
+      docker exec $(cat /tmp/container) octave --eval "pkg install -forge doctest";
```
mtmiller: Likewise?

cbm755:
all those comments sound good... I'm about to do some travel, feel free to push to this branch if you want.

.travis.yml
cbm755: Probably.  I wasn't sure about the running docker persisting.

mtmiller:
@cbm Confused about the matrix, you have `PYTHON_VERSION=2` for some combinations but not others. Is that just to ensure all versions of Octave and SymPy are tested once against 2.7 but not all cross combinations? If that's the intent maybe it would be clearer to group the `PYTHON_VERSION=2` rows so they stand separately?

mtmiller:
I don't understand what's going on with all Python 3 tests hanging and timing out on `python_cmd`. I can reproduce locally, but still not understanding yet.

Outside of Docker on my system

```
$ octave --version | head -n1
GNU Octave, version 5.1.0
$ python --version
Python 3.7.2
$ octave --path $PWD/inst --eval "test python_cmd"
Symbolic pkg v2.7.1+: Python communication link active, SymPy v1.2.
  File "<stdin>", line 4, in <module>
  File "/home/mike/.pyenv/versions/sympy/lib/python3.7/site-packages/sympy/core/singleton.py", line 110, in __getattr__
    name, self))
AttributeError: Attribute 'make_an_attribute_err_exception' was not installed on SymPy registry S
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 12, in octoutput_drv
  File "<stdin>", line 99, in octoutput
ValueError: octoutput does not know how to export type <class 'type'>
warning: test: file /home/mike/src/octave/octsympy/inst/python_cmd.m leaked file descriptors
PASSES 46 out of 46 tests
```

Ok, a couple exceptions thrown and a warning about leaked file descriptors, but all tests pass and Octave returns to the shell.

Now in Docker, replicating the Travis build rules

```
$ docker run --name=oc --detach --init --volume=$PWD:/octsympy mtmiller/octave sleep inf
$ docker exec oc pip3 install sympy==1.2
$ docker exec oc octave --version | head -n1
octave: X11 DISPLAY environment variable not set
octave: disabling GUI features
GNU Octave, version 5.1.0
$ docker exec oc python3 --version
Python 3.6.7
$ docker exec oc env PYTHON=python3 octave --path /octsympy/inst --eval "test python_cmd"
octave: X11 DISPLAY environment variable not set
octave: disabling GUI features
Symbolic pkg v2.7.1+: Python communication link active, SymPy v1.2.
  File "<stdin>", line 2
    _ins = []
            ^
SyntaxError: 'ascii' codec can't decode byte 0xe6 in position 17: ordinal not in range(128)
  File "<stdin>", line 2
    octoutput_drv("PYTHON: successful variable import")
    ^
IndentationError: unexpected indent
  File "<stdin>", line 1
    except:
         ^
SyntaxError: invalid syntax
  File "<stdin>", line 1
    echo_exception_stdout("while copying variables to Python")
    ^
IndentationError: unexpected indent
  File "<stdin>", line 1
    raise
    ^
IndentationError: unexpected indent
Waiting......^C
```

mtmiller:
Got it, locale needs to be set because there is some UTF-8 going on.

cbm755:
#972 needs fixed but I don't think that needs to hold this PR back

cbm755:
@mtmiller does this look ok?  I'll probably rebase a bit before merging.

mtmiller:
> @mtmiller does this look ok? I'll probably rebase a bit before merging.

Pushed one more small reorganizing nit, but yeah this looks great to me, feel free to rebase and rewrite any way you want.

cbm755:
Saw this failure, hopefully something intermittent:
```
0.14s$ if [ "x$DOCTEST" = "xyes" ]; then docker exec oc make -C octsympy PYTHON=python${PYTHON_VERSION} doctest; fi
OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "process_linux.go:90: adding pid 4176 to cgroups caused \"failed to write 4176 to cgroup.procs: write /sys/fs/cgroup/cpu,cpuacct/docker/c39a85604c19eabf1e0b9d3719ab98781e844e7fea5bdc99be361164048c1270/cgroup.procs: invalid argument\"": unknown
The command "if [ "x$DOCTEST" = "xyes" ]; then docker exec oc make -C octsympy PYTHON=python${PYTHON_VERSION} doctest; fi" exited with 126.
```

PR #980, opened 2019-05-03
[maint] CI: install and use pythonic with pkg

mtmiller:
The Pythonic package can now be installed with `pkg`.~, on the pkg branch only for now. This PR is a test to make sure that will work with OctSymPy once it's merged into master.~

~Not intended to be merged until [octave-pythonic!6](https://gitlab.com/mtmiller/octave-pythonic/merge_requests/6) is done.~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ script:
 - stty cols $COLUMNS rows 40
 - tput cols; stty size
 - if [ "x$PYTHONIC" = "xyes" ]; then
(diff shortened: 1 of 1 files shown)


mtmiller:
Still needs a little more work.

mtmiller:
@cbm755 This works now, ready to merge, unless you have any changes to suggest.

The part that kept tripping me up was making sure the container has the right `PYTHON`/`PYTHON_VERSION` environment variables, both when the pythonic package is installed **and** when running the test suite, because of `test sympref`.

cbm755:
This looks good to me!

I guess it would be easier if there was an env var for IPC?  Or maybe we should just default to `native` whenever Pythonic is available?  At least longer term, I'll file an issue for that if there isn't one....

PR #991, opened 2019-06-04
Use Pythonic by default when available

### LINKED ISSUE #982: default to ipc=native if pythonic is available
Now that pythonic is a package, and presumably we think it will eventually be a dependency, maybe we should default to `ipc=native` and fall back to `ipc=popen2`.

Or we can wait until there is at least an (alpha) released version pythonic.

See merged PR #980, where some of the testing logic could be simpler...

cbm755:
Fixes #982.

Various other fixes and improvements to CI.


.travis.yml
```diff
@@ -33,41 +33,50 @@ matrix:
...
+      docker exec oc octave --eval "pkg install https://octave-pythonic-downloads.s3.amazonaws.com/releases/octave-pythonic-0.0.1.tar.gz";
+  fi
+- if [ "x$PYTHONIC" = "xdev" ]; then
       docker exec oc octave --eval "pkg install https://gitlab.com/mtmiller/octave-pythonic/-/archive/master/octave-pythonic-master.tar.gz";
   fi
+- if [ "x$PYTHONIC" = "xno" ]; then
+      docker exec oc octave --eval "f=fopen('~/.octaverc','a');fprintf(f,'#empty\n');fclose(f)";
+  else
+      docker exec oc octave --eval "f=fopen('~/.octaverc','a');fprintf(f,'pkg load pythonic\n');fclose(f)";
+  fi
```
cbm755: @mtmiller can you help with this part?  Now that Pythonic is used by default, we only need to load the package and it will be used.

1.  Does modifying (creating) `.octaverc` seem like a reasonable way to do that?

2.  Surely it should be something like `docker exec oc echo "pkg load pythonic" >> ${HOME}/.octaverc`.  But its not bash, and that actually calls `/bin/echo`...  Should it be something like `docker exec oc bash ...`"?  I gave up and just used Octave.

mtmiller: Sure, I'll take a look

cbm755: eh?  That seemed to complain about `${HOME}` undefined or something when I did it...  Anyway looks much to me!

.travis.yml
```diff
@@ -32,42 +32,51 @@ matrix:
...
-- docker run --name=oc --detach --init --env=LC_ALL=C.UTF-8 --env=PYTHON_VERSION=${PYTHON_VERSION} --volume=$PWD:/octsympy:z mtmiller/octave:${OCT} sleep inf
+- docker run --name=oc --detach --init --env=LC_ALL=C.UTF-8 --env=PYTHON_VERSION=${PYTHON_VERSION} --env=PYTHON=python${PYTHON_VERSION} --volume=$PWD:/octsympy:z mtmiller/octave:${OCT} sleep inf
 
 install:
 - docker exec oc pip${PYTHON_VERSION} install sympy==${SYMPY_VER}
 - if [ "x$DOCTEST" = "xyes" ]; then
       docker exec oc octave --eval "pkg install -forge doctest";
   fi
 - if [ "x$PYTHONIC" = "xyes" ]; then
+      docker exec oc octave --eval "pkg install https://octave-pythonic-downloads.s3.amazonaws.com/releases/octave-pythonic-0.0.1.tar.gz";
```
mtmiller: We could pull the package version out into a variable, but probably not worth it

inst/private/python_ipc_driver.m
```diff
@@ -44,7 +44,10 @@
   end
 
   if (strcmp(lower(which_ipc), 'default'))
-    if (exist('popen2') > 1)
+    # TODO: may need to adjust, ideally just 'py'
```
mtmiller: Yeah, in the future it would be really interesting to see if OctSymPy can work with Matlab's built in Python support using the same interface

cbm755: Yes, that would be nice.  I'm a bit hesitant to do that kind of work myself.

inst/private/assert_pythonic_and_sympy.m
```diff
@@ -0,0 +1,160 @@
...
+    disp ('    Python environment than the one Pythonic is linked to?  Please try')
+    disp ('    without Pythonic as described above.')
+    return
+  end
+
+  spver = strtrim (char (output));
+
+  if (verbose)
+    disp ('')
+    fprintf ('SymPy must be at least version %s; you have version "%s".\n', ...
```
mtmiller: This output is slightly different from the original.

Without pythonic:

```
SymPy must be at least version 1.2; you have version 1.4.

Good, a working version of SymPy is installed.
```
With pythonic:

```
SymPy must be at least version 1.2; you have version "1.4".
Good, a working version of SymPy is installed.
```

PR #993, opened 2019-09-11
round: sympy 1.5 upstream behaviour changed to round-to-even

cbm755:
Not ideal for us.  Mark an xtest.  Document the change.  Link docs
to `roundb`, the equivalent double function.  Update copyright header.

### DIFF
--- a/inst/@sym/round.m
+++ b/inst/@sym/round.m
@@ -1,4 +1,6 @@
 %% Copyright (C) 2016 Lagu
+%% Copyright (C) 2016 Mike Miller
+%% Copyright (C) 2016, 2019 Colin B. Macdonald
 %%
 %% This file is part of OctSymPy.
 %%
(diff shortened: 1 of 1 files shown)


mtmiller:
Ugh. It seems ok for what it is, but I think people will probably report compatibility bugs as this rolls out. Can we work around it somehow? Something like the equivalent of

    sym(round(double(x)))

if `x` is a `Number`?

cbm755:
Thanks, good point: we don't have to use SymPy's `round`...  I'll think so more on it, but rolling our own may be complicated enough that I don't want to go there.

I don't want to go through double as I think there could be problems with high-precision Floats created by `vpa`.

cbm755:
```
>> x = vpa('1.5', 64) - vpa('1e-48', 64)
x = (sym) 1.4999999999999999999999999999999999999999999999990000000000000000
```
This number really is less than 1.5, it should (and does) round to 1:
```
>> round(x)
ans = (sym) 1
```
But `double` will convert it to the float `1.5` which then rounds up or down depending on the rounding system.
```
>> double(x)
ans = 1.5
>> round(ans)    # wrong result
ans = 2
```

mtmiller:
Hmm, the Matlab behavior is a little strange:

```
x(1) = vpa('1.5', 100) - vpa('1e-98', 100);
x(2) = sym('3/2');
x(3) = vpa('2.5', 100) - vpa('1e-98', 100);
x(4) = sym('5/2');
round(x)

ans = (2  2  2  3)
```

cbm755:
yech Matlab 2018b:
```
>> N = 100;
>> a  = vpa(sym(3)/2, N) - vpa('1e-36', N)
 a =
 1.5
 >> b  = vpa(sym(3)/2, N) - vpa('1e-48', N)
 b =
 1.5
 
>> round(a)
 ans =
 1
 
>> round(b)
ans =
2
``` 
Seems to be some special difference between `1e-39` and `1e-40`, somewhat independent of `N`.  Software: bah!  Its all terrible.

cbm755:
I'm going to merge this; I need to get us working on sympy 1.5 *NOW*; future improvements left for future...

PR #994, opened 2019-09-11
changes for newer sympy

cbm755:
Some changes to adjust for changes in upcoming SymPy releases.


cbm755:
build failures due to some octave.org time outs, merging

PR #1002, opened 2019-11-21
Changes for upcoming Octave 6

cbm755:
`format short` has changed and warnings are slightly different.


mtmiller:
LGTM, only failure is due to `sizemax` warning, already fixed in octave-pythonic master.

PR #1004, opened 2020-02-04
dsolve changes to make it more conforming to matlab return types

mass2010chromium:
Changes:
- `dsolve` returns the rhs of the resulting equation if it is a single equation with a function as lhs.
- Return values for implicit solutions unchanged. (Matches matlab)
- For a system of equations, returns a struct with fields labelled by the variable names, and values as the rhs of their corresponding equations. (Matches matlab)
- Updated test cases to remove calls to `rhs()` and index structs by the names of the variables.

Notable things that weren't changed:
- Return type is still [solution, classification], versus matlab's [solution1, solution2, ...]

### DIFF
--- a/inst/@sym/dsolve.m
+++ b/inst/@sym/dsolve.m
@@ -39,8 +39,8 @@
 %% @group
 %% sol = dsolve (DE)
 %%   @result{} sol = (sym)
-%%     4⋅x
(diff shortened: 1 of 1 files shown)


cbm755:
These commits were merged in #1168.  Thanks for your contributions!

PR #1011, opened 2020-03-11
On Windows, default to "python" not "python3"

cbm755:
It seems Anaconda and the official Python.org installer only provide
"python" and not "python3"

### DIFF
--- a/inst/private/defaultpython.m
+++ b/inst/private/defaultpython.m
@@ -32,6 +32,12 @@
 
 function python = defaultpython ()
 
-  python = 'python3';
+  if (ispc () && (~isunix ()))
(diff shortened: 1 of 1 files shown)


cbm755:
@mtmiller thoughts?

mtmiller:
_Sigh_ Yeah.

I'm tempted to want to write this as

```
  python = 'python3';

  if (ispc () && (~isunix ()))
    % 2020-03: Python.org installer/Anaconda do not provide python3
    python = 'python';
  end
```

to make it clear that Windows is the big exception, and `python3` should be considered normal.

cbm755:
I agree, will do.  I'm inclined to do a 2.9.1 with this change (and anything else the crops up in the next 24 hours or so).

PR #1016, opened 2020-04-01
close and reset Python subprocess state on clear

### LINKED ISSUE #1015: close Python subprocess on `clear all` or `clear functions`
Use `clear all` or `clear functions` after creating some symbolic variables, and the persistent function state is reset, but the pipe and subprocess are left open.

First reported on the Octave bug tracker [here](https://savannah.gnu.org/bugs/?58070).

mtmiller:
Move popen2 reset logic to a new helper function, and call it using a
persistent onCleanup object.

Fixes #1015.

### DIFF
--- a/inst/private/python_ipc_popen2.m
+++ b/inst/private/python_ipc_popen2.m
@@ -34,7 +34,7 @@
 
 function [A, info] = python_ipc_popen2(what, cmd, varargin)
 
-  persistent fin fout pid
+  persistent cleanup fin fout pid
 
   py_startup_timeout = 30;  % seconds
 
@@ -43,24 +43,11 @@ function [A, info] = python_ipc_popen2(what, cmd, varargin)
   info = [];
 
   if (strcmp(what, 'reset'))
-    if (~isempty(pid))
-      if (verbose)
-        disp ('Closing the Python communications link.')
-      end
-    end
-    if (~isempty(fin))
-      % produces a single newline char: not sure why
-      t = fclose(fin); fin = [];
-      waitpid(pid);
-      pid = [];
-      A = (t == 0);
-    else
-      A = true;
-    end
-    if (~isempty(fout))
-      t = fclose(fout); fout = [];
-      A = A && (t == 0);
-    end
+    cleanup = [];
+    fin = [];
+    fout = [];
+    pid = [];
+    A = true;
     return
   end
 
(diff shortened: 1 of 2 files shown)


inst/private/python_ipc_popen2.m
```diff
@@ -43,24 +43,11 @@
...
-      pid = [];
-      A = (t == 0);
-    else
-      A = true;
-    end
-    if (~isempty(fout))
-      t = fclose(fout); fout = [];
-      A = A && (t == 0);
-    end
+    cleanup = [];
```
mtmiller, line 46: This assignment triggers a call to the new reset helper function

inst/private/python_ipc_popen2.m
```diff
@@ -43,24 +43,11 @@
...
+    fin = [];
+    fout = [];
+    pid = [];
+    A = true;
```
mtmiller, line 50: Unfortunately I can't figure out a way to both use `onCleanup` and capture the return value

inst/private/python_ipc_popen2_reset.m
```diff
@@ -0,0 +1,51 @@
...
+%% If not, see <http://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @deftypefun {@var{A} =} python_ipc_popen2_reset (@var{fin}, @var{fout}, @var{pid})
+%% Private helper function for Python IPC.
+%%
+%% @var{A} is the resulting object, which might be an error code.
+%% @end deftypefun
+
+function A = python_ipc_popen2_reset (fin, fout, pid)
```
mtmiller, line 26: Maybe no point in returning any status here now?

inst/private/python_ipc_popen2_reset.m
```diff
@@ -0,0 +1,51 @@
+%% Copyright (C) 2020 Mike Miller
```
mtmiller, line 1: This file was copied from `python_ipc_popen2.m`, do you think me should propagate the original Copyright here?

inst/private/python_ipc_popen2_reset.m
cbm755: I think I agree.  I'm merging this to a local branch so I can try it on CI as well as try removing these return vlaues.

PR #1022, opened 2020-05-18
subsasgn: fix growing a scalar equality to matrix

### LINKED ISSUE #1021: elementwise equality assignment in octave fails when going from size 1->2
crossposted as an Octave bug:
https://savannah.gnu.org/bugs/?58366

the following produces a Python error in Octave, but runs fine with Matlab symbolic toolbox:

```
>> pkg load symbolic
>> syms a b c d
Symbolic pkg v2.9.0: Python communication link active, SymPy v1.3.
>> e = [a == b, c == d]
e = (sym) [a = b  c = d]  (1x2 matrix)
>> f(1) = a==b
f = (sym) a = b
>> f(2) = c==d
error: Python exception: TypeError: cannot determine truth value of Relational
    occurred at line 7 of the Python code block:
    if not A or not A.is_Matrix:
error: called from
    pycall_sympy__ at line 178 column 7
    mat_rclist_asgn at line 78 column 5
    mat_replace at line 144 column 5
    subsasgn at line 101 column 13
>> g(2) = c==d
g = (sym) [0  c = d]  (1x2 matrix)
>> g(1) = a==b
g = (sym) [a = b  c = d]  (1x2 matrix)
>> h(2) = c==d
h = (sym) [0  c = d]  (1x2 matrix)
>> h(1) = a==b
h = (sym) [a = b  c = d]  (1x2 matrix)
>> h(3) = a==c
h = (sym) [a = b  c = d  a = c]  (1x3 matrix)
>> i = a
i = (sym) a
>> i(2) = b
i = (sym) [a  b]  (1x2 matrix)
```

seems to be a problem going from size 1->2, but not 2->3, and just occurs with equalities not a variable array.

(Octave 5.2.0, symbolic 2.9.0, Python 3.7 (nonbundled))

mtmiller:
Fix test for LHS in indexed matrix assignment. Allow growing a scalar equality or inequality expression into a matrix. Fixes #1021.

### DIFF
--- a/inst/@sym/private/mat_rclist_asgn.m
+++ b/inst/@sym/private/mat_rclist_asgn.m
@@ -61,7 +61,7 @@ function z = mat_rclist_asgn(A, r, c, B)
           '    B = sp.Matrix([[B]])'
           'BT = B.T'
           '# copy of A, expanded and padded with zeros'
-          'if not A or not A.is_Matrix:'
+          'if A is None or not A.is_Matrix:'
           '    n = max( max(r)+1, 1 )'
           '    m = max( max(c)+1, 1 )'
           'else:'
--- a/inst/@sym/subsasgn.m
+++ b/inst/@sym/subsasgn.m
@@ -425,6 +425,13 @@ end
 %! e(1:2) = true;
 %! assert (isequal (e, [sym(1)==1  sym(2)==2  x==10  sym(3)==4]))
 
+%!test
+%! % grow scalar equality expression into a matrix of equalities
+%! syms a b c d
+%! e = a == b;
+%! e(2) = c == d;
+%! assert (isequal (e, [a==b  c==d]))
+
 
 %% 2D arrays from mat_mask_asgn
 


mtmiller:
Looks like this change breaks two existing tests in subsasgn.m, may need some help resolving all of these different indexing cases.

feralber:
@mtmiller, Try to apply this patch. It passes the tests

```
--- a/inst/@sym/private/mat_rclist_asgn.m
+++ b/inst/@sym/private/mat_rclist_asgn.m
@@ -61,7 +61,7 @@ function z = mat_rclist_asgn(A, r, c, B)
           '    B = sp.Matrix([[B]])'
           'BT = B.T'
           '# copy of A, expanded and padded with zeros'
-          'if not A or not A.is_Matrix:'
+          'if not isinstance(A,Matrix):'
           '    n = max( max(r)+1, 1 )'
           '    m = max( max(c)+1, 1 )'
           'else:'
```

cbm755:
Thanks @feralber, it might also be `if not isintance(A, MatrixBase)`.  I will merge this and experiment some more in a new PR.

Issue #1026, opened 2020-06-29
Error when assigning an empty vector to a vector/matrix block


feralber:
An error is issued when assigning an empty vector to a vector subblock:
```
>> a=sym(1:5);
>> a(1:3)=[]
error: Python exception: TypeError: cannot add <class 'sympy.matrices.dense.MutableDenseMatrix'> and <class 'int'>
    occurred at line 1 of the Python code block:
    _ins[0].col_del(_ins[1] - 1); return _ins[0],
error: called from
    pycall_sympy__ at line 178 column 7
    mat_replace at line 56 column 17
    subsasgn at line 101 column 13
```
An error is also issued when assigning an empty matrix to a set of matrix columns/rows:
```
>> A=sym(eye(4));
>> A(1:2,:)=[]
error: Python exception: TypeError: cannot add <class 'sympy.matrices.dense.MutableDenseMatrix'> and <class 'int'>
    occurred at line 1 of the Python code block:
    _ins[0].row_del(_ins[1] - 1); return _ins[0],
error: called from
    pycall_sympy__ at line 178 column 7
    mat_replace at line 75 column 13
    subsasgn at line 101 column 13
```

Issue #964, opened 2019-04-16
emptying a scalar by assigning to its first entry fails


cbm755:
```
a = sym(10)
a(1) = []
```
a should be empty but instead we get an error

PR #1027, opened 2020-06-29
Fix errors when assigning empty vector to a vector/matrix block

feralber:
This should fix issues #1026 and #964


cbm755:
Merging to a side branch `asgn_empty` to add tests and whatnot.

Thanks for this!

PR #1028, opened 2020-07-18
fix: compatible with sympy 1.6(#1023)

### LINKED ISSUE #1023: symbolic not working with sympy 1.6
Symbolic is not compatible with sympy 1.6

```
>> pkg load symbolic
>> syms x
Symbolic pkg v2.9.0: /usr/local/lib/python3.8/dist-packages/sympy/__init__.py:672: SymPyDeprecationWarning:

importing sympy.core.compatibility with 'from sympy import *' has been
deprecated since SymPy 1.6. Use import sympy.core.compatibility
instead. See https://github.com/sympy/sympy/issues/18245for more
info.

  self.Warn(
  File "<stdin>", line 12, in octoutput_drv
  File "<stdin>", line 54, in octoutput
  File "<stdin>", line 55, in octoutput
  File "/usr/local/lib/python3.8/dist-packages/sympy/__init__.py", line 677, in __getattr__
    return getattr(self.mod, name)
AttributeError: module 'sympy.core.compatibility' has noattribute 'integer_types'
Waiting...............warning: readblock: timeout of 30 exceeded, breaking out
warning: called from
    readblock at line 101 column 7
    python_ipc_popen2 at line 99 column 16
    python_ipc_driver at line 62 column 15
    pycall_sympy__ at line 163 column 11
    valid_sym_assumptions at line 38 column 10
    assumptions at line 82 column 7
    syms at line 97 column 13
no output so far

error: ipc_popen2: something wrong? timed out starting python
error: called from
    python_ipc_popen2 at line 101 column 7
    python_ipc_driver at line 62 column 15
    pycall_sympy__ at line 163 column 11
    valid_sym_assumptions at line 38 column 10
    assumptions at line 82 column 7
    syms at line 97 column 13
```
Downgrade to sympy-1.5 with pip3 and it works again.

OS Ubuntu 20.04
Octave 5.2.0
symbolic 2.9.0 (as shown in the output above)
sympy 1.6 installed with pip

billchenchina:
Hi, this PR fixes #1023.

However, I'm not quite sure how to edit `inst/private/check_and_convert.m` to support python2. Help needed!

### DIFF
--- a/inst/private/python_header.py
+++ b/inst/private/python_header.py
@@ -184,7 +184,8 @@ try:
             c = ET.SubElement(et, "list")
             for y in x:
                 octoutput(y, c)
-        elif isinstance(x, sp.compatibility.integer_types):
+        elif isinstance(x, int) or \
+             (sys.version_info[0] <= 2 and isinstance(x, long)):
             a = ET.SubElement(et, "item")
             f = ET.SubElement(a, "f")
             f.text = str(OCTCODE_INT)
@@ -205,7 +206,8 @@ try:
             f.text = d2hex(x.real)
             f = ET.SubElement(a, "f")
             f.text = d2hex(x.imag)
-        elif isinstance(x, sp.compatibility.string_types):
+        elif isinstance(x, str) or \
+             (sys.version_info[0] <= 2 and isinstance(x, unicode)):
             a = ET.SubElement(et, "item")
             f = ET.SubElement(a, "f")
             f.text = str(OCTCODE_STR)
--- a/inst/private/check_and_convert.m
+++ b/inst/private/check_and_convert.m
@@ -30,8 +30,8 @@ function obj = check_and_convert(var_pyobj)
 
     sp = py.sympy;
     _sym = py.tuple({sp.Basic, sp.MatrixBase});
-    string_types = sp.compatibility.string_types;
-    integer_types = sp.compatibility.integer_types;
+    string_types = py.str;
+    integer_types = py.int;
   end
 
 


inst/private/python_header.py
```diff
@@ -205,7 +206,8 @@ def octoutput(x, et):
             f.text = d2hex(x.real)
             f = ET.SubElement(a, "f")
             f.text = d2hex(x.imag)
-        elif isinstance(x, sp.compatibility.string_types):
+        elif isinstance(x, str) or \
+             (sys.version_info[0] <= 2 and isinstance(x, unicode)):
```
asmeurer, line 210: As of 1.6, SymPy no longer itself supports Python 2, so there should be no need to use Python 2 compatibility idioms in this codebase, unless you are still supporting SymPy 1.5 and Python 2.

cbm755: thanks.  I kind of am, b/c my windows package relies on something (py2exe) that doesn't work on Py3.

cbm755:
Thanks!

I've been ignoring this project for too long: must clean these up and do a new release.  Will look later

inst/private/python_header.py
asmeurer: The py2exe on PyPI apparently supports Python 3 https://pypi.org/project/py2exe/

cbm755: Last I checked I was held up by https://github.com/manthey/pyexe/issues/25

pkarakal: May I suggest switching to pyinstaller? It can also produce a single exe binary which packs all the needed dlls. It is also actively developed unlike pyexe.

mmuetzel: @cbm755: The Windows installer for Octave 6.x (built with MXE Octave) bundles a minimal version of python 3.8, sympy 1.4 and the symbolic package 2.9.0. So, it should no longer be necessary to provide a separate Windows installer for the symbolic package. (At least for newer versions of Octave.)
It would probably be possible to change to different versions of python or sympy in MXE Octave if that simplifies an update of the symbolic package.
If this is holding back getting the symbolic package to work with newer versions of sympy, you could probably just ignore backward compatibility with python2 (for Windows).

Artturin:
@cbm755 Hi, its later, mind taking a look?

Issue #952, opened 2019-03-26
docs: functions with arguments that switch order in SymPy 


cbm755:
- [x] gammainc
- [x] igamma
- [ ] lambertw
- [ ] others?

mtmiller:
I added `lambertw` to your list (does GitHub notify you if I edit your post?).

`gammainc` and `igamma` are switched order with respect to each other, and that's already documented in `@sym/igamma.m`. `igamma(n, x)` is the same as Python `uppergamma(n, x)`.

cbm755:
Thanks!

> notify you if I edit your post?

apparently not.

cbm755:
igamma and gammainc are not really the issue here: they are backwards in Matlab's SMT too.  I did find and fix other bugs in gammainc in #956.

PR #1051, opened 2021-10-16
fix of #1023 with changes from #1028

jmfrank63:
One year no changes, then two PR just 20 minutes apart ;-) I've seen #1028 was just a little faster. I put this one in, nevertheless, as it contains the fix for the argument reverse of #952 which also seems to be broken.


jmfrank63:
I have tested importing symbols with sympy 1.9 and octave 5.2.0.

PR #1061, opened 2022-03-04
Fix for DeprecationWarning on newer python

chrisjohgorman:
The following warning is generated when calls are made to distutils

octave:3> syms a
Symbolic pkg v2.9.0+: <stdin>:23: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives

This changes the call from distutils to packaging.  Shamelessly borrowed
from gnuradio.


chrisjohgorman:
Hi Collin,

On my system I am using python3.10 which causes a warning when I import distutils.  This patch changes the call to import from packaging instead.  This is my first pull request, so I'm not certain I got it all right, but this patch gets rid of the warning on my system.

Thanks.

Chris

cbm755:
This looks good.  Note that `packaging` is an [external library](https://pypi.org/project/packaging/) so we'll need to consider that.

chrisjohgorman:
Hi Colin,

I didn't consider the problem packaging being an external library.
How do you want to handle that?  I can write a try block to see if
packaging loads and if it fails install packaging via pip.  Do you
want something like that?  I found a stack overflow link where someone
wanted to do just that [1].

So something like...

import subprocess
import pkg_resources

try:
     from packaging.version import Version as LooseVersion
except
subprocess.check_call([python, '-m', 'pip', 'install', *missing],
stdout=subprocess.DEVNULL)


[1] https://stackoverflow.com/questions/44210656/how-to-check-if-a-module-is-installed-in-python-and-if-not-install-it-within-t

chrisjohgorman:
Sorry Colin.  I'm fighting with gmail.

Something like ...

import subprocess
import pkg_resources

try:
     from packaging.version import Version as LooseVersion
except
     subprocess.check_call([sys.executable, '-m', 'pip', 'install',
'packaging'], stdout=subprocess.DEVNULL)
     from packaging.version import Version as LooseVersion

Is there a better way?

Chris

cbm755:
Perhaps we can just document that it is required, in the same sense that `sympy` is required.

I.e., folks will have to do `pip install packaging sympy`

(I'm not too keen on having programs run `pip` themselves...  like maybe someone uses other package managers.)

chrisjohgorman:
Sounds great.  I was a little worried about non networked machines and pip
installing something without the user knowing, etc.

I have another warning due to deprecation with a newer sympy.  I'll do
another push.  The offending code is again in inst/private/python_header.p
and is 'from sympy.core.compatibility import unicode'.  Simply commenting
the code fixes the warning although I am not sure of the consequences of
this.

Take care.

Chris

octave:1> pkg load symbolic
octave:2> syms x y z a b c k;
Symbolic pkg v2.9.0+: <stdin>:1: SymPyDeprecationWarning:

The sympy.core.compatibility submodule is deprecated.

This module was only ever intended for internal use. Some of the functions
that were in this module are available from the top-level SymPy namespace,
i.e.,

    from sympy import ordered, default_sort_key

The remaining were only intended for internal SymPy use and should not be
used
by user code.

See
https://docs.sympy.org/latest/explanation/active-deprecations.html#deprecated-sympy-core-compatibility
for details.

This has been deprecated since SymPy version 1.10. It
will be removed in a future version of SymPy.

ImportError: cannot import name 'unicode' from 'sympy.core.compatibility'
(/usr/lib/python3.10/site-packages/sympy/core/compatibility.py)
Python communication link active, SymPy v1.10.1.

cbm755:
This will cause the new Github Actions CI to fail.  I'm not sure why they don't show up here (probably b/c I don't have it configured properly for PR) but if you go to your fork:

https://github.com/chrisjohgorman/octsympy

And click "Actions", you should see lots of failing tests.

For now, I suggest putting `pip install packaging` somewhere in `.github/workflows/main.yml`.

chrisjohgorman:
Hi Collin,

Just to update you.  I've put two pip installs into the main.yml and
am trying to test them out.  Currently I'm in a holding pattern
'waiting for a runner' to pick my job up.  I'll let you know when I
have made some progress.

Chris

chrisjohgorman:
Hi again Colin,

Here's my patch.  I can't seem to get a runner to execute with my
changes.  I believe I would have to set up a self hosted runner to
test this.  (Sorry for misspelling your name earlier.)

Chris


Chris

cbm755:
Probably its not your fault: I am still learning Github Actions.

chrisjohgorman:
Oh no, this was my fault.  I have succeeded in getting github actions to
work now.  I just can't seem to get the try_stuff job to work.  I'll play
with it and see if I can get it to go.  Otherwise, I have successful runs
from bits and doctests jobs.

Chris

cbm755:
When you get a chance, can you also merge master into this?  I'm curious if a recent change will make the CI run happen here on this PR.

.github/workflows/main.yml
```diff
@@ -44,9 +44,9 @@ jobs:
           docker run --rm gnuoctave/octave:6.2.0 octave-cli --eval 'disp(42)'
           docker run --rm -v $PWD:/home/jovyan:rw gnuoctave/octave:6.2.0 octave-cli --eval 'pwd; ls; disp(42); help sympref'
           docker run --rm -v $PWD:/home/jovyan:rw gnuoctave/octave:6.2.0 octave-cli --eval "sympref diagnose"
-          docker run --rm -v $PWD:/home/jovyan:rw gnuoctave/octave:6.2.0 octave-cli --eval "x=sym('x'); y=sin(x/17); disp(y)"
+          #docker run --rm -v $PWD:/home/jovyan:rw gnuoctave/octave:6.2.0 octave-cli --eval "x=sym('x'); y=sin(x/17); disp(y)"
           echo "Try a test"
-          docker run --rm -v $PWD:/home/jovyan:rw gnuoctave/octave:6.2.0 octave-cli --eval "x=sym('x'); test @sym/qr"
+          #docker run --rm -v $PWD:/home/jovyan:rw gnuoctave/octave:6.2.0 octave-cli --eval "x=sym('x'); test @sym/qr"
```
cbm755, line 49: The issue here is that `docker run --rm` is stateless: there is no change to change the image before `octave-cli` is run.

I suggest ignoring it for now; I'll file a new issue to discuss.

chrisjohgorman:
Sounds good.  I will leave it for the time being.  Thanks for the help.

Chris

cbm755:
See comments in #1083, let's fallback on the deprecated `distuils`.  Then we can merge this and the CI jobs will all pass.

cbm755:
Looks good!  For next time, try using a local branch (rather than master) on your fork.

cbm755:
FYI: I squashed all these commits to a single 4580d8be2393f9eb0d1044783f9c2edf43f8039c

chrisjohgorman:
Hi Colin,

I didn't consider the problem packaging being an external library.  How do
you want to handle that?  I can write a try block to see if packaging loads
and if it fails install packaging via pip.  Do you want something like
that?  I found a stack overflow link where someone wanted to do just that
[1].

so

import subprocess
import pkg_resources

try:
from packaging.version import Version as LooseVersion






[1]
https://stackoverflow.com/questions/44210656/how-to-check-if-a-module-is-installed-in-python-and-if-not-install-it-within-t

PR #1071, opened 2022-03-29
close and reset Python subprocess state on clear [updated]

cbm755:
Following up on #1016.


cbm755:
I still got this:
```
test sympref
warning: test: file /home/cbm/work/octsympy/octsympy.git/inst/sympref.m leaked file descriptors
```

Issue #1093, opened 2022-04-03
doctest: update Catalan printing for SymPy 1.10


cbm755:
In sympy 1.10.1, doctests are failing because of changes to how Catalan pretty prints.

- `pip install sympy==1.10.1`
- `pkg install -forge doctest`
- `pkg load doctest`
- `doctest catalan`, maybe other places too.

cbm755:
@chrisjohgorman next time put:

`Fixes #1093` somewhere in the BODY (3rd line onward, not first line) of your git commit message and this would close automatically when #1097 was merged.

No big deal, I'm closing it manually now.

chrisjohgorman:
Thanks, will do.

PR #1097, opened 2022-04-06
doctest: update to Catalan printing for Sympy 1.10 issue #1093

chrisjohgorman:
- Added copyright
- Fix expected return string from Sympy 1.10+ so doctest will pass
- note this change will make doctest fail for sympy < 1.10+

### DIFF
--- a/inst/@sym/isconstant.m
+++ b/inst/@sym/isconstant.m
@@ -1,4 +1,5 @@
 %% Copyright (C) 2014-2016, 2019 Colin B. Macdonald
+%% Copyright (C) 2022 Chris Gorman
 %%
 %% This file is part of OctSymPy
 %%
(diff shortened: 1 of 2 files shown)


inst/catalan.m
```diff
@@ -25,7 +26,7 @@
 %% @example
 %% @group
 %% catalan ()
-%%   @result{} (sym) Catalan
+%%   @result{} (sym) G
```
cbm755, line 29: you can workaround this failing use `doctest: +SKIP_UNLESS...`, see `@sym/harmonic` for example

cbm755 (commented):
Looks good so far, see comments.

I think there also might another file with this Catalan printing in the docs.

inst/@sym/isconstant.m
```diff
@@ -24,12 +25,20 @@
...
 %%       ⎡x   1      π   ⎤
 %%       ⎢               ⎥
 %%       ⎣2  2⋅y  Catalan⎦
+%% @c doctest: +SKIP_UNLESS(pycall_sympy__ ('return Version(spver) >= Version("1.10")'))
+%% syms x y
+%% A = [x 1 pi; 2 2*y catalan()]
+%%   @result{} A = (sym 2×3 matrix)
+%%      ⎡x   1   π⎤
+%%      ⎢         ⎥
+%%      ⎣2  2⋅y  G⎦
```
cbm755, line 41: yes right idea, but this `@c doctest:` is not a conditional include.

The primary point of the docs is to be human-readable (`help @sym/constant`).

I think you should just have the `>=` case.  That is, in most cases, we assume the reader is using the latest SymPy.

(This is somewhat dealt with on [the wiki](https://github.com/cbm755/octsympy/wiki/Development-Guide), but probably not clearly enough.)

chrisjohgorman:
I think I get what you mean.  I will redo this this evening.

chrisjohgorman:
I have updated catalan.m and @sym/isconstant.m removing the duplicate doctests.  These three changes together do the following

- update copyright
- add @c doctest: +SKIP_UNLESS call to catalan.m disabling the test for sympy less than 3.10
- update result for catalan() call
- add two @c doctest: +SKIP_UNLESS calls to @sym/isconstant.m disabling the test for sympy less than 3.10
- update result for A = [x 1 pi; 2 2*y catalan()] call

cbm755:
I still don't have our CI working properly (#1098) but this looks good to me!

PR #1101, opened 2022-04-08
unique: Fix #1047 and add test case accordingly.

### LINKED ISSUE #1047: octave symbolic unique
syms a;

unique([a,a])  works fine, but

unique([a])

leads to the following error:
error: Python exception: UnboundLocalError: local variable ‘s’ referenced before assignment
occurred at line 1 of the Python code block:
return sp.Matrix([list(uniq(*ins))]),
error: called from
pycall_sympy_ at line 178 column 7
unique at line 55 column 5

alexvong1995:
This should fix #1047. We also add a test case accordingly.

### DIFF
--- a/inst/@sym/unique.m
+++ b/inst/@sym/unique.m
@@ -1,5 +1,6 @@
 %% Copyright (C) 2016, 2019 Colin B. Macdonald
 %% Copyright (C) 2016 Lagu
+%% Copyright (C) 2022 Alex Vong
 %%
 %% This file is part of OctSymPy.
 %%
@@ -52,7 +53,13 @@ function r = unique(A)
     print_usage ();
   end
 
-  r = pycall_sympy__ ('return sp.Matrix([list(uniq(*_ins))]),', A);
+  cmd = {'(A,) = _ins'
+         'if A.is_Matrix:'
+         '    return sp.Matrix([list(uniq(*_ins))]),'
+         'else:'
+         '    return _ins'};
+
+  r = pycall_sympy__ (cmd, A);
 
 end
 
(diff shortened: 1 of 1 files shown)


inst/@sym/unique.m
```diff
@@ -67,3 +72,8 @@
 %! A = [1 2 3 3 4 5 5 6 7 7 x x y y];
 %! B = [1  2  3  4  5  6  7  x  y];
 %! assert (isequal (unique(A), B))
+
+%!test
+%! syms x
+%! A = [x]
```
cbm755, line 78: Maybe this is just `A = x`?

inst/@sym/unique.m
```diff
@@ -47,12 +47,17 @@
 
 
 function r = unique(A)
+  cmd = {'(A,) = _ins'
+         'if A.is_Matrix:'
+         '    return sp.Matrix([list(uniq(*_ins))]),'
+         'else:'
+         '    return _ins'};
```
cbm755, line 54: I think we normally put this `cmd` after the `nargin` input processing (but could be wrong!)

cbm755 (commented):
Thanks for the fix!  Please add your name and the year to the copyright header, decide if the above comments should be addressed, and then we'll get this merged!

inst/@sym/unique.m
alexvong1995: I've checked other files in the same directory. This is indeed the convention.

inst/@sym/unique.m
alexvong1995: Yep and thus it is actually not needed. We can just use `x`...

PR #1102, opened 2022-04-08
Update documentation for @sym/heaviside.m

### LINKED ISSUE #1091: update docs for heaviside changes in SymPy 1.9
`heaviside(sym(0))` is now 1/2---yah!

How to get started:
- See `help @sym/heaviside` and `doctest @sym/heaviside`.
- `pip install sympy==1.9` or `pip install sympy==1.10.1`.
- `pkg install -forge doctest`

chrisjohgorman:
- rewrote examples 2-5 in the documentation, noting new default value at
  sym(0), as well as how to override the definition at 0 for both right
  and left continuous values.
- also noted that the function can be used on a matrix and is called for
  each element.
- Fixes #1091

### DIFF
--- a/inst/@sym/heaviside.m
+++ b/inst/@sym/heaviside.m
@@ -1,4 +1,5 @@
 %% Copyright (C) 2014-2016, 2018-2019 Colin B. Macdonald
+%% Copyright (C) 2022 Chris Gorman
 %%
 %% This file is part of OctSymPy.
 %%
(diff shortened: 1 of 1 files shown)


inst/@sym/heaviside.m
```diff
@@ -32,34 +33,41 @@
 %% @end group
 %% @end example
 %%
-%% By default, the value at zero is undefined:
+%% There are various conventions for 'heaviside(sym(0))'; this function
+%% returns (sym) 1/2 by default.
```
cbm755, line 37: I think you should just say "returns 1/2 by default" or maybe "returns the midpoint by default:" (and let the example show it is 1/2).

Other than that, this looks ready to go1

Issue #1079, opened 2022-03-29
A=C/B solving for A in A*B = C is broken on SymPy > 1.5.1


cbm755:
These sorts of matrix operations seem to work in SymPy 1.5.1.

TODO: Investigate, report upstream if necessary.  In short term, `xtest` will rescue our CI runs.

chrisjohgorman:
I did a git bisect and found the problem, it lies in one of the decorator functions for `__rtruediv__` in `sympy/core/expr.py.`  I'm not sure how to fix yet perhaps someone with more experience with sympy could have a look to see what they would propose for the fix.  In short, with decorator function commented out, our test passes.  I've attached a patch.

[octsympy-1079.patch.txt](https://github.com/cbm755/octsympy/files/8474830/octsympy-1079.patch.txt)

chrisjohgorman:
It appears that right division is not working with sympy.  I have opened an issue on this. [Sympy issue 23356](https://github.com/sympy/sympy/issues/23356).

cbm755:
And the conclusion seems to be its not regression, was never supposed to work in the first place.  So we'll need to support it here.

In the [upstream issue](https://github.com/sympy/sympy/issues/23356#issuecomment-1098448658) there are some hints about how we might do this.

We'll probably need to hack `mldivide.m` and `mrdivide.m`

cbm755:
I think this is fixed.

PR #1114, opened 2022-04-16
Allow for A = C / B when C is generated by A * B = C

chrisjohgorman:
- remove doctest SKIP_IF sympy > 1.5.1
- replace xtests with test and remove reference to issue
- use solve command that oscarbenjamin suggested from upstream to fix
  issue.
- fixes issue #1079
- please use extra care with this fix as I am sick and may have made a
  mistake.

### DIFF
--- a/inst/@sym/mrdivide.m
+++ b/inst/@sym/mrdivide.m
@@ -1,4 +1,5 @@
 %% Copyright (C) 2014, 2016, 2018-2019, 2022 Colin B. Macdonald
+%% Copyright (C) 2022 Chris Gorman
 %%
 %% This file is part of OctSymPy.
 %%
@@ -41,9 +42,7 @@
 %% @end example
 %%
 %% The forward slash notation can be used to solve systems
-%% of the form A⋅B = C using @code{A = C / B}.
-%% Unfortunately, this is currently broken
-%% (see @url{https://github.com/cbm755/octsympy/issues/1079}):
+%% of the form A⋅B = C using @code{A = C / B}:
 %% @example
 %% @group
 %% B = sym([1 0; 1 2]);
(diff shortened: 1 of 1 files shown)


inst/@sym/mrdivide.m
```diff
@@ -80,17 +77,21 @@
...
-  % not working as of 2016-01).  We don't disallow this but its not a
-  % good thing to encourage for working in Octave (since it won't work
-  % with doubles).
+  z = pycall_sympy__ (cmd, sym(x), sym(y));
 
 end
 
 
-%!test
+%!xtest
```
cbm755, line 94: this doesn't look right?

cbm755 (commented):
Looks great!  There was one test -> xtest change that I didn't understand, maybe a typo?

Do we need to do something like this for mldivide too?  (could be a different PR or maybe you can file an issue so we don't forget)

chrisjohgorman:
I thank you for catching my mistake.  I have pushed the changes and
expected to have a chance to comment when issuing a pull request, but I
think the pull request was automatically updated.

I don't think we need to make changes to ***@***.***/mldivide.m, but I will
open an issue just in case.

chrisjohgorman:
I just found a use case that will cause this to fail.  If B is
non-invertible, then the division will fail.

A = [sym(-1) sym(3)/2; sym(1) sym(-1)]
B = [sym(-1) sym(3)/2; sym(2)/3 sym(-1)]
C = A*B
A2 = C/B

I don't think there is a solution to this, and it fails with the old
code and sympy-1.5.1 too.  FYI.

cbm755:
In theory some kind of minimal norm solution would be best in non-invertible situations but I suspect we should just do whatever upstream solve does for now.  Maybe you could file another issue with your non-invertible exam and we can consider there (?)

For now, I will merge this.

PR #1116, opened 2022-04-23
Changes octave pythonic interface 

chrisjohgorman:
@cbm755 I believe I have made a git mess here.  Sorry.  If I'm spamming you with pull requests, please let me know.  My intention was to address issue 1115, on the pythonic interface failing to load properly, and I have run into issue 1079.

So I will recap, left to do for 1079 is to create an issue so we don't forget that changes may be needed for `mldivide.m` and `mrdivide.m`.  I will do this after issuing this PR.  At the very worst we will have two issues on the same topic.  If the code needs to be changed before merging into master please let me know.

For 1115, the edit is to `check_and_convert.m`, and a call to py.type(py.str) is needed to allow the creation of a symbolic variable.

### DIFF
--- a/inst/private/check_and_convert.m
+++ b/inst/private/check_and_convert.m
@@ -46,7 +46,7 @@ function obj = check_and_convert(var_pyobj)
       obj{i} = get_sym_from_python(x);
     elseif (py.isinstance(x, list_or_tuple))
       obj{i} = check_and_convert(x);
-    elseif (py.isinstance (x, py.type(py.str())))
+    elseif (isa (x, 'py.str'))
       obj{i} = char (x);
     elseif (py.isinstance(x, builtins.dict))
       make_str_keys = pyeval ('lambda x: {str(k): v for k, v in x.items()}');
@@ -55,7 +55,7 @@ function obj = check_and_convert(var_pyobj)
       % make sure values are converted to sym
       s = structfun (@(t) check_and_convert (t){:}, s, 'UniformOutput', false);
       obj{i} = s;
-    elseif (py.isinstance(x, py.type(py.int())))
+    elseif (isa (x, 'py.int'))
       if (py.isinstance(x, pyeval('bool')))
         error ('unexpected python bool')
       end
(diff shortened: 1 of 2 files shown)


cbm755:
> Sorry. If I'm spamming you with pull requests

Don't worry about this!  And sorry for not getting back to the previous one on this topic.

If you are worried or self-conscientious about it, one thing you can do is start the PR with "Draft:".  Then make all the mess you want! :)

chrisjohgorman:
Thanks.  I will remember the Draft: trick.  And no worries about not
getting back to me.  I can imagine the number of balls you're juggling
running octsympy.

inst/private/check_and_convert.m
```diff
@@ -46,7 +46,7 @@
       obj{i} = get_sym_from_python(x);
     elseif (py.isinstance(x, list_or_tuple))
       obj{i} = check_and_convert(x);
-    elseif (py.isinstance (x, py.str))
+    elseif (py.isinstance (x, py.type(py.str)))
```
cbm755, line 49: I can see that this works but it doesn't feel semantically correct to me (with my Python hat one).

How about `py.type(py.str(''))` or `py.type(py.str())` (better? if its well-defined).  That way, it would DTRT even if Pythonic were to distinguish (again?) between `py.str` and `py.str()`

chrisjohgorman:
Sounds good.  I also ran into the same error with py.int on line 58 while
running rho = sym('(1 + sqrt(5))/2').  I propose the same fix.  So, I'll
push a change.

inst/private/check_and_convert.m
mtmiller: How about `isa (x, 'py.str')`?

cbm755:
Thank you both, I'll merge this.

chrisjohgorman:
I pushed a new change set to use `isa (x, 'py.str')` and `isa (x,
'py.int')`.  Thanks for the help on this @mtmiller and @cbm755.  Just
waiting on the tests to pass.

Chris

PR #1121, opened 2022-05-03
ztrans: Add new function.

alexvong1995:
This PR includes the commit from PR #1120. We need it in order to pass the tests.


BobKruse:
Thanks Alex

cbm755:
Very cool!  This works better than I would've expected!

Very thorough test set.

The docs have a `@seealso` to a function that is not yet implemented.  I'll file another issue to comment that out before 3.0.0

alexvong1995:
Great! `iztrans` will be the next target. But I'm getting busy so it'll probably take longer. (Also, it's a harder problem to solve.)

Issue #1060, opened 2022-02-19
`sympref` test fails when in read-only directory


mmuetzel:
When the current directory has read-only access, the following BIST is failing:
https://github.com/cbm755/octsympy/blob/5b58530f4ada78c759829ae703a0e5d9832c32d4/inst/sympref.m#L424-L434

The error message is:
```
>>>>> processing [snip]\share\octave\packages\symbolic-2.9.0\sympref.m
***** test
 sympref('ipc', 'systmpfile');
 syms x
 delete('tmp_python_cmd.py')
!!!!! test failed
fprintf: invalid stream number = -1
shared variables     sympref_orig =

      scalar structure containing the fields:

        ipc = default
        display = ascii
        digits = 32
        quiet = 0

***** error <line 1> pycall_sympy__ ('raise ValueError')
!!!!! error failed.
Expected <line 1>, but got <fprintf: invalid stream number = -1>

***** error <line 1> pycall_sympy__ ('raise ValueError', struct('a', 1, 'b', 'word'))
!!!!! error failed.
Expected <line 1>, but got <fprintf: invalid stream number = -1>
```

That is the case if the symbolic package is installed as a "system" package at a read-only location and a user executes `pkg test symbolic`.

Would it be possible to have the temporary file be written to a location where the user should have write access? E.g., the `TMP` directory?

See also: https://savannah.gnu.org/bugs/index.php?59343

cbm755:
I have not yet been able to reproduce this but it makes sense and would be good to fix (so that we could use `pkg test symbolic` instead fo the hacky forked `octsympy_tests` that I have in place right now).

cbm755:
More background: `sympref ipc systmpfile` turns this mode on.  What we need here is to change that code (somewhere in `inst/private`) to make the tmp file somewhere other than `pwd`.   See the linked upstream bugs for hints.

cbm755:
@alexvong1995 are you able to reproduce this?

(I think I understand the problem but unsure about how to make it happen on my own computer)

cbm755:
Here's part of it:
```
> pkg load symbolic
> sympref ipc systmpfile
> cd /home   # or anywhere else without write perm
> syms x
> syms x
error: fprintf: invalid stream number = -1
error: called from
    python_ipc_system at line 111 column 5
    python_ipc_driver at line 73 column 15
    pycall_sympy__ at line 164 column 11
    sym at line 452 column 9
    syms at line 160 column 9
```
[...]

PR #1140, opened 2022-06-21
tempname to fix systmpfile ipc when cwd not writeable

cbm755:
Fixes Issue #1060.


cbm755:
Tests finally passed, so I'll merge this.

inst/private/python_ipc_system.m
```diff
@@ -106,14 +107,19 @@
...
+      if (verbose)
+        disp (['This session will use the temp file "' tmpfilename '"'])
+      end
+    end
+    fd = fopen (tmpfilename, 'w');
     fprintf(fd, '# temporary autogenerated code\n\n');
     % we just added two more lines at the top
     info.prelines = info.prelines + 2;
     fputs(fd, bigs);
     fclose(fd);
```
alexvong1995, line 121: I think there's a potential race condition here due to the use of `tempname`. In the best case, a race condition would merely cause inconvenience by creating nondeterministic test failure. In the worst case, it could be a security hole. 

Quoting the [python documentation](https://docs.python.org/3/library/tempfile.html),

```
A historical way to create temporary files was to first generate a file name with the mktemp() function and then create a file using this name. Unfortunately this is not secure, because a different process may create a file with this name in the time between the call to mktemp() and the subsequent attempt to create the file by the first process. The solution is to combine the two steps and create the file immediately. This approach is used by mkstemp() and the other functions described above.
```

Unfortunately, while octave do provide `tmpfile` and `mkstemp` to create temp file w/o race condition, matlab doesn't, unlike most modern programming languages.

I would suggest creating temp file using `mkstemp` in octave and use `tempfile.mkstemp` in python otherwise.

cbm755: yeah, I noticed that about Matlab, which is why I didn't do `mkstemp`.   We do not reliably have a Python connect at this point.

I think best we could do is Octave-dependent code, using the proper secure approach and then tempname on Matlab...   I haven't tested on Matlab in a long time so its not worth much effort.

File an issue for this for a future improvement?

cbm755: The other thing to keep in mind is that this `systmpfile` stuff is basically for debugging, not production use (although we do call it in our CI)

We could probably just deprecate this code, especially if/when Pythonic matures.

alexvong1995: Sounds good to me. So for now, we workaround this by using Octave-dependent code and deprecate this code in the future.

I believe we should also deprecate code using `system` in the future as well. It is well-known that passing non-constant command string (those depends on user inputs) to shell is dangerous because escaping string is error-prone. There has been an unlimited supply of arbitrary code execution bugs caused by incorrect escaping.

I think the octave documentation should be amended as well to state these well-known risks for `tempname` and `system`.

cbm755: Good point!  Let's file some new issues from this discussion:
 * `help sympref` should mention the inherent insecurity of these alternative IPC mechanisms.
 * systmpfile could use some `if is_octave()`  (see other code, something like a variable OCTAVE_VER is defined or not).

I suspect a patch to upstream Octave for tempname and system would be accepted.

PR #1147, opened 2022-06-22
@sym/mrdivide: Simplify implementation.

### LINKED ISSUE #1118: Insure left and right division is working properly with mldivide.m and mrdivide.m
Due to some failing tests we noticed that `A = C/B` where `C` was generated by `A*B=C` was failing in `mrdivide.m`.  The issue is issue #1079.  In an attempt to fix this issue made two minor changes to `mrdivide.m`.  One was the a call to regular division if B is a scalar, or more accurately isn't a matrix, the other was solve call, from upstream in the case where B is a matrix.  No changes were needed for to fix `mldivide.m`.

Experimentation showed that this will fail if `B` is not invertable.

Should we add an expected failure or documentation to outline what won't work with these operators?

alexvong1995:
Use @sym/mldivide instead of running sympy code directly.

Closes #1118.

* inst/@sym/mrdivide.m: Simplify it.
* inst/@sym/mldivide.m: Adjust accordingly to avoid infinite loop.

### DIFF
--- a/inst/@sym/mldivide.m
+++ b/inst/@sym/mldivide.m
@@ -1,4 +1,5 @@
 %% Copyright (C) 2014, 2016, 2018-2019 Colin B. Macdonald
+%% Copyright (C) 2022 Alex Vong
 %%
 %% This file is part of OctSymPy.
 %%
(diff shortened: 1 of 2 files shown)


cbm755:
LGTM!

Issue #1142, opened 2022-06-21
Get `pkg test symbolic` working


cbm755:
I think there is a new `pkg test ...` feature upstream.  We should use it, speed up the deprecation of `octsympy_tests`.

I'd like to see it used in our CI.  C.f., #1141 and #1060.

cbm755:
And just checked and this now works, presumably b/c of the #1060 fix.  So I guess we can use it in CI for example

cbm755:
Still seems to run tests for all packages not just symbolic :(  Doesn't happen locally

alexvong1995:
> Still seems to run tests for all packages not just symbolic :( Doesn't happen locally

Running `docker exec oc octave-cli --eval "pkg test symbolic"` in CI actually works!

```
  PASS                             2400
  FAIL                                0
  XFAIL (expected failure)           29
```

Maybe there's something wrong with the Makefile...

alexvong1995:
Looks like an upstream bug to me. In m-file https://hg.savannah.gnu.org/hgweb/octave/file/04120d65778a/scripts/pkg/pkg.m, if we remove those `;` in `case "test"`..., we can see the value of the variables:

```
installed_pkgs_lst =
{
  [1,1] =
    scalar structure containing the fields:
      name = symbolic
      version = 2.9.0+
      date = 2021-10-16
      author = Colin B. Macdonald <cbm@m.fsf.org>
      maintainer = Colin B. Macdonald <cbm@m.fsf.org>
      title = Octave Symbolic Package using SymPy
      description = Adds symbolic calculation features to GNU Octave.  These  include common Computer Algebra System tools such as algebraic operations,  calculus, equation solving, Fourier and Laplace transforms, variable  precision arithmetic and other features.  Compatibility with other symbolic  toolboxes is intended.
      categories = symbolic
      url = https://octave.sourceforge.io/symbolic
      depends =
      {
        [1,1] =
          scalar structure containing the fields:
            package = octave
            operator = >=
            version = 4.2
      }
      systemrequirements = python, sympy (>= 1.2), mpmath (>= 1.0)
      license = GPL-3.0-or-later
      dir = /usr/share/octave/packages/symbolic-2.9.0+
      archprefix = /usr
      loaded = 1
}
Testing functions in package 'symbolic':
installed_pkgs_dirs =
{
  [1,1] = /usr/share/octave/packages/symbolic-2.9.0+
  [1,2] = /usr
}
installed_pkgs_dirs =
{
[...]

PR #1152, opened 2022-06-24
CI: clean up jobs

cbm755:
* trying to use `pkg test symbolic` for #1142
* trying to use the Makefile more
* trying not do call octsympy_test (although the Makefile might).
    - later that makefile should use `oruntests` or `pkg test symbolic` or whatever, but only on newest Octave


cbm755:
@alexvong1995 here's some CI cleanup.  Merge if you want, or I'll merge it later.

PR #1153, opened 2022-06-25
private/store_vars_in_python: Fixes regression...

### LINKED ISSUE #1151: Regression caused by cd263b3853482226dbffe99486ff98c8445d6a63
I realize cd263b3853482226dbffe99486ff98c8445d6a63 causes a regression because in the following code:

```
     elseif (isequal (x, []))
      var_pyobj.append (py.list ());
```

If `x` is `''`, we have `isequal ('', [])` being evaluated to true which is not what we want.

See also the discussion in #1136.

alexvong1995:
...caused by cd263b3853482226dbffe99486ff98c8445d6a63.

Closes #1151.

* inst/private/store_vars_in_python.m: Update it.
* inst/pycall_sympy__.m: Add test accordingly.


cbm755:
LGTM!

PR #1158, opened 2022-06-29
CI: try to unload and uninstall packages

### LINKED ISSUE #1156: CI: exercise more "pkg" commands
Part of https://wiki.octave.org/Reviewing_Octave_Forge_packages says:

- - - - - 
### pkg-tool runs without errors or warnings

- [ ] `pkg install   foo`
- [ ] `pkg load      foo`
- [ ] `pkg unload    foo`
- [ ] `pkg uninstall foo`
- Above steps were run on Octave and OS versions:
  -  `________`
  -  `________`

- - - - 

So let's do those "install, load, unload, uninstall" cycle in our CI jobs then we can have a larger list of Octave versions on the checklist.

cbm755:
See #1156.

~~Not a fix: `make test` does not install the package but runs it in place...~~

Fixes #1156.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -98,7 +98,7 @@ jobs:
           docker exec oc      git config --global --add safe.directory /home/jovyan/octsympy
           docker exec oc pip3 install packaging
           docker exec oc pip3 install sympy==$SYMPY
-      - name: Running in-place BIST
+      - name: Run BIST in-place
         run: docker exec oc make -C octsympy test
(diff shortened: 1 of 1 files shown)


cbm755:
@alexvong1995 I cleaned up the CI a bit here: see multiple `name: ... run: ...` stages inside each job.  I also combined the BIST and doctests for pythonic: maybe we want to do that for the bigger matrix too.

Anyway, I plan to merge this soon, or go ahead once you've seen it.

PR #1165, opened 2022-07-01
CI: Build HTML documentation.

### LINKED ISSUE #1157: CI: run a "make html" job
Is it worth generating artifacts etc?

Main thing is to fail on warnings.  IIRC, that would catch PR with errors in texinfo docstring formatting for example.

alexvong1995:
Closes #1157.

* .github/workflows/main.yml: Add step in 'bist' job accordingly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -245,6 +245,11 @@ jobs:
       - name: Load package, run tests
         run: |
           octave --eval "pkg test symbolic"
+      - name: Build HTML documentation
+        run: |
+          octave --eval "pkg install -forge generate_html"
+          make html
+          octave --eval "pkg uninstall generate_html; pkg list"
       - name: uninstall package
         run: |
           octave --eval "pkg uninstall symbolic; pkg list"


cbm755:
This adds 4 mins to each BIST job.  Maybe we should just do it once, say in that generic Ubuntu job...?  Or as a branch new job.

If there is an error in the docs, does this fail?

cbm755:
I tried deleting a tag and it certainly does die:
```
octave-help-XYWRQY:145: `@end' expected `example', but saw `defmethod'
octave-help-XYWRQY:145: `@end' expected `example', but saw `defmethod'
error: __makeinfo__ returned with error code 1
. Couldn't parse      texinfo:
###### OCTAVE START ######

@documentencoding UTF-8
@defmethod @@sym atan (@var{x})
Symbolic atan function.

Example:
@example
@group
syms x
y = atan (x)
  @result{} y = (sym) atan(x)
@end group

Note
error: called from
    __texi2html__ at line 45 column 5
    __html_help_text__ at line 35 column 17
    generate_package_html>wrote_html at line 842 column 5
    generate_package_html at line 207 column 7
make: *** [Makefile:79: tmp/symbolic-html] Error 1

```
So this is good.   Just not completely sure we should do it every job, but that could be tweaked later.

alexvong1995:
Agree, we only have to build the HTML doc in one job, as the HTML doc doesn't change with the octave and sympy version.

Issue #1003, opened 2020-02-04
(symbolic) dsolve should return expression instead of equation


mass2010chromium:
When I execute the following code in matlab, it works without any problems:
```
clear all
clc
syms y(t)
dy = diff(y)
deq = diff(y, 2) + dy + y == 0
soln = dsolve(deq, y(0) == 1, dy(0) ==  0)
y_eval = double(vpa(subs(soln, t, 4)))
```
and gets out a double value. But running this code in octave breaks because `dsolve` returns an Equality, which gets processed by `vpa()` into another equality, which can't be converted to a single number by `double()`.

![octave](https://user-images.githubusercontent.com/11793417/73709508-ebb1fa00-46c6-11ea-9feb-8edd0c613ba0.png)
Notice how in the second to last line in the image, the result is type sym instead of an expression / just a vpa number.

A simple fix would be to force dsolve to return the rhs of the value it gets from sympy's dsolve, but this might break some things that rely on multiple return values from dsolve or implicit solutions.

I don't know what matlab does if it can't find an explicit solution, more testing needed.

TL;DR `dsolve()` should return an expression for compatibility with matlab.

mtmiller:
Thanks for the detailed bug report!

Just for some more info, according to the docs for [Matlab `dsolve`](https://www.mathworks.com/help/symbolic/dsolve.html), the return value should be analogous to `solve`. If solving one differential equation, return a scalar symbolic expression. If solving a system of differential equations, return a struct with fields named after the functions, or as multiple return values if `nargout` is greater than 1.

mass2010chromium:
Do we want octave's `dsolve` to behave just like matlab's `dsolve`, and not like sympy's `dsolve`? Because right now there's some code in dsolve that is doing classifying that definitely isn't in the matlab spec, and it gets in the way of returning multiple solution equations.

mass2010chromium:
PR: #1004

cbm755:
It would be nice if it behaved more like Matlab, but we also want to minimize how much code we're maintaining (to be different from SymPy).  Can be hard to balance those things.

I have try to review your merge request when I get a chance: thanks very much for that.

PR #1168, opened 2022-07-03
dsolve: return type changes for compatibility

cbm755:
This builds on #1004.

Sorry for the long delay @mass2010chromium but thank for your edits!

Fixes Issue #183.  Fixes Issue #1003.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -26,6 +26,7 @@ Johannes Maria Frank
 Chris Gorman
 Fernando Alvarruiz
 Tasos Papastylianou
+Jing-Chen Peng
 
 (Please contact the developers if your name should be here but isn't!)
 
--- a/inst/@sym/dsolve.m
+++ b/inst/@sym/dsolve.m
@@ -189,7 +189,7 @@ function [soln,classify] = dsolve(ode,varargin)
           '    return_data = dict()'
           '    for solution_part in sol:'
           '        results = convert(solution_part)'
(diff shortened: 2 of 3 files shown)


inst/@sym/dsolve.m
```diff
@@ -180,13 +167,43 @@
...
+          % Not iterable
+          '    pass'
+          'else:'
+          % Iterable
+          '    return_data = dict()'
+          '    for solution_part in sol:'
+          '        results = convert(solution_part)'
+          '        if results is None:'
+          '            return 0, -1'
+          '        return_data[results[0]] = results[1]'
```
cbm755, line 194: I'm concerned here if the thing we iterate over could have multiple solutions then maybe two things in the iterator end up with key `"y"` (say).  In this case the 2nd time would overwrite the first one.

cbm755:
@alexvong1995 tests pass, will merge soon.

CONTRIBUTORS
```diff
@@ -26,6 +26,7 @@ Johannes Maria Frank
 Chris Gorman
 Fernando Alvarruiz
 Tasos Papastylianou
+Jing-Chen Peng
```
cbm755, line 29: @mass2010chromium Is this how you'd like you name to appear in the credits?

(I know this is a very old pull request and I apologize for taking so long to get to it, but thank you for your contributions!)

mass2010chromium: works for me, thank you!

PR #1170, opened 2022-07-03
Run bist and doctests in same job

cbm755:
This will slow down each job by a couple of minutes but will decrease
overall number of jobs.


cbm755:
Github doesn't have a MWPS button?  Anyway my intention is to merge when pipeline succeeds.

PR #1175, opened 2022-07-04
Relegate the on-close for emergencies

cbm755:
@alexvong1995 how about this?

We do the waiting and verbosity etc in the explicit reset case.  Then the `onClose` stuff becomes kind of an emergency catch-all for situations (like "pkg unload").  To me, this makes it clear, that its `pkg unload`'s fault for not giving us a hook ;-)


inst/private/python_ipc_popen2.m
```diff
@@ -74,9 +93,13 @@
...
 
+    then = @(f, g) g (f ());
     % this will be called when we `clear cleanup`: avoids dangling file handles
-    cleanup = onCleanup (@() python_ipc_popen2_reset (fin, fout, pid));
-
+    cleanup = onCleanup (@() ...
+                          then (@(varagin) ...
+                                 [is_valid_file_id(fin) && fclose(fin), ...
+                                  is_valid_file_id(fout) && fclose(fout)], ...
+                                @(varargin) waitpid(pid)));
```
cbm755, line 102: how big a crime is closing here **without** calling `waitpid`?  Then we would not need the `then`...

Note: in the explicit reset case (and a glorious future where `pkg unload` has a hook), we will have waited on `pid` earlier and all of these various will be empty.

alexvong1995: If we don't `waitpid`, we'll be left with a zombie process `python3 <defunct>`, because `fclose (fin)` will terminate the child process (i.e. `python3`). On the other hand, if we `waitpid` before `fclose (fin)`, it will block, because `waitpid` will wait for the child process (i.e. `python3`) to terminate.

cbm755: thanks!

cbm755:
So can you look over this one?  It builds on top of your commit, but restores things to be closer to how they were before @mtmiller split off python_ipc_popen2_reset.m in dfe236d471989f8eb062ab56935bed8866b9700f

So in the explicit reset situation, the onCleanup code becomes a no-op.

inst/private/python_ipc_popen2.m
```diff
@@ -43,11 +44,29 @@
...
+      waitpid (pid);
+      pid = [];
+      A = A && (t == 0);
+    end
+    if (~ isempty (fout))
+      t = fclose (fout);
+      fout = [];
+      A = A && (t == 0);
+    end
+    emergency_cleanup = [];  % note: triggers emergency file-close
```
alexvong1995, line 66: I'm confused about this line. Why do we need to "trigger emergency file-close"? IIUC we've just cleaned things up above.

inst/private/python_ipc_popen2.m
```diff
@@ -43,11 +44,29 @@
...
+    A = true;
+    if (~ isempty (fin))
+      % produces a single newline char: not sure why
+      t = fclose (fin);
+      fin = [];
+      waitpid (pid);
+      pid = [];
+      A = A && (t == 0);
+    end
+    if (~ isempty (fout))
```
alexvong1995, line 61: Should we change `~ isempty (fin)` and `~ isempty (fout)` to `is_valid_file_id (fin)` and `is_valid_file_id (fout)` respectively?

alexvong1995 (commented):
Otherwise LGTM!

inst/private/python_ipc_popen2.m
cbm755: Its going to get triggered eventually.  Might as well trigger it now, when I know it is an no-op.

I guess I'm vague uncomfortable with this split: here on line 63 we close `fout`, let's say its 24.  My understanding of how anonymous functions work is that the onCleanup still has "24" hardcoded in it (it has no idea `fout` has been replaces with `[]`).

Some other `fopen` call later reuses 24.

Now our ticking timebomb onCleanup goes off and closes that 24.  I don't know if that can really happen but the idea that it might means I don't like that onCleanup lying around.   Let it go off right now while its harmless.

cbm755: maybe I should have a comment that gives that idea: `% trigger while we know it is a no-op`?

inst/private/python_ipc_popen2.m
cbm755: good point!  That does seem safer if the onCleanup somehow triggered *before* this code!

inst/private/python_ipc_popen2.m
alexvong1995: Sure, this should make things clearer

mmuetzel:
> its `pkg unload`'s fault for not giving us a hook

Not sure what exactly you need. But the `PKG_DEL` scripts might do something similar. See: https://docs.octave.org/latest/Creating-Packages.html
Essentially, these scripts are executed whenever the path containing them is removed from Octave's load path. That is happening on `pkg unload`. (But admittedly, it's not a real hook but just a script.)
Maybe you could use (`mlock`ed) functions that keep some values stored in persistent variables if you need to access them from the `PKG_DEL` script?

As a side note (and I don't know if that matters here): Persistent variables are cleared whenever a function gets reparsed. There are no guarantees when that is happening (unless you `mlock` that function).

PR #1177, opened 2022-07-06
Fixes #1145 (Create temporary files w/o race conditions whenever possible)

### LINKED ISSUE #1145: Create temporary files w/o race conditions whenever possible
We should use `tmpfile` or `mkstemp` to create temporary files whenever possible. Otherwise, fall back to `tempname` and print a warning message.

See also the discussion in #1140.

alexvong1995:
This PR could be too ambitious but adding 2 new functions is the cleanest way I can think of to fix #1145 once and for all. These functions cannot be private because they are needed by tests so we append `__` to their names.


inst/make_temp_dir__.m
```diff
@@ -0,0 +1,50 @@
...
+%% This function is not really intended for end users.
+%%
+%% @end deftypefun
+
+
+function path = make_temp_dir__ (tmpdir, prefix)
+  cmd = {'import tempfile'
+         '(tmpdir, prefix) = _ins'
+         'return tempfile.mkdtemp(dir=tmpdir, prefix=prefix)'};
+  path = pycall_sympy__ (cmd, tmpdir, prefix);
```
cbm755, line 34: not a huge fan of having temp dir creation depend on Python...!

Is this really not do-able on Octave?

inst/make_temp_dir__.m
```diff
@@ -0,0 +1,50 @@
...
+%! assert (~isempty (strfind (path, tempdir ())));
+%! assert (~isempty (strfind (path, 'octsympy-')));
+%! assert (isfolder (path));
+%! assert (isequal (ls (path), ''));
+%! assert (rmdir (path));
+
+%!error <Python exception: FileNotFoundError>
+%! if (exist ('OCTAVE_VERSION', 'builtin'))
+%!   make_temp_dir__ ('/nonexistent', '');
+%! end
```
cbm755, line 50: Bikeshedding here but seems to me this should verify that `/nonexistent` is in fact non-existent after line 47.

inst/@sym/sym.m
```diff
@@ -953,7 +954,7 @@
 %! syms x
 %! y = 2*x;
 %! a = 42;
-%! myfile = tempname ();
+%! [fd, myfile] = make_temp_file__ (tempdir (), 'octsympy-myfile-');
 %! save (myfile, 'x', 'y', 'a')
 %! clear x y a
 %! load (myfile)
```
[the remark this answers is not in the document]
alexvong1995: Agree, should have closed the unused fd here

inst/make_temp_dir__.m
alexvong1995: Good idea. Debian-based distros use `/nonexistent` as a convention, but we should make sure it's indeed non-existent

inst/make_temp_dir__.m
alexvong1995: AFAIK octave has no function to create temporary directory.

If java is enabled, we can just use java. For *nix, we can use the shell command `mktemp -d`... So basically we are left with the case of octave without java running on windows. Do you know if all octave builds on windows include java? If it's the case, problem solved!

inst/@sym/sym.m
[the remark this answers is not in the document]
cbm755: but then this is no better than any other racy way...   i.e., does this offer any benefit over `tempname`?

inst/make_temp_dir__.m
mmuetzel: > Do you know if all octave builds on windows include java?

Java isn't included in the installer of Octave for Windows. Octave just tries to detect if a compatible JRE "happens" to be installed that it can use.
So, you shouldn't rely on being able to use Java.

inst/@sym/sym.m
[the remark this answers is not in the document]
alexvong1995: This is better than `tempname` because `tempname` does not choose the filename and create the file in a single step (aka atomic operation). The fd is opened for convenience but it's not essential. The `mktemp` shell command similarly returns a filename instead of a fd, but it's still secure.

inst/make_temp_dir__.m
alexvong1995: I see. Since cygwin / msys2 is part of octave in windows (as indicated in #1182), I think we can use the `mktemp` program.

mmuetzel:
> Is it possible to call a private function in bist?

That's [bug #38776](https://savannah.gnu.org/bugs/index.php?38776).

PR #1184, opened 2022-07-08
Explicit error handling on file ops

cbm755:
I found a bunch of `fopen`, `fputs`, `fclose` with no error handling.  So add that.


inst/private/python_ipc_system.m
```diff
@@ -113,12 +113,21 @@
         disp (['This session will use the temp file "' tmpfilename '"'])
       end
     end
-    fd = fopen (tmpfilename, 'w');
+    [fd, msg] = fopen (tmpfilename, 'w');
+    if fd < 0
+      error (msg)
+    end
```
alexvong1995, line 119: Should we prepend `system ipc:` to the error message and show the temp file name as well?

```
      error ('system ipc: failed to create temporary file %s: %s',
             tmpfilename, msg);
```

inst/private/python_ipc_system.m
```diff
@@ -113,12 +113,21 @@
...
-    fputs(fd, bigs);
-    fclose(fd);
+    status = fputs(fd, bigs);
+    if status < 0
+      error ('failed to write the big string to temporary file')
+    end
+    status = fclose(fd);
+    if status < 0
+      error ('failed to close temporary file after writing')
+    end
```
alexvong1995, line 130: Similarly, we can prepend `system ipc:` to the error message and show the temp file fd here (if needed).

alexvong1995 (commented):
should be all good otherwise!

cbm755:
Thanks @alexvong1995.

Will merge when pipeline succeeds

PR #1185, opened 2022-07-08
Change non-existent name and verbosely warn user

### LINKED ISSUE #1183: reduce scary-looking false positive in testsuite
rename `make_an_attribute_err_exception` to `This_is_supposed_to_raise_an_error`?

Or can we remove this test altogether?  It scares me every time...

cbm755:
We have some scary verbose errors during our testsuite.  I don't know
how to make them quieter.  So instead wrap them with even more
verbosity.  Fixes #1183.


alexvong1995 (approved):
The message displays correctly in CI jobs https://github.com/cbm755/octsympy/runs/7257961144?check_suite_focus=true#step:4:427 and https://github.com/cbm755/octsympy/runs/7257960571?check_suite_focus=true#step:4:527 so the change should be good!

PR #1192, opened 2022-07-15
@double/adjoint: Add new function.

### LINKED ISSUE #1049: adjoint function for numerical input
https://savannah.gnu.org/bugs/?60878

I can help with this issue.

alexvong1995:
Fixes #1049.

* inst/@double/adjoint.m: New file.
* INDEX: Adjust accordingly.
* NEWS: Adjust accordingly.

### DIFF
--- a/inst/@double/adjoint.m
+++ b/inst/@double/adjoint.m
@@ -1,0 +1,86 @@
+%% Copyright (C) 2022 Alex Vong
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <https://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @documentencoding UTF-8
+%% @defun adjoint (@var{M})
+%% Numerical Classical Adjoint / Adjugate of a square matrix.
+%%
+%% @strong{Note on Terminology}: This function computes the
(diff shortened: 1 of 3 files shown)


inst/@double/adjoint.m
```diff
@@ -0,0 +1,89 @@
...
+%% @end group
+%% @end example
+%% @seealso{@@sym/adjoint}
+%% @end defun
+
+%% Reference: https://en.wikipedia.org/wiki/Adjugate_matrix
+
+
+function A = adjoint (M)
+  A = double (adjoint (vpa (M)));
```
cbm755, line 76: If user sets `digits(3)` then this code will respect that and (I guess?) give a poor result even though they would not expect a double function to care...

Similarly, if user sets `digits(1000)` this will take a lot longer before truncating to double precision.

inst/@double/adjoint.m
```diff
@@ -0,0 +1,89 @@
...
+end
+
+
+%!test
+%! M = randn (2);
+%! A = [M(2,2) -M(1,2); -M(2,1) M(1,1)];
+%! assert (isequal (adjoint (M), A));
+%! assert (isequal (adjoint (adjoint (M)), M));
```
cbm755, line 84: these look like floating point equality tests.  Is it reasonable to expect a double-precision function to satisfy this test?

alexvong1995: Indeed, should I do `abs (x - y) < 1e-15` instead?

inst/@double/adjoint.m
alexvong1995: Good catch! It's probably better to call `pycall_sympy__` directly instead of relying on `vpa`.

inst/@double/adjoint.m
cbm755: I suppose so, in a relative sense would be better.  Maybe some of the other `@double/<fcn>` have some clues.

I worry a bit about the numerical stability of this sort of implementation (SymPy is not well-known for stability of its implementations when `sympy.Float` is involved: basically unless it calls `mpmath` I don't really trust it).

The short-term fix is that I think a random test is ill-advised unless you know the algorithm used is numerically stable.  I'd suggest a couple hardcoded cases.

cbm755: How about a nearly singular one?  `[0 1; 0.0001 1]`.  Does that still give 1e-15?

alexvong1995: It works for this particular case in sympy 1.10.1 with the default method of computing determinant. But let's keep it simple and use a fixed constant to avoid any future issues...

Issue #1055, opened 2021-10-17
eq: giving TypeError: cannot determine truth value of Relational


cbm755:
On SymPy 1.9:
```
>> eqn = 2*x + 7 == 17
error: Python exception: TypeError: cannot determine truth value of Relational
    occurred at line 16 of the Python code block:
    q[i] = _op(*[k[i] if isinstance(k, MatrixBase) else k for k in _ins])
error: called from
    pycall_sympy__ at line 178 column 7
    elementwise_op at line 99 column 5
    ineq_helper at line 35 column 5
    eq at line 91 column 5
```
Same for `eq(2*x+7, 17)`.  This is leading to a lot of test failures.

Might be useful to determine the behaviour on 1.6, 1.7, 1.8 as well

cbm755:
Works on 1.8

cbm755:
Most (all?) of these have been worked around, closing in favour of #1124 where we look for a more permanent solution

Issue #1052, opened 2021-10-16
laplace test giving deprecation warning on SymPy > 1.9


cbm755:
octsympy_tests:
```
.../octsympy.git/inst/@sym/laplace.m .......

/inst/sympy/matrices/repmatrix.py:85: SymPyDeprecationWarning: 

non-Expr objects in a Matrix has been deprecated since SymPy 1.9. Use
list of lists, TableForm or some other data structure instead. See
https://github.com/sympy/sympy/issues/21497 for more info.

  SymPyDeprecationWarning(
```

cbm755:
Good idea for a wiki page.   I don't think I have anything written down about it, but I can jot down a few things...

alexvong1995:
Has this been fixed on main? I can't see the deprecation warning for `laplace` in the latest CI build anymore. There're deprecation warnings for other functions though.

cbm755:
Before closing this, we should look at https://github.com/sympy/sympy/issues/21504

CarloWood:
I found this issue, but added a new one anyway because mine is unrelated to laplace:
https://github.com/gnu-octave/symbolic/issues/1310

Nevertheless, it is likely related.

Issue #1124, opened 2022-05-06
tracker issue for non-Expr in matrices


cbm755:
We have various issues stemming from upstream SymPy's decision (sensible decision, that is!) to not allow things that are not expressions into matrices.

Examples of things we do:
- True/False
- Equations/Inequalities:  `Eq(x, y)` in Python
- `None`

I'll link issues / failing tests / etc to this Issue number.

## TODO list

- [ ] failing test in `isNone`
- more

cbm755:
For 3.0.0, we don't need permanent fix (its only deprecated upstream), but of course would be nice...

alexvong1995:
I think the the test failure is partly due to `dbout` being define in `inst/private/python_header.py` (for non-Pythonic IPC) but not in `inst/private/python_ipc_native.m` (for Pythonic IPC).

I'm not sure what to do about it. It seems there're more code in `inst/private/python_header.py` than in `inst/private/python_ipc_native.m`.

cbm755:
> Alternatively if the goal is just printing support then perhaps TableForm can be used.

https://docs.sympy.org/latest/explanation/active-deprecations.html#deprecated-non-expr-in-matrix

@alexvong1995  We should investigate whether TableForm would work instead of Array.

alexvong1995:
From the doc-string of `TableForm`, the only documented way to extract the original data is to use `as_matrix()`, which is obviously not what we want. One can do it using the undocumented attribute of `_lines` though.

```
Help on class TableForm in module sympy.printing.tableform:

class TableForm(builtins.object)
 |  TableForm(data, **kwarg)
 |  
 |  Create a nice table representation of data.
 |  
 |  Examples
 |  ========
 |  
 |  >>> from sympy import TableForm
 |  >>> t = TableForm([[5, 7], [4, 2], [10, 3]])
 |  >>> print(t)
 |  5  7
 |  4  2
 |  10 3
 |  
 |  You can use the SymPy's printing system to produce tables in any
 |  format (ascii, latex, html, ...).
 |  
 |  >>> print(t.as_latex())
 |  \begin{tabular}{l l}
 |  $5$ & $7$ \\
 |  $4$ & $2$ \\
 |  $10$ & $3$ \\
 |  \end{tabular}
 |  
 |  Methods defined here:
 |  
 |  __init__(self, data, **kwarg)
 |      Creates a TableForm.
 |      
 |      Parameters:
 |      
 |          data ...
[...]

PR #1194, opened 2022-07-17
draft: Use a SymPy Array not a Matrix for non-Expr

cbm755:
It has mostly the same semantics as Matrix but can contain more general
things.  There might be some issues about different shapes of empties
though...  WIP on a fix for #1052 and #1055.


cbm755:
@alexvong1995 here's a half-baked idea for Issue #1124: roughly when we have non-`Expr` to be put in a `Matrix`, we instead make an `Array`.  This will be important for future SymPy as non-Expr in Matrix is deprecated.

cbm755:
@alexvong1995 What do you think of this?  Basically in a few places (probably not that many) we return Array instead of Matrix.  Specifically we return Matrix greedily whenever everything is a Expr, else Array.

(It will need fixes upstream but I'm less worried about that.)

Maybe you can try doing this in the `mat_rclist_` and other helper functions, just to get a feel for the idea.

alexvong1995:
I think the the test failure is partly due to `dbout` being define in `inst/private/python_header.py` (for non-Pythonic IPC) but not in `inst/private/python_ipc_native.m` (for Pythonic IPC).

I'm not sure what to do about it. It seems there're more code in `inst/private/python_header.py` than in `inst/private/python_ipc_native.m`.

cbm755:
that could be true: I haven't looked at pythonic much recently.  You can always comment out "dbout()" (it is just debugging).  With Pythonic you can probably just `print()` instead.

Anyway, I think even concentrating on `system`-based, this idea is not complete.

cbm755:
@alexvong1995 you can rebase and force push this on top of main (if you like).

alexvong1995:
I think what we should do next is to make `@sym/horzcat` and `@sym/vertcat` Array-compatible. I haven't done any tests yet but I suspect all those `[t t; f f]` where `t` is `sym(true)` and `f` is `sym(false)` are causing many test failures.

alexvong1995:
On Monday, August 29th, 2022 at 12:04 AM, Colin B. Macdonald ***@***.***> wrote:


> @alexvong1995 you can rebase and force push this on top of main (if you like).
> 

rebased onto main and force pushed as a7c44c0d78bd6be66b6795ed339751657b1c3a3d

inst/@sym/horzcat.m
```diff
@@ -45,25 +46,8 @@
...
-          '            _proc.append(i)'
-          'return sp.MatrixBase.hstack(*_proc),'
-          };
-
-  for i = 1:nargin
-    varargin{i} = sym(varargin{i});
-  end
-  h = pycall_sympy__ (cmd, varargin{:});
+  args = cellfun (@transpose, varargin, 'UniformOutput', false);
+  h = vertcat (args{:}).';
```
cbm755, line 50: while I think of it: I approve of this kind of "surface decrease" (having fewer functions that call `pycall_sympy__`) BUT it is important to be aware that there is a downside: namely more "roundtrips" between Python and Octave.  This is most noticeable with `ipc-system` but even with popen2, there are additional copies being made as everything is serialized back-and-forth over the pipe.

Its analogous to HTTP where all objects keep getting converted back-and-forth to JSON.

No action required but its something to keep in mind: perfect software design here might blow-up the latency and IO costs.

cbm755:
I think there was a way to turn an warning into an exception in Python.  That might be useful when debugging here...  like `SymPyDeprecationWarning` becomes an exception.

inst/@sym/private/elementwise_op.m
```diff
@@ -97,7 +97,10 @@
           'elements = []'
           'for i in range(0, len(q)):'
           '    elements.append(_op(*[k[i] if isinstance(k, MatrixBase) else k for k in _ins]))'
```
cbm755, line 99: I'm not sure what test hits this code on line 99 yet, but I think from doing #1226 that this will need to be replaced with either 2-d indexing or some kind of `flatten()`

cbm755: wait, I think I already did this in #1223...  :-X

alexvong1995:
On Friday, September 2nd, 2022 at 4:57 PM, Colin B. Macdonald ***@***.***> wrote:


> I think there was a way to turn an warning into an exception in Python. That might be useful when debugging here... like `SymPyDeprecationWarning` becomes an exception.
>
The following example works for me:


--8<---------------cut here---------------start------------->8---
from sympy import Matrix
from sympy.utilities.exceptions import SymPyDeprecationWarning
import warnings


warnings.filterwarnings("error", category=SymPyDeprecationWarning)
Matrix([[True]])
--8<---------------cut here---------------end--------------->8---


I figure this out after reading multiple issues https://github.com/sympy/sympy/issues?q=is%3Aissue+is%3Aopen+SymPyDeprecationWarning of sympy. It would be good to know if this is documented somewhere or not.

alexvong1995:
On Friday, September 2nd, 2022 at 4:57 PM, Colin B. Macdonald ***@***.***> wrote:


> I think there was a way to turn an warning into an exception in Python. That might be useful when debugging here... like `SymPyDeprecationWarning` becomes an exception.
> 

The following example works for me:

--8<---------------cut here---------------start------------->8---
from sympy import Matrix
from sympy.utilities.exceptions import SymPyDeprecationWarning
import warnings

warnings.filterwarnings("error", category=SymPyDeprecationWarning)
Matrix([[True]])
--8<---------------cut here---------------end--------------->8---

I figure this out after reading multiple issues https://github.com/sympy/sympy/issues?q=is%3Aissue+is%3Aopen+SymPyDeprecationWarning of sympy. It would be good to know if this is documented somewhere or not.

alexvong1995:
On Friday, September 2nd, 2022 at 4:57 PM, Colin B. Macdonald ***@***.***> wrote:


> I think there was a way to turn an warning into an exception in Python. That might be useful when debugging here... like `SymPyDeprecationWarning` becomes an exception.
> 

# The following example works for me:


from sympy import Matrix
from sympy.utilities.exceptions import SymPyDeprecationWarning
import warnings


warnings.filterwarnings("error", category=SymPyDeprecationWarning)
Matrix([[True]])


# I figure this out after reading multiple issues https://github.com/sympy/sympy/issues?q=is%3Aissue+is%3Aopen+SymPyDeprecationWarning of sympy.
# It would be good to know if this is documented somewhere or not.

inst/@sym/horzcat.m
alexvong1995: Interesting thoughts!

Normally when working with a single language, the classic solution is to use an (optimizing) compiler to optimize away all the redundant conversions. Unfortunately, this doesn't work for us because of language barrier: a compiler can't optimize across language boundary :(

---

Some of my thoughts:

First, pythonic may have already solved our problem! While I am not familiar with the internals of pythonic, perhaps using pythonic's `py.`... instead of `pycall_sympy__` can avoid the needless conversions when going through a pipe. Of course, we should benchmark to see if it's indeed the case.

I believe this was the original *future* subproject but we changed it to the current (more urgent) one.

---

Another way is to create a octsmypy python library which contains one python function for each m-file function.

For example, say we want to implement m-file function `@sym/transpose`, `@sym/vertcat` and `@sym/horzcat`. Then our octsmypy python library should contain them as well. More precisely, we can do:

```
from octsympy.sym import transpose, vertcat, horzcat
```

To implement horzcat in terms of vertcat and transpose, we do it inside the octsmypy python library:

```
def horzcat(*XX):
    return transpose(vertcat(*[transpose(X) for X in XX]))
```

without the redundant conversions.
    
Finally, the m-file functions `@sym/transpose`, `@sym/vertcat` and `@sym/horzcat` are now thin wrappers over the corresponding octsympy python library functions, such as:

```
function h = vertcat (varargin)
    args = ...
    h = pycall_sympy__ ('from return octsympy.sym.vertcat(*_ins)', args{:});
end
```

and so on...

Perhaps this can be mentioned in the final report as possible future projects (?)

cbm755: I agree Pythonic offers possibilities.  I think it currently still does conversion: a 'sym' is really just a struct of the 'srepr' (string representation) from SymPy, indep of IPC mechanism.  But your 'xsym' idea would change that.

Agree about mentioing these things ad future projects.

Another thing to mention: CI pythonic runs are twice as slow as Popen2.  I susect there migit be some 'low hanging fruit' in the Pythonic project.

cbm755: And just for the record: another alternative that requires very little effort is (a) being mindful of this issue and aiming for 1 (or a small number) of language crosses per function (b) balancing that against technical debt of code duplication etc.  This is the approach we've been using so far.

Issue #1182, opened 2022-07-07
tempfile related problems on MXE windows build


cbm755:
from octave.discourse.group:

```
A couple of tests are failing in the CI with the nightly builds for Windows:
[freshly brewed Octave · gnu-octave/octave-buildbot@43f176e (github.com)](https://github.com/gnu-octave/octave-buildbot/runs/7238133932?check_suite_focus=true#step:13:15477)

Some of the errors seem to emit things like this:

  python: can't open file '/mingw64/share/octave/packages/symbolic-3.0.0/D:\a\_temp\oct-EEh6sa_octsympytmp.py': [Errno 2] No such file or directory

It looks like this could be two absolute paths that are concatenated. That is probably ok for most paths on a Linux file system. But it is failing on Windows.

Edit: That’s still with SymPy 1.5 in case that should matter.
```

**update**I have trouble finding things later on discourse, not sure why, so I don't waste 5 mins in the future, here's the link https://octave.discourse.group/t/symbolic-3-0-0-released/2975/5

cbm755:
@alexvong1995 can you look into this?

alexvong1995:
Could this be a `tempname()` problem? How should we reproduce it?

cbm755:
I'm not sure :(  I did run our test suite on a Windows machine (Octave 5.2.0) before releasing and I did not see this.  Perhaps its something about cross-compiling?  But I agree we're just using what `tempname()` gives us here, which is apparently

```
/mingw64/share/octave/packages/symbolic-3.0.0/D:\a\_temp\oct-EEh6sa_
```
That "D:" in the middle of a path seems invalid on Windows to me...  @mmuetzel is it possible this is an upstream Octave / MXE builder problem...

cbm755:
The other errors in that log look like:
```
   Traceback (most recent call last):
    File "<string>", line 1, in <module>
  FileNotFoundError: [Errno 2] No such file or directory: 'D:\x07\\octave-buildbot\\octave-buildbot\\octave-2022-07-07-00-12-w64\\mingw64\\share\\octave\\packages\\symbolic-3.0.0\\private\\python_header.py'
```
Note the very different paths which look more reasonably and don't have a mixture of slashes and backslashes... unlike what tempname is giving us: `/mingw64/share/octave/packages/symbolic-3.0.0/D:\a\_temp\oct-EEh6sa_`

cbm755:
Hold on, this construction is maybe suspicious to me:
```
[...]

Issue #1009, opened 2020-02-26
request more detail in 'sympref diagnose'


kwanlowe:
Hello:

This is opened in reference to this thread: https://octave.1599824.n4.nabble.com/Octave-unable-to-find-python-when-using-symbolic-toolbox-tt4695912.html#a4695924

To assist with troubleshooting, is it possible to provide the full path to the python binary that is being used?  I use Python virtualenv and run Octave either from the host installation (e.g., dnf install octave, apt install octave) or via Snaps or Flatpaks. Teasing out which Python is actually being called can help troubleshoot various issues.

Thanks and best regards,
Kwan

mtmiller:
> is it possible to provide the full path to the python binary

You mean you'd like `sympref diagnose` to show the full path to whatever `python` or `python3` that it found in the `PATH`, right?

Just clarifying, at first glance it looked like you were asking if _you_ could "provide" the full path to a Python of your choosing, which yes you can.

mtmiller:
See also #1008 I opened for a slightly different troubleshooting idea.

kwanlowe:
Yes, I'd like the sympref diagnose to show the python path that is being used. 

 I was trying do this with python itself. For example, to get the module info:
```
octave:25> unix("python3 -c 'help(\"modules\")'|grep -i sympy ")
ans =  1
```

I couldn't get a working snippet that would output the detail though..

kwanlowe:
Here's the snippet that should do it:
```
octave:1> unix("python3 -c 'import sys;  print(sys.executable)'")
/home/kwan/bin/python_venv/p36/bin/python3
```

mtmiller:
In OctSymPy terms, that would be

    exe = python_cmd ("return sys.executable");

Would you like to work on PR to add this info to `diagnose`?

PR #1195, opened 2022-07-18
private/python_ipc_system: Convert to Cygwin path when needed.

alexvong1995:
This should fix some errors when Python is running in Cygwin-like
environment. But there could still be errors in other places.

See #1182.

* inst/private/cygpath.m: New function.
* inst/private/python_env_is_cygwin_like.m: New function.
* inst/private/python_ipc_system.m: Use them.

### DIFF
--- a/inst/private/cygpath.m
+++ b/inst/private/cygpath.m
@@ -1,0 +1,39 @@
+%% Copyright (C) 2022 Alex Vong
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <https://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @defun  cygpath ()
+%% Convert Windows native path to Cygwin POSIX-style path.
+%%
+%% @seealso{python_env_is_cygwin_like}
+%% @end defun
+
+function posix_path = cygpath (native_path)
+  %% FIXME: only allow safe characters inside "..."
+  if ~isempty (strfind (native_path, '"'))
+    error ('cygpath: native path %s must not contain "', native_path);
+  end
+
+  [status, out] = system (['cygpath -u "' native_path '"']);
+  if status ~= 0
+    error ('cygpath: cygpath exited with status %d', status);
+  end
+
+  posix_path = regexprep (out, '[\r]?[\n]$', ''); % strip trailing newline
+  assert (logical (regexp (posix_path, '^[^\r\n]+$'))); % validate path
+end
--- a/inst/private/python_env_is_cygwin_like.m
+++ b/inst/private/python_env_is_cygwin_like.m
@@ -1,0 +1,54 @@
+%% Copyright (C) 2022 Alex Vong
+%%
+%% This file is part of OctSymPy.
+%%
+%% OctSymPy is free software; you can redistribute it and/or modify
+%% it under the terms of the GNU General Public License as published
+%% by the Free Software Foundation; either version 3 of the License,
+%% or (at your option) any later version.
+%%
+%% This software is distributed in the hope that it will be useful,
+%% but WITHOUT ANY WARRANTY; without even the implied warranty
+%% of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See
+%% the GNU General Public License for more details.
+%%
+%% You should have received a copy of the GNU General Public
+%% License along with this software; see the file COPYING.
+%% If not, see <https://www.gnu.org/licenses/>.
+
+%% -*- texinfo -*-
+%% @defun  python_env_is_cygwin_like (pyexec)
+%% Check if Python @var{pyexec} is running in a Cygwin-like POSIX environment,
+%% such as Cygwin or MSYS2.  The result is memoized to speed up subsequent
+%% calls.
+%%
+%% @seealso{cygpath}
+%% @end defun
+
+function r = python_env_is_cygwin_like (pyexec)
+  persistent python_env_is_cygwin_like_memo
+
+  if ~isempty (python_env_is_cygwin_like_memo)
+    r = python_env_is_cygwin_like_memo;
+    return
+  end
+
+  if ispc ()
+    if system ('where /q cygpath') ~= 0
+      r = false;
+      python_env_is_cygwin_like_memo = r;
+      return
+    end
+
+    [status, out] = system ([pyexec ' -c "import os; print(os.name)"']);
+    if status ~= 0
+      error ('python_env_is_cygwin_like: %s exited with status %d', ...
+             pyexec, status);
+    end
+    r = ~isempty (regexp (out, 'posix', 'match'));
+    python_env_is_cygwin_like_memo = r;
+  else
+    r = false;
+    python_env_is_cygwin_like_memo = r;
+  end
+end
--- a/inst/private/python_ipc_system.m
+++ b/inst/private/python_ipc_system.m
@@ -1,4 +1,5 @@
 %% Copyright (C) 2014-2016, 2022 Colin B. Macdonald
+%% Copyright (C) 2022 Alex Vong
 %%
 %% This file is part of OctSymPy.
 %%
@@ -130,6 +131,10 @@ function [A, info] = python_ipc_system(what, cmd, mktmpfile, varargin)
     if status < 0
       error ('system ipc: failed to close %s (fd %d) after writing', ...
 	     tmpfilename, fd);
+    end
+
+    if python_env_is_cygwin_like (pyexec)
+      tmpfilename = cygpath (tmpfilename);
     end
     [status, out] = system ([pyexec ' ' tmpfilename]);
   end


alexvong1995:
It should work in theory. I did some tests using shim but not in a real cygwin-like
environment, as wine does not support cygwin / msys2.

cbm755:
Nice work!   Maybe we can have `sympref diagnose` report this information too, see #1009.

inst/private/cygpath.m
```diff
@@ -0,0 +1,50 @@
...
+    err = errno ();
+  end
+
+  assert (ischar (posix_path));
+  assert (feof (fout));
+  assert (fclose (fout) == 0);
+
+  [retpid, status, msg] = waitpid (pid);
+  if retpid < 0
+    error ('cygpath: failed to wait for python: %s', msg);
```
mmuetzel, line 48: The error message doesn't look correct. Afaict, this isn't waiting for Python.

alexvong1995: How about lengthening it to "...wait for python to terminate"?

mmuetzel: Why "python"? Isn't this waiting for "cygpath"?

alexvong1995:
Aha, I see. I probably forgot to change it.

inst/private/cygpath.m
alexvong1995: Aha, I see. I probably forgot to change it.

mmuetzel:
IIUC, it is true that a malevolent user might be able to able to execute arbitrary programs by passing specifically grafted strings. But I'd say that this risk is pretty low if the string to be executed is essentially a literal. (Or the risk is basically the same with `popen2`, e.g., if binaries are replaced.)
On the other hand, if a user is malevolent, they might as well just use the `system` function directly in Octave. I don't currently see why they would try to route their commands through this package...

mmuetzel:
The input to `cygpath ()` (at least for the changes you propose) is the output of `tempname` with a literal suffix. That should be pretty safe. Additionally, it is something for which `fopen` returned successfully. Maybe with some Unicode shenanigans it might be possible to construct a byte sequence that evaluates to something that `fopen` is happy with but that could still cause harm when passed as a argument to `cygpath`. But that is already highly theoretical.
That would still leave the "task" of getting `tempname` to return that byte sequence. Maybe possible???
But even if it would be possible to exploit that (and I don't know if it is), Octave in that scenario would hopefully run isolated on that web service. So, the amount of damage would hopefully be limited.

But to get this PR back on track: Your changes look good. But at least for the sake of compatibility with Matlab, it might be better to use `system` imho. And maybe check pretty early on what `system ('where /Q cygpath')` returns (`0` means it could find `cygpath`, anything else means it couldn't).

cbm755:
> Does octsympy try to be Matlab compatible?

Yes, to some extent.  But the popen2 IPC stuff is Octave-only, and is faster than`system`.  Those are general answers.  But here within the system-based IPC we should use system.

mmuetzel:
I'd guess opening a pipe is only faster if there is repeated communication over the pipe. So, you don't need to spawn a new process each time.
For these "one time" invocations, there shouldn't be much of a difference with respect to performance. (I haven't done any measurements though.)

inst/private/cygpath.m
```diff
@@ -0,0 +1,42 @@
...
+
+  [status, out] = system (['cygpath -u "' native_path '"']);
+  if status ~= 0
+    error ('cygpath: cygpath exited with status %d', status);
+  end
+  assert (ischar (out));
+
+  match = regexp(out, '.*', 'match', 'dotexceptnewline');
+  assert (length (match) == 1);
+  posix_path = match{1};
```
mmuetzel, line 40: I'm not entirely sure what these lines are used for. But in case, you just want to strip a trailing newline, you could probably also use `strtrim`. Or am I missing something?

inst/private/python_ipc_system.m
```diff
@@ -131,7 +132,13 @@
       error ('system ipc: failed to close %s (fd %d) after writing', ...
 	     tmpfilename, fd);
     end
-    [status, out] = system ([pyexec ' ' tmpfilename]);
+
+    if python_env_is_cygwin_like (pyexec)
+      converted_tmpfilename = cygpath (tmpfilename);
```
mmuetzel, line 137: Do you still need the original Windows path of the temporary file?
If not, you could also re-use the existing variable and simplify this portion of the code a bit.

inst/private/cygpath.m
alexvong1995: I was thinking of using `strtrim` but `strtrim` removes whitespaces as well from both ends. AFAIK whitespaces are pretty common in windows path. Do you know of any function which is similar to Python `rstrip('\n')`?

inst/private/python_ipc_system.m
alexvong1995: I mostly code in a [functional](https://en.wikipedia.org/wiki/Functional_programming) style, avoiding unnecessary mutations, maintaining [referential transparency](https://en.wikipedia.org/wiki/Referential_transparency), etc whenever possible. I would prefer using a new variable instead of overwriting old ones. (Basically, it's the same "use `const` by default" advice provided by many es6 coding styles.)

Sadly, we can't use conditional expression. Otherwise, it'd be a one-liner `converted_tmpfilename = cygpath (tmpfilename) if python_env_is_cygwin_like (pyexec) else tmpfilename`

inst/private/cygpath.m
mmuetzel: You are right: Spaces in filenames are (unfortunately) pretty common in Windows paths. But I tried to rename a file in Windows Explorer such that it ends with a space (" ") character. It looks like it is not possible to do that. That space is automatically stripped from the file name. So, using `strtrim` or `deblank` would most probably do the "correct thing". (Especially with the automatically generated file names this changeset is about.)
Using those functions might also make it a bit easier to understand what this part of the code is actually meant to be doing. (An additional comment might also help.)

inst/private/python_ipc_system.m
mmuetzel: I guess that is a question of style and preference.
My personal preference is to avoid else branches if they "do nothing".

inst/private/cygpath.m
alexvong1995: Agree, the current approach is too clumsy... Simplified and added comment as well.

inst/private/python_ipc_system.m
cbm755: No strong feelings here but in this case I'd probably do as @mmuetzel.  A few years of Python coding as also made me adverse to `else` (unless needed of course!)

inst/private/python_env_is_cygwin_like.m
```diff
@@ -0,0 +1,53 @@
...
+%%
+%% @seealso{cygpath}
+%% @end defun
+
+function r = python_env_is_cygwin_like (pyexec)
+  persistent python_env_is_cygwin_like_memo
+
+  if ~isempty (python_env_is_cygwin_like_memo)
+    r = python_env_is_cygwin_like_memo;
+  elseif ispc ()
```
cbm755, line 33: here I would definitely do an early return.  Then `if ispc ()`.  For me, this makes the persistent semantics clearer.

BTW, I like `_memo` for such variables!  I shall try to remember that.

cbm755:
Thanks @alexvong1995 and @mmuetzel for detailed reviewing.  Merge this whenever you are both happy.

cbm755 (approved):

inst/private/python_ipc_system.m
alexvong1995: I personally find it more difficult to mentally keep track of all the places where a mutation (of binding or value) has occurred. I prefer to treat variable declarations as mathematical definitions, which cannot be changed in the middle of a proof (an example of programming in the functional style). Anyway, I've removed the else branch in this simple case as both of you find it less readable...

inst/private/python_env_is_cygwin_like.m
alexvong1995: > here I would definitely do an early return. Then `if ispc ()`. For me, this makes the persistent semantics clearer.
> 
Make sense to me, it also makes the code look more symmetric.

> BTW, I like `_memo` for such variables! I shall try to remember that.
> 
The name is borrowed from [memoization](https://en.wikipedia.org/wiki/Memoization). I think I saw it somewhere as well. This is one of the reasons why [pure function](https://en.wikipedia.org/wiki/Pure_function) (and [referential transparency](https://en.wikipedia.org/wiki/Referential_transparency)) is good!

inst/private/cygpath.m
```diff
@@ -0,0 +1,40 @@
...
+    error ('cygpath: native path %s must not contain "', native_path);
+  end
+
+  [status, out] = system (['cygpath -u "' native_path '"']);
+  if status ~= 0
+    error ('cygpath: cygpath exited with status %d', status);
+  end
+
+  %% validate output
+  assert (ischar (out) && logical (regexp (out, '^[^\r\n]+[\r]?[\n]$')));
```
mmuetzel, line 38: The second part of this condition might be too specific. I'm not sure if `system` should really add a new line to the end of the `out` string.
That might actually be a bug in Octave. (But I didn't come around to investigate.)

Maybe better check that the string after removal of the (potential) trailing newline is non-empty?

alexvong1995: > The second part of this condition might be too specific. I'm not sure if `system` should really add a new line to the end of the `out` string.
> That might actually be a bug in Octave. (But I didn't come around to investigate.)
>
I would expect a console program to append exactly 1 newline to the output path. Otherwise, it would be unreadable to human. (0 or 1 is plausible, > 1 would be a bug)

> Maybe better check that the string after removal of the (potential) trailing newline is non-empty?
>
Indeed, it must be chars according to documentation. We can validate the path at the end, which is actually safer.

inst/private/python_env_is_cygwin_like.m
```diff
@@ -0,0 +1,55 @@
...
+      python_env_is_cygwin_like_memo = r;
+      return
+    end
+
+    [status, out] = system ([pyexec ' -c "import os; print(os.name)"']);
+    if status ~= 0
+      error ('python_env_is_cygwin_like: %s exited with status %d', ...
+             pyexec, status);
+    end
+    assert (ischar (out));
```
mmuetzel, line 48: Similarly to your other recent change, this assertion might not be needed.

alexvong1995: It's now updated.

mmuetzel:
LGTM.

Issue #1191, opened 2022-07-11
adopt .oct-config files to indicate source in is utf-8


cbm755:
There are new-ish .oct-config files with this content:
```
encoding=utf-8
```
These are used to indicate that the source code is in utf-8.  Which ours is.

note: we also use have this texinfo: `@documentencoding UTF-8` in-place but I'm not sure that is used by Octave's `__mfile_encoding` / `dir_encoding` etc.  *both*, we want *both*.

#### References:
* https://github.com/catch22/octave-doctest/pull/252
* https://github.com/catch22/octave-doctest/issues/251
* https://octave.discourse.group/t/should-i-use-oct-config-when-i-have-utf-8-strings-in-m-files/3001

cbm755:
I'm getting failures in Octave 6.4.0 about this when `LANG=C` (in Fedora 35/36 package builds for example).

A workaround is to do this
```
__mfile_encoding__("utf-8")
```
before running doctests.

mmuetzel:
Afaict, those files have been added to the repository. Is there anything left to do for this?

cbm755:
I don't think Symbolic has such files yet.  We added them to the Doctest package.

My current understanding packages that have utf-8-encoded source files (like we do) shoulddeclare that so that Octave will DTRT when the system is set to something else.

Actually, it is already declared in most files like this:
```
%% -*- texinfo -*-
%% @documentencoding UTF-8
```
but I don't think Octave considers that; rather it is an indicator for texinfo tooling.

mmuetzel:
Oops. Sorry. Got confused as to which repository this was meant for. Sorry for the noise.

Issue #981, opened 2019-05-10
Last UTF character breaks terminal on Windows


E3V3A:
I'm not sure how to report this properly, since I  just tried to install this a few moments ago. 

Trying to follow the examples, it seem to sort-of  work. However, when the last UTF-8 (?) character is printed, it just prints a garbage symbol and then the CLI prompt become useless. I.e. doesn't react to any further commands. 

This is the output I get, using Octave 5.1.0 (for Win64 on Win-8.1).

![octave-gui_2019-05-10_18-43-19](https://user-images.githubusercontent.com/194392/57539896-03786f80-733b-11e9-8cf3-de53d1cca977.png)

BTW. How can I check which Python interpreter is used by the SymPy ?  
(I have 4 installed, with SymPy in all of them, but haven't set any `PYTHON` var.)

E3V3A:
A related issue seem to be:
*  point 5 in: https://github.com/cbm755/octsympy/issues/83#issuecomment-217352182
* #941 
* #912

cbm755:
`sympref diagnose` should tell you some stuff (e.g., python version).

As for the rest, I agree it looks a lot like https://github.com/cbm755/octsympy/issues/83#issuecomment-217352182

I don't know what to do about any of this except wait for some motivated windows developer to fix upstream Octave.  In the meantime, this package just has `unicode=false` on Windows.  Not ideal.

E3V3A:
@cbm755 
Yes, it's really a shame, since it pretty much works and then falls on the finishing line. Unfortunately I am not going to report anything on that shamelessly useless *Savannah* bug tracking system. I'm fed up with these old/email based bug tracking that is so poorly maintained that stuff reported more than 3 years ago, still haven't been properly addressed. If they were to migrate to GitHub I'd reconsider, but any useful contributions get lost in the bureaucratic filing process over there.

cbm755:
I certainly understand your frustration with the upstream tooling, but I think an even more urgent issue is a lack of developers.  Octave is all volunteers, and probably not enough of them run Windows systems themselves.

If you're good with Windows, an important (and faster moving) Octave project that could probably use your help would be https://gitlab.com/mtmiller/octave-pythonic/

mtmiller:
> If you're good with Windows, an important (and faster moving) Octave project that could probably use your help would be https://gitlab.com/mtmiller/octave-pythonic/

Yes please! :pray:

E3V3A:
[...]

Issue #941, opened 2019-03-11
remove utf-8 chars from our code


cbm755:
See #912, and `__mfile_encoding__`.

Try using `\x` instead of having utf8 in our code.

Does not effect docs: those stay in utf-8 and we await upstream solution

cbm755:
Upstream: https://savannah.gnu.org/bugs/index.php?56072

PR #1196, opened 2022-07-23
Add .oct-config to indicate source is in utf-8

cbm755:
These are not needed inside `@class` and `private` directories.
Fixes Issue #1191.


cbm755:
Fixes #981 (hopefully).  See also #941.

PR #1200, opened 2022-08-04
CI: Run tests on Octave 7.2.0 and latest Octave.

### LINKED ISSUE #1197: CI: test on Octave 7.2.0 and latest
I think we should be testing on 7.2.0 but also to future proof a bit can we add `latest` to the list of containers to use?  Even if a bit wasteful, I think `[6.4.0, 7.1.0, 7.2.0, latest]` might make future updates less urgent.

alexvong1995:
Fixes #1197.

* .github/workflows/main.yml: Enable them.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        octave: [7.1.0]
+        octave: [7.2.0]
         sympy: [1.4, 1.5.1, 1.6.2, 1.7.1, 1.8, 1.9, 1.10.1]
         include:
           - octave: 5.1.0
@@ -68,6 +68,8 @@ jobs:
           - octave: 6.3.0
             sympy: 1.10.1
           - octave: 6.4.0
+            sympy: 1.10.1
+          - octave: 7.1.0
             sympy: 1.10.1
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


cbm755:
I didn't realize there was no 'latest', I thought that was a builtin feature of Containers...

Perhaps its not worth the complexity?

alexvong1995:
On Thursday, August 4th, 2022 at 3:44 PM, Colin B. Macdonald ***@***.***> wrote:


> I didn't realize there was no 'latest', I thought that was a builtin feature of Containers...
> 

According to https://www.howtogeek.com/devops/understanding-dockers-latest-tag/ "latest" is used as the default tag if no tag is provided. On the other hand, it won't be created automatically if a tag is specified.

> Perhaps its not worth the complexity?
> 

I think we have 3 options here. The 1st option is to simply not test the image with the "latest" tag, which is the status quo. The 2nd option is to modify the build process of octave image so that it tags the latest image with the "latest" tag. The 3rd option is to find out the latest image ourselves, which indeed introduces some complexity. WDYT?

cbm755:
I vote for 1 for now.  And possibly file an issue upstream if you think 2 is something that *should* be done.

cbm755:
@alexvong1995 this sort of CI failure is relatively common:
```
octave: X11 DISPLAY environment variable not set
octave: disabling GUI features
error: pkg: could not download file /tmp/generate_html-0.3.3-n6hTZo.tar.gz from URL https://packages.octave.org/download/generate_html-0.3.3.tar.gz
error: called from
    pkg at line 507 column 13
Error: Process completed with exit code 1.
```
I assume its some kind of rate limiter that gets grouchy when certain IPs download lots of packages (but no evidence).

Anyway, just FYI.  I never assume CI should have 100% reliability: we get what we pay for afterall!

alexvong1995:
The test now passes on `main`. So probably just a temporary issue.

If the problem reoccurs, we can always use the `torsocks -i` command from Tor to switch to a fresh IP for each invocation.

PR #1201, opened 2022-08-05
Fixes #1009

### LINKED ISSUE #1009: request more detail in 'sympref diagnose'

alexvong1995:
This PR fixes #1009 by updating `inst/private/assert_have_python_and_sympy.m` so that `sympref diagnose`

1. Show full path of Python executable when using non-Pythonic IPC.
2. Detect if Python is running in a Cygwin-like POSIX environment when using non-Pythonic IPC.

### DIFF
--- a/inst/private/assert_have_python_and_sympy.m
+++ b/inst/private/assert_have_python_and_sympy.m
@@ -97,8 +97,8 @@ function assert_have_python_and_sympy (pyexec, verbose)
     disp ('')
     disp ('Let''s check what version of Python we are calling...')
     disp ('')
-    fprintf ('Attempting to run %s -c "import sys; print(sys.version)"\n\n', pyexec);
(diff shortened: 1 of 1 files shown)


cbm755:
I split the commands into two steps b/c I feel the commands should be simple and understandable (maybe users want to copy-paste them to their terminal for example).

This looks good to me, +1 to merge

PR #1213, opened 2022-08-29
Test Array in mat_rclist_*

### LINKED ISSUE #1211: common helper python function for Array/Matrix return
@alexvong1995, after our discussion, I think you are right: we should have some shared code for this Array versus Matrix stuff:

```
if all(isinstance(elt, Expr)'
          '           for elt in itertools.chain.from_iterable(MM))):'
          '    M = Matrix(MM)'
          'else:'
          '    M = Array(MM)'
```

I think a helper python function in python_header is a good way to do this.

alexvong1995:
This PR makes the mat_rclist_* functions `Array`-compatible.

We also add the `make_matrix_or_array` helper function as well, which fixes #1211. `dbg_no_array` is set to `True` for now to avoid failing tests in `main`.


cbm755:
This looks good at a quick glance.

Should we merge it to #1194 instead of main?

alexvong1995:
Sure, I can merge this to https://github.com/cbm755/octsympy/tree/Array_not_Matrix instead of `main` (after rebasing https://github.com/cbm755/octsympy/tree/Array_not_Matrix onto `main`).

This should be safer than merging to `main` directly.

alexvong1995:
Merged onto https://github.com/cbm755/octsympy/tree/Array_not_Matrix in 5385713ccf34534366de6838691359e3ecca4065

PR #1216, opened 2022-08-30
Make *cat Array-compatible

alexvong1995:
This PR makes `@sym/vertcat`, `@sym/horzcat` and `@sym/tranpose` Array-compatible.

In `@sym/horzcat`, we make use of the property that

```
[A_1; A_2; ...; A_n).' == [A_1.' A_2.' ... A_n.']
```
for all `A_k` with compatible sizes.

### DIFF
--- a/inst/@sym/horzcat.m
+++ b/inst/@sym/horzcat.m
@@ -46,25 +46,8 @@
 
 function h = horzcat(varargin)
 
-  % special case for 0x0 but other empties should be checked for
-  % compatibilty
-  cmd = {
-          '_proc = []'
-          'for i in _ins:'
-          '    if i is None or not i.is_Matrix:'
-          '        _proc.append(sp.Matrix([[i]]))'
-          '    else:'
-          '        if i.shape == (0, 0):'
-          '            pass'
-          '        else:'
-          '            _proc.append(i)'
-          'return sp.MatrixBase.hstack(*_proc),'
-          };
-
-  for i = 1:nargin
-    varargin{i} = sym(varargin{i});
-  end
-  h = pycall_sympy__ (cmd, varargin{:});
+  args = cellfun (@transpose, varargin, 'UniformOutput', false);
+  h = vertcat (args{:}).';
 
 end
 
--- a/inst/@sym/horzcat.m
+++ b/inst/@sym/horzcat.m
@@ -1,4 +1,5 @@
 %% Copyright (C) 2014-2017, 2019 Colin B. Macdonald
+%% Copyright (C) 2022 Alex Vong
 %%
 %% This file is part of OctSymPy.
 %%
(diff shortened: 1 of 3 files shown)


inst/@sym/transpose.m
```diff
@@ -66,8 +66,11 @@
   end
 
   cmd = { 'x = _ins[0]'
-          'if x.is_Matrix:'
+          'if isinstance(x, MatrixBase):'
           '    return x.T'
+          'elif isinstance(x, NDimArray):'
+          '    xx = x.tolist()'
+          '    return Array(list(zip(*xx)))'
```
cbm755, line 73: should these go through your new helper function?  (not saying they should, just asking)

inst/@sym/horzcat.m
```diff
@@ -45,25 +45,8 @@
...
-          '            _proc.append(i)'
-          'return sp.MatrixBase.hstack(*_proc),'
-          };
-
-  for i = 1:nargin
-    varargin{i} = sym(varargin{i});
-  end
-  h = pycall_sympy__ (cmd, varargin{:});
+  args = cellfun (@transpose, varargin, 'UniformOutput', false);
+  h = vertcat (args{:}).';
```
cbm755, line 49: slick!

inst/@sym/horzcat.m
```diff
@@ -1,4 +1,4 @@
-%% Copyright (C) 2014-2017, 2019 Colin B. Macdonald
+%% Copyright (C) 2022 Alex Vong
```
cbm755, line 1: suggest leaving my copyright notice here, unless its a brand new file.

alexvong1995: Aha, this is a typo. I was suppose to a copy & edit.

inst/@sym/transpose.m
alexvong1995: Is it alright if I do

```
If isinstance(x, (MatrixBase, NDimArray)):
    xx = x.tolist()
    return make_matrix_or_array(list(zip(*xx)))
```

directly instead of splitting the case?

inst/@sym/horzcat.m
alexvong1995: I also forget to add copyright notice in `inst/@sym/transpose.m`

inst/@sym/transpose.m
alexvong1995: This actually doesn't work because of issues with empty matrices. We have to split it into 2 cases.

alexvong1995: Replying to your original Q:

I thought about this but don't have a clear ans. Using `make_matrix_or_array` means sticking to a more consistent interface, while using `Array` is better for performance because no Expr checks are performed. I guess sticking to a more consistent interface is more important (?)

alexvong1995: As per my latest comment in #1124, we may want to use `is_2d_sym` and `is_non_matrix_2d_sym` for generality sake.

alexvong1995:
PR updated to fix issues raised, I think we can worry about generalizing to non-`Array` (e.g. `TableForm`) in another PR

cbm755:
I agree for TableForm.  One thing at a time

inst/@sym/transpose.m
cbm755: Yeah that sounds reasonable.   A new PR perhaps

PR #1219, opened 2022-09-02
@sym/repmat: Re-implement function without using 'pycall_sympy__'.

### LINKED ISSUE #1218: @sym/repmat: Not outputing empty matrices the same as repmat
Currently,

```repmat(sym([1 2; 3 4]), 3, 0)```
outputs
```ans = (sym) []  (empty 3×0 matrix)```

But

```repmat([1 2; 3 4], 3, 0)```
outputs
```ans = [](6x0)```

Which one is the correct behaviour? I don't think SMT documentation mentions this...

### PR DESCRIPTION
**Pull request by alexvong1995**
This PR mainly re-implements `@sym/repmat` without using `pycall_sympy__`. It has the advantage of making the function compatible with non-Matrix 2D sym. It also fixes #1218 at the same time.

### DIFF
--- a/inst/@sym/repmat.m
+++ b/inst/@sym/repmat.m
@@ -1,4 +1,5 @@
 %% Copyright (C) 2014, 2016, 2019 Colin B. Macdonald
+%% Copyright (C) 2022 Alex Vong
 %%
 %% This file is part of OctSymPy.
 %%
@@ -51,18 +52,21 @@ function B = repmat(A, n, m)
     print_usage ();
   end
 
-  cmd = { '(A, n, m) = _ins'
-          'if n == 0 or m == 0:'
-          '    return sp.Matrix(n, m, [])'
-          'if A is None or not A.is_Matrix:'
-          '    A = sp.Matrix([A])'
-          'L = [A]*m'
-          'B = sp.Matrix.hstack(*L)'
-          'L = [B]*n'
-          'B = sp.Matrix.vstack(*L)'
-          'return B' };
+  if n == 0 || m == 0
+    [nrows_A, ncols_A] = size (A);
+    B = sym (zeros (n * nrows_A, m * ncols_A));
+    return
+  end
 
-  B = pycall_sympy__ (cmd, sym(A), int32(n), int32(m));
+  %% duplicate A horizontally m times
+  MM = cellfun (@(varargin) A, cell (1, m), 'UniformOutput', false);
+  M = horzcat (MM{:});
+
+  %% now duplicate that result vertically n times
+  NN = cellfun (@(varargin) M, cell (1, n), 'UniformOutput', false);
+  N = vertcat (NN{:});
+
+  B = N;
 
 end
 
@@ -95,3 +99,11 @@ end
 %! assert (isequal (size(A), [0 3]))
 %! A = repmat(sym(pi), [2 0]);
 %! assert (isequal (size(A), [2 0]))
+
+%!test
+%! % even more empties, see also https://github.com/cbm755/octsympy/issues/1218
+%! A = randi (16, 5, 7);
+%! assert (isa (repmat (sym (A), [0 3]), 'sym'));
+%! assert (isa (repmat (sym (A), [2 0]), 'sym'));
+%! assert (isequal (sym (repmat (A, [0 3])), (repmat (sym (A), [0 3]))));
+%! assert (isequal (sym (repmat (A, [2 0])), (repmat (sym (A), [2 0]))));
--- a/inst/@sym/private/mat_rclist_access.m
+++ b/inst/@sym/private/mat_rclist_access.m
@@ -33,8 +33,7 @@ function z = mat_rclist_access(A, r, c)
     error('this routine is for a list of rows and cols');
   end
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: inst/@sym/repmat.m [hunk 1]
```diff
@@ -51,18 +52,20 @@
...
-          'if A is None or not A.is_Matrix:'
-          '    A = sp.Matrix([A])'
-          'L = [A]*m'
-          'B = sp.Matrix.hstack(*L)'
-          'L = [B]*n'
-          'B = sp.Matrix.vstack(*L)'
-          'return B' };
+  if n == 0 || m == 0
+    [nrows_A, ncols_A] = size (A);
+    B = zeros (n * nrows_A, m * ncols_A);
```
cbm755, line 57: this looks like B would be double (I didn't try it yet!)

but the return type should be sym right?

alexvong1995: Nice catch! Let's add tests to check this.

inst/@sym/repmat.m
```diff
@@ -51,18 +52,20 @@
...
+    return
+  end
+
+  make_1d_cell = @(n, init) cellfun (@(unused) init, cell (1, n), ...
+                                     'UniformOutput', false);
+  MM = make_1d_cell (m, A);
```
cbm755, line 63: I find this a bit hard to read but I don't see an alternative way ...

consider:
1.  can we use a function function instead of the outer anon function?  (or not, not sure about matlab compat)
2. naming it "make_1d_cell_copy"?
3. better word than `init`?  maybe `A` or `X`?

alexvong1995: What we are doing here is basically creating a cell array of length `n` with `init` as initial value. We need this because `cell(...)` doesn't provide a way for specifying initial value.

For 1, I haven't used nested functions before but I seldom see them being mentioned. Perhaps there're some compat issue / other drawbacks?

For 2 & 3, sure I can change that and add a comment to document the function as well (similar to python docstring).

```
  # Construct a 1D cell array of length N with initial value INIT
  make_1d_cell = @(n, init) cellfun (@(unused) init, cell (1, n), ...
                                     'UniformOutput', false);
```

or

```
  # Construct a 1D cell array of length N with initial value X
  make_1d_cell_copy = @(n, x) cellfun (@(unused) x, cell (1, n), ...
                                       'UniformOutput', false);
```

cbm755: Or how about just repeat the cellfun part:
```
%% construct 1-d cell array of length m and concatenate horizontally
MM = cellfun (@(unused) A, cell (1, m), 'UniformOutput', false);
M = horzcat (MM{:});
%% now duplicate that result vertically n times
NN = cellfun (@(unused) A, cell (1, n), 'UniformOutput', false);
B = vertcat (NN{:});
```
(not tested but you get the idea)

cbm755: maybe without `unused` var?
```
MM = cellfun (@() A, cell (1, m), 'UniformOutput', false);
```

cbm755:
Anyway, modulo few things above, this look good to me

cbm755:
So merge when you're happy with the cellfun bikeshedding ;-)

inst/@sym/repmat.m
alexvong1995: `@() ...` doesn't work because we need a function that takes >= 1 arguments. How about using `varargin`? I've also changed the comment a bit but the idea should be the same.

```
  %% duplicate A horizontally m times
  MM = cellfun (@(varargin) A, cell (1, m), 'UniformOutput', false);
  M = horzcat (MM{:});

  %% now duplicate that result vertically n times
  NN = cellfun (@(varargin) M, cell (1, n), 'UniformOutput', false);
  N = vertcat (NN{:});
```

cbm755: sure, `@()` seemed ok when I tried it:
```octave:1> A = [1 2 3; 4 5 6]; A = sym(A)
Symbolic pkg v3.0.1+: Python communication link active, SymPy v1.10.1.
A = (sym 2×3 matrix)
  ⎡1  2  3⎤
  ⎢       ⎥
  ⎣4  5  6⎦
octave:2> MM = cellfun (@() A, cell (1, 3), 'UniformOutput', false)
MM =
{
  [1,1] =
    <class sym>
  [1,2] =
    <class sym>
  [1,3] =
    <class sym>
}

octave:3> horzcat (MM{:})
ans = (sym 2×9 matrix)
  ⎡1  2  3  1  2  3  1  2  3⎤
  ⎢                         ⎥
  ⎣4  5  6  4  5  6  4  5  6⎦
```


But maybe its not portable (that was on Octave 6.4.0).  Anyway, `varargin` is fine

alexvong1995: Yep, I think it's not portable. On Octave 7.2.0, I get

```
A = (sym 2×3 matrix)

  ⎡1  2  3⎤
  ⎢       ⎥
  ⎣4  5  6⎦

octave:2> MM = cellfun (@() A, cell (1, 3), 'UniformOutput', false)
error: @<anonymous>: function called with too many inputs
error: called from
    @<anonymous>
```

This is also how it's usually handled in other languages so it shouldn't be a bug...

PR #1223, opened 2022-09-02
elementwise_op: use 2d loop for Array

### LINKED ISSUE #1222: Array_not_Matrix: boolean ops on Arrays are broken
```
>> t = sym(true)
>> w = [t t]
pydebug: make_matrix_or_array: making 2D Array...
pydebug: I am here with an array with shape (2, 1)
pydebug: make_matrix_or_array: making 2D Array...
pydebug: I am here with an array with shape (1, 2)
w = (sym) [[True  True]]  (1×2 matrix)
>> sympy(w)
ans = ImmutableDenseNDimArray(Tuple(true, true), Tuple(Integer(1), Integer(2)))
```
So far so good!  But try boolean op:
```
>> w & w
error: Python exception: TypeError: expecting bool or Boolean, not `[[True, True]]`.
    occurred at line 14 of the Python code block:
    elements.append(_op(*[k[i] if isinstance(k, MatrixBase) else k for k in _ins]))
error: called from
    pycall_sympy__ at line 179 column 7
    elementwise_op at line 105 column 5
    and at line 47 column 5
```

cbm755:
Array linear indexing is not the same as Matrix linear indexing.
Fixes #1222.

### DIFF
--- a/inst/private/python_header.py
+++ b/inst/private/python_header.py
@@ -241,7 +241,7 @@ except:
 
 
 try:
-    def make_matrix_or_array(ls_of_ls, dbg_no_array=True):
+    def make_matrix_or_array(ls_of_ls, dbg_no_array=False):
         # should be kept in sync with the same function
         # defined in inst/private/python_ipc_native.m
         # FIXME: dbg_no_array is currently used for debugging,
--- a/inst/@sym/private/elementwise_op.m
+++ b/inst/@sym/private/elementwise_op.m
@@ -95,8 +95,10 @@ function z = elementwise_op(scalar_fcn, varargin)
           % at least one input was a matrix:
           '# dbout(f"at least one matrix param, shape={q.shape}")'
           'elements = []'
-          'for i in range(0, len(q)):'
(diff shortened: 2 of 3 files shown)


cbm755:
@alexvong1995 OK to merge this and #1224 and #1227?

inst/private/python_header.py
```diff
@@ -241,7 +241,7 @@ def octoutput(x, et):
 
 
 try:
-    def make_matrix_or_array(ls_of_ls, dbg_no_array=True):
+    def make_matrix_or_array(ls_of_ls, dbg_no_array=False):
         # should be kept in sync with the same function
         # defined in inst/private/python_ipc_native.m
```
alexvong1995, line 246: I think we should also change `dbg_no_array` from `True` to `False` in `inst/private/python_ipc_native.m`.

LGTM otherwise as this test (for `@sym/and`)

```
%!test
%! % array
%! w = [t t f f];
%! z = [t f t f];
%! assert (isequal (w & z, [t f f f]))
```

now passes!

alexvong1995: Resolved in 37724d624c6952669709c85e058f3e0fc72b604c

Issue #1226, opened 2022-09-03
subs: needs to use  make_matrix_or_array


cbm755:
MWE, from `test @sym/and`:

```
syms x
e = (x == 3) & (x^2 == 9);
subs(e, x, [-3 0 3])
sympy(ans)
ans = MutableDenseMatrix([[false, false, true]])
```

Here `e` is a scalar but `subs` "broadcasts" to the size of `[-3 0 3]`.

I guess the implementation needs to do a list-of-lists thing and then call `make_matrix_or_array`.

Blocks #1124.

alexvong1995:
Fixed by #1227

PR #1227, opened 2022-09-03
subs: use make_matrix_or_array and use 2-d indexing

cbm755:
Fixes Issue #1226.  Related to Issue #1222.

This makes `test @sym/and`, or, etc pass.


alexvong1995 (approved):
I don't see any problems. Perhaps it's possible to simplify things a bit in `@sym/subs` but I don't know how to do it right now.

Anyway, the
```
syms x
e = (x == 3) & (x^2 == 9)
subs(e, x, [-3 0 3])
```
test case passes, so it's good to merge.

PR #1231, opened 2022-09-06
Use sympy functions and generalise `make_matrix_or_array`

alexvong1995:
This PR is mostly about changing embedded python code to use sympy functions, similar to #1230.

We also fix an (im)mutability bug exposed by this change. Finally, we generalise `make_matrix_or_array` to accept an iterable of iterables.


cbm755:
> We also fix an (im)mutability bug exposed by this change. 

Does this need a new BIST?

- - - -

LGTM, please merge when you're ready

alexvong1995:
Currently, the test
```
%!test
%! % Disassemble matrix
%! a = sym([1 2; 3 4; 5 6]);
%! b = sym([3 5 2 4 6]);
%! a(1) = [];
%! assert (isequal (a, b));
```
fails if we revert 321ee1709f011ea568a243d5f8722fa6e3a14db9 so it should be fine (?)

PR #1233, opened 2022-09-07
elementwise_op: call make_array_or_matrix

cbm755:
Instead of doing it ourselves, prepare a list of lists and call the centralized helper function.


cbm755:
@alexvong1995 I tried gating `dbg_no_array` to True on old (that is, currently released) SymPy.  We're going to want all CI tests to pass in that case.

But they do not.

Consider:
```
 a = zeros (sym(3), 0);
 assert (size (a == a), [3, 0])
```
Here `a == a` gives 0x0.  Perhaps there is a fundamental limitation of list-of-lists representation: you can make `[[], [], []]` for a 3x0 but there is no representation for a 0x3.

One fix here would be to introduce an even lower-level "flat-list + shape" helper function...  What do you think?

inst/@sym/private/elementwise_op.m
```diff
@@ -94,15 +94,13 @@
...
           '# dbout(f"at least one matrix param, shape={q.shape}")'
-          'elements = []'
-	  'assert len(q.shape) == 2, "non-2D arrays/tensors not yet supported"'
-          'for i in range(0, q.shape[0]):'
-          '    for j in range(0, q.shape[1]):'
-          '        elements.append(_op(*[k[i, j] if isinstance(k, (MatrixBase, NDimArray)) else k for k in _ins]))'
-          'if all(isinstance(x, Expr) for x in elements):'
-          '    return Matrix(*q.shape, elements)'
-          'dbout(f"elementwise_op: returning an Array not a Matrix")'
-          'return Array(elements, shape=q.shape)' ];
```
cbm755, line 105: note in the old code here we have `elements` a flat list.  Then we'd simply make `make_matrix_or_array(<iterable>, <shape>)`...

cbm755:
@alexvong1995 I moved this discussion to #1239: very relevant there.  And yes we can keep both.

PR #1235, opened 2022-09-07
[WIP] Add 2D sym functions and use them.

alexvong1995:
We add 2D sym functions to abstract over the idea of 2D sym so that we
can choose to represent non-`Matrix` 2D sym to be something other than
an `Array` in the future. For example, we could use `TableForm`.

This is incomplete. More functions need to use these 2D sym
functions. Not sure if this is needed anymore as `TableForm` was shown
to be lacking various features such as indexing and taking
transpose. No other candidates are being considered at the moment.

This PR is introduced to avoid these changes to be forgotten. We can
close it later if this is definitely not needed or it bit rots.


cbm755:
I force-pushed this onto top of the latest Array_not_Matrix

cbm755:
I cherry-picked a one of the commits out of here into #1194.

PR #1237, opened 2022-09-13
subs/vertcat: after Array port was creating incorrect empties

### LINKED ISSUE #1236: make_2d_sym: support flat list + shape input
@alexvong1995 I tried gating `dbg_no_array` to True on old (that is, currently released) SymPy.  We're going to want all CI tests to pass in that case.

But they do not.

Consider:
```
 a = zeros (sym(3), 0);
 assert (size (a == a), [3, 0])
```
Here `a == a` gives 0x0.  Perhaps there is a fundamental limitation of list-of-lists representation: you can make `[[], [], []]` for a 3x0 but there is no representation for a 0x3.

One fix here would be to introduce an even lower-level "flat-list + shape" helper function...  What do you think?

_Originally posted by @cbm755 in https://github.com/cbm755/octsympy/issues/1233#issuecomment-1238818601_

cbm755:
Fixes #1236.  Or at least the part of #1236 that pertains to `subs`. Added new tests.

### DIFF
--- a/inst/@sym/horzcat.m
+++ b/inst/@sym/horzcat.m
@@ -1,4 +1,4 @@
-%% Copyright (C) 2014-2017, 2019 Colin B. Macdonald
+%% Copyright (C) 2014-2017, 2019, 2022 Colin B. Macdonald
 %% Copyright (C) 2022 Alex Vong
 %%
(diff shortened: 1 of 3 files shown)


cbm755:
transient test failures, merging

PR #1239, opened 2022-09-13
Port helpers from list-of-lists to flat+shape

cbm755:
Related to #1236.  I don't see a problem with the list-of-lists approach, but there are only 2 routines using it so we can delete some code if we port them to flat + shape.

@alexvong1995 can you port `@sym/private/mat_rclist_access.m` to `_make_2d_sym`?  And then delete the `make_2d_sym` function.

### DIFF
--- a/inst/@sym/private/mat_rclist_access.m
+++ b/inst/@sym/private/mat_rclist_access.m
@@ -35,9 +35,9 @@ function z = mat_rclist_access(A, r, c)
 
   cmd = {'(A, rr, cc) = _ins'
          'AA = A.tolist() if isinstance(A, (MatrixBase, NDimArray)) else [[A]]'
-         'MM = [[AA[i][j]] for i, j in zip(rr, cc)]'
-         'M = make_2d_sym(MM)'
-         'return M,'};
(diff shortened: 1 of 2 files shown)


cbm755:
>  In any case, I think we should keep the high-level function so that we use the low-level function only when we have to. I think it's quite common for types / data structures / classes to have multiple constructors. WDYT?

Let's discuss here. 

1. We could keep both "flat + shape" and "list-of-lists".
2. do some polymorphic thing (when `shape=` kwarg is passed, we expect a flat array).
3. keep only flat + shape.

I was working toward 3 but I don't feel that strongly about it.   I don't think list-of-lists is possible in general b/c of the reasons outlined elsewhere (and even in the help of SymPy's `.tolist()` operator).

cbm755:
How about `make_2d_sym_list_of_lists` flattens and then calls `_make_2d_sym`?  That seems the easiest way to keep both.

However, I think every caller of `make_2d_sym_list_of_lists` will need to think very carefully if they handle empties correctly: which is why I think it might not be worth keeping.

cbm755:
Merging this for now, above decisions not really resolved but I don't think this hurts

PR #1264, opened 2023-07-24
Metainfo: fix extends, change ID

cbm755:
Also, a good test if CI still works after repo move

### DIFF
--- a/io.github.gnu_octave.symbolic.metainfo.xml
+++ b/io.github.gnu_octave.symbolic.metainfo.xml
@@ -8,7 +8,7 @@ notice and this notice are preserved.  This file is offered as-is,
 without any warranty.
 -->
 <component type="addon">
-  <id>octave-symbolic</id>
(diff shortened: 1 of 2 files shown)


io.github.gnu_octave.symbolic.metainfo.xml
```diff
@@ -1,15 +1,15 @@
...
 
 Copying and distribution of this file, with or without modification,
 are permitted in any medium without royalty provided the copyright
 notice and this notice are preserved.  This file is offered as-is,
 without any warranty.
 -->
 <component type="addon">
-  <id>octave-symbolic</id>
-  <extends>www.octave.org-octave.desktop</extends>
+  <id>io.github.gnu_octave.symbolic.metainfo.xml</id>
```
cbm755, line 11: FTR, this underscore is not a typo: hyphens are not recommended by the AppStream docs

PR #1271, opened 2023-08-16
CI: Add tests on Windows (mingw-w64)

mmuetzel:
This adds some tests on a Windows runner.

It uses the version of Octave that is distributed by MSYS2. That version is not identical to the one that gets installed with the Octave for Windows installer built with MXE Octave. But it is similar in that it's also based on MinGW.
The main difference is probably that MSYS2 has a native Python (instead of the Cygwin Python that is included in Octave for Windows).
But that difference also makes it easier to install the necessary packages in the CI.
Additionally, MSYS2 follows a rolling release model. That means that some packages are likely newer than the ones included in Octave for Windows (which follows a more conservative model of updating packages only on major releases).

I tried to stay close to the existing tests on other platforms. But I needed to add work-arounds for some issues that occur in the CI. I left some comments inline that hopefully clarify why they are there.

I hope you'll find this useful.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -272,3 +272,90 @@ jobs:
       - name: uninstall package
         run: |
           octave --eval "pkg uninstall symbolic; pkg list"
+
+
+  windows_msys2_mingw:
+
+    runs-on: windows-latest
+
+    name: mingw-w64 ${{ matrix.msystem }}
+
+    defaults:
+      run:
+        # Use MSYS2 as default shell
+        shell: msys2 {0}
+
+    strategy:
+      # Allow other runners in the matrix to continue if some fail
+      fail-fast: false
+
+      matrix:
+        msystem: [MINGW64, UCRT64]
+        include:
+          - msystem: MINGW64
+            target-prefix: mingw-w64-x86_64
+          - msystem: UCRT64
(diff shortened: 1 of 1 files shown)


cbm755:
This is very cool!  I'm going to merge it.  I'm slightly nervous that it hasn't run for a month and won't work in main...

Here's the versions it ran when you submitted it:
```
MINGW64_NT-10.0-20348 fv-az553-958 3.4.7.x86_64 2023-08-09 17:32 UTC x86_64 Msys
GNU Octave, version 8.3.0
Python 3.11.4
1.12


MINGW64_NT-10.0-20348 fv-az89-447 3.4.7.x86_64 2023-08-09 17:32 UTC x86_64 Msys
GNU Octave, version 8.3.0
Python 3.11.4
1.12
```

cbm755:
New run:
```
MINGW64_NT-10.0-20348 fv-az843-488 3.4.9.x86_64 2023-09-07 08:43 UTC x86_64 Msys
GNU Octave, version 8.3.0
Python 3.11.5
1.12

MINGW64_NT-10.0-20348 fv-az252-667 3.4.9.x86_64 2023-09-07 08:43 UTC x86_64 Msys
GNU Octave, version 8.3.0
Python 3.11.5
1.12
```

PR #1283, opened 2024-01-03
Add collect function

### LINKED ISSUE #1019: SymPy collect function is not implemented
Collect function is used to collect common powers of a variable in an expression and has the syntax "collect(expr, var)".

Documentation:
https://docs.sympy.org/latest/tutorial/simplification.html#collect

Proposal for implementation of collect.m:

function y = collect(expr, var)

  cmd = { '(expr, var) = _ins'
          'return sympy.collect(expr, var),' };
  y = pycall_sympy__ (cmd, sym(expr), sym(var));

end

cbm755:
Fixes #1019.  Thanks Robert Jenssen, who sent such an implementation years ago.  I added some doctests.  Also fixes upstream issue [1].

[1] https://savannah.gnu.org/bugs/?59014

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,7 @@ octsympy 3.1.1+
 
   * New symbolic commands:
 
+          collect
           fplot
 
   * Move repo to https://github.com/gnu-octave/symbolic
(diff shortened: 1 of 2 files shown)


cbm755:
Thank you @robertgj sorry this took so long!

PR #1301, opened 2024-05-16
Codespell: trying to get CI running

cbm755:
Not sure I'm a fan of GitHub Actions: on GitLab I would just script precisely what I want to happen.  Here, I have no idea what "PWD" is, where to put the support files etc... :(


cbm755:
Got it!  the issue is you must **also** use the checkout action, like this:

```yaml
  codespell:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: codespell-project/actions-codespell@v2
        with:
```

Shall MWPS!

PR #1305, opened 2024-07-04
Improve description of 'order' for 'taylor' function

vitaminace33:
Currently, [`taylor` returns one order less than given](https://github.com/gnu-octave/symbolic/issues/1303). This improves the description of the 'order' key.


cbm755:
This looks good, sorry for taking so long to get back to this!

Issue #1315, opened 2024-11-30
"test @sym/intersect" failed


dasergatskov:
With octave 10..0 symbolic 3.2.1 and current git clone of SymPy 
```
commit ba68537e73dd40fb34fa9c5aaf46c5a6e1ef12c7 (HEAD -> master, origin/master, origin/HEAD)
Merge: 357ede9874 e0f229da64
Author: Björn Dahlgren <bjodah@gmail.com>
Date:   Wed Nov 27 07:40:39 2024 +0100
```
I see:
```
octave:5> test @sym/intersect
***** test
 % empty input
 A = sym([1 2]);
 C = intersect(A, []);
 assert (isequal (C, sym([])))
!!!!! test failed
assert (isequal (C, sym ({}))) failed
octave:6> A = sym([1 2])
A = (sym) [1  2]  (1×2 matrix)
octave:7> C = intersect(A, [])
C = (sym) []  (empty 1×0 matrix)
octave:8> sym([])
ans = (sym) []  (empty 0×0 matrix)
```

Sonu0305:
Hi @dasergatskov , 
I think this test failing was fixed in #1311
Thank You!

dasergatskov:
Thanks. "make test" on git pull passed, but I cannot install it. After "make dist":
```
octave:1> pkg install -verbose ./symbolic-3.2.1+.tar.gz 
mkdir (/var/folders/ds/5xsh6x4914x78jswflyz20c40000gn/T/oct-YDETJd)
untar (./symbolic-3.2.1+.tar.gz, /var/folders/ds/5xsh6x4914x78jswflyz20c40000gn/T/oct-YDETJd)
error: dirlist(3): out of bound 2 (dimensions are 2x1)
error: called from
    install at line 120 column 11
    pkg at line 619 column 9
octave:2> 
```

dasergatskov:
OK. I figured that out.  My "tar" is BSD tar. My GNU tar is "gtar".
I modified Makefile:
```
diff --git a/Makefile b/Makefile
index 12dcbcb4..243798b5 100644
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ HTML_TARBALL := ${HTML_DIR}.tar.gz
 
 OCTAVE ?= octave
 MATLAB ?= matlab
+TAR ?= tar
 
 .PHONY: help clean install test doctest dist dist_zip html matlab_test matlab_pkg
 
[...]

PR #1316, opened 2024-11-30
Fix compatibility with BSD tar by supporting both tar and gtar

Sonu0305:
This PR addresses the issue (#1315) with packaging and tarball creation on systems using BSD `tar`. On such systems, the default `tar` command may not work as expected, and GNU `tar` (`gtar`) is required instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ HTML_TARBALL := ${HTML_DIR}.tar.gz
 
 OCTAVE ?= octave
 MATLAB ?= matlab
+TAR ?= tar
 
 .PHONY: help clean install test doctest dist dist_zip html matlab_test matlab_pkg
 
@@ -46,7 +47,7 @@ define create_tarball
 $(shell cd $(dir $(1)) \
     && find $(notdir $(1)) -print0 \
     | LC_ALL=C sort -z \
-    | tar c --mtime="$(GIT_DATE)" \
+    | $(TAR) c --mtime="$(GIT_DATE)" \
             --owner=root --group=root --numeric-owner \
             --no-recursion --null -T - -f - \
     | gzip -9n > "$(2)")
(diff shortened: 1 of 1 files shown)


Sonu0305:
Hi @cbm755 ,
Since I do not have write access to the repository, I kindly request if you could review the pull request. Please let me know if any further changes or adjustments are needed.
Thank you for your time and support!

cbm755:
Looks good to me!  This seems familiar, maybe something that has come up in other Octave packages...   If you have a moment, can you also check https://github.com/gnu-octave/octave-doctest?   maybe `image` and `interval` too?  IIRC, the Makefile was pretty common...

Sonu0305:
> Looks good to me! This seems familiar, maybe something that has come up in other Octave packages... If you have a moment, can you also check https://github.com/gnu-octave/octave-doctest? maybe `image` and `interval` too? IIRC, the Makefile was pretty common...

For **octave-doctest**:
The `Makefile` for `doctest` also sets up support for `tar`, but in a slightly different way. It attempts to find `gtar` (GNU tar) first and, if not found, defaults to using the system `tar`. This is done through the line:
```makefile
TAR := $(shell which gtar 2>/dev/null || echo tar)
``` 
Do you want me to modify the `doctest` `Makefile` to use the same method as in `symbolic`'s `Makefile`, where the `TAR` variable is not prioritized (just uses whatever `tar` is available in the environment).

Sonu0305:
> > Looks good to me! This seems familiar, maybe something that has come up in other Octave packages... If you have a moment, can you also check https://github.com/gnu-octave/octave-doctest? maybe `image` and `interval` too? IIRC, the Makefile was pretty common...
> 
> @cbm755 For **octave-doctest**: The `Makefile` for `doctest` also sets up support for `tar`, but in a slightly different way. It attempts to find `gtar` (GNU tar) first and, if not found, defaults to using the system `tar`. This is done through the line:
> 
> ```makefile
> TAR := $(shell which gtar 2>/dev/null || echo tar)
> ```
> 
> Do you want me to modify the `doctest` `Makefile` to use the same method as in `symbolic`'s `Makefile`, where the `TAR` variable is not prioritized (just uses whatever `tar` is available in the environment)?

Hi @cbm755 ,
Could you review this comment of mine, when you have a chance
Thank You.

